PR #28, opened 2013-06-03
Make Pact Contract constructors public

aljoscha:
Make them public so that they can be invoked with a builder and a mixin trait from scala code.


StephanEwen:
Can you verify that "protected" is not enough? Since mixing in a trait causes the Scala compiler to generate a sub-class, I thought protected suffices.

aljoscha:
You're right, I'm sorry. At some point I did something where I did not have the mixin and that wouldn't work without the ctor being public. But now I have mixins again and it works again.

Issue #37, opened 2013-06-17
Enable visualization of custom iteration statistics


sscdotopen:
The Nephele visualizer should be able to display custom iteration statistics such as the rate of convergence e.g.

PR #38, opened 2013-06-17
#37 Enable visualization of custom iteration statistics

sscdotopen:


StephanEwen:
This is unfortunately outdated code for the SWT visualization that was used during the SIGMOD demo. It needs quite a bit of reworking and merging. Since we are pushing for the browser-based visualization at the moment, this code will become obsolete anyways in a bit.

StephanEwen:
The SWT visualization was removed to to licence issues: LGPL conflicts with the ASL 2. So we need to come up with a new way of doing this.

PR #41, opened 2013-06-22
Updated Maven artifacts version to 0.2-ozone.

### LINKED ISSUE #27: Change Maven Artifact Names
Currently, ozone uses the same artifact names as Stratosphere 0.2. That means it is not possible to work with both in parallel (as in code migration, cross testing, ...)

Ozone should have new artifact names and a new Version. I propose Version 0.5 (some leap from 0.2)

aalexandrov:
Allows deploying Ozone to a local repository next to a vanilla Stratosphere distribution. Closes #27.


StephanEwen:
Merged in 35ff4dcd2ed84aa8711e4d3fea1caae622428457

PR #42, opened 2013-07-01
Make stratosphere work on Windows/cygwin

aljoscha:
Also fix WebInterfaceServer.java to properly check whether path
is absolute on Windows.

Also, default directory entries in nephele-user.xml and pact-user.xml
are now commented out because java code will load proper default.


StephanEwen:
Can you please update the code to the latest HEAD? There are two issues:

1) The pull request to split the libs directory into lib (for core libs) and lib_client requires to mangle the client part of the classpath as well in the relevant scripts (webfrontent, pact-client).

2) Not all windows setups are cygwin (I know that Daniel has been running stratosphere on bare windows, I think with gnuwin utils or so). Please check for a cygwin env variable, or the presence of the cygpath command when mangeling paths.

aljoscha:
I rebased to the latest HEAD and also fixed the Plain Windows vs. Cygwin thing.

Where is the pull request to split the libs directory? Or was this already merged?

StephanEwen:
That one is already merged. You should find the references to the client directory in the client shell script files.

aljoscha:
Ok, I fixed the issue with lib_client as well.

PR #49, opened 2013-07-05
Support for Cloudera Hadoop

mleich:
# Cloudera Hadoop Support for Ozone

These changes allow to use and easily switch between different Hadoop versions.
## How to use:
- adjust the hadoop version in the main pom (samples for Hadoop 0.20.203.0 and 2.0.0-cdh4.2.1 are provided)
- adjust all poms that require Hadoop to use the correct jar names (hadoop-core changed to hadoop-common, hadoop-hdfs may need to be added)
## Side effects:

Different Hadoop versions use different jar names and have different dependencies that need to be added to the classpath.
Previously all jar names were explicitly listed in the application startup scripts. While this allows for tailor-made classpaths it is also a major maintenance overhead.
The behavior was changed so the startup scripts now add all jars from the <ozone_root>/lib folder to the classpath.
Jars required by pact-clients have been moved to lib_clients, to avoid unnecessary overhead for worker tasks.
In the end this means far less classpath management overhead while retaining a reasonably clean classpath.

### DIFF
--- a/nephele/nephele-hdfs/pom.xml
+++ b/nephele/nephele-hdfs/pom.xml
@@ -23,10 +23,23 @@
 			<version>${project.version}</version>
 		</dependency>
 
-
+		<!-- Ye Olde Hadoop
 		<dependency>
 			<groupId>org.apache.hadoop</groupId>
 			<artifactId>hadoop-core</artifactId>
+		</dependency>
+		-->
+		
+		
+		<!-- Coudera Hadoop (no need for hadoop-core anymore) -->
+		<dependency>
+			<groupId>org.apache.hadoop</groupId>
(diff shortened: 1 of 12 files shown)


mleich:
OK, right now eu.stratosphere.pact.test.contracts.io.CustomDataTypeTest seems to run infinitely.
I'll take a look at that, not sure why this happens.

mleich:
Don't know why this wasn't triggered before, but changing the Hadoop version also changed the log level to debug during build.
Should work just fine now.

StephanEwen:
Rebased and merged in b407f8b37a4ded820a25cc1cc03da064988c2f0d.

Thanks!

PR #51, opened 2013-07-12
re-openable hashjoin integration test

rmetzger:
This is the fix for issue: https://github.com/dimalabs/ozone/issues/32

I used some of the tests of HashTableITCase as a starting point to test the new re-openable hash join. The current implementation probes the input five times.


StephanEwen:
Merged in 42cdca4006f6f521f0e497732f7fda0295b01d7d.

PR #54, opened 2013-07-14
Fix Multiplexing Deadlock

### LINKED ISSUE #20: Fix Multiplexing Deadlock
Change assignment from logical channels to physical connections to prevent starvation effects that cause distributed deadlocks in the network stack.

uce:
This fixes #20 by assigning a unique connectionID (per stage) to every ExecutionEdge.


uce:
The failing ~~ConnectedComponentsNepheleITCase~~ `CoGroupConnectedComponentsITCase` passes only randomly for me locally already _before_ this fix. I just ran it 10 times and only 1 run passed. The failing runs throw the "Vertex in wrong component" `AssertionError`. Can anybody confirm this?

I just noticed it after the travis build failed. Before issuing the pull request, everything passed for me (i.e. it was the 1 out of 10 runs that passes?). Also #55 doesn't change anything used in the test, but fails.

uce:
Stephan just pointed out that I referenced the wrong test... sorry. I meant that `CoGroupConnectedComponentsITCase` works randomly. And this seems to be known problem. So everything's fine.

StephanEwen:
Merged in 4542555ea4cd4e1d7253eb7ce96e140fd0375e83

PR #55, opened 2013-07-14
Fix ArrayIndexOutOfBoundsException when reading test input > 32KB

uce:
This fixes a bug, where tests in `pact-tests` using `Tuple` or `IntTupleDataInFormat` would throw `ArrayIndexOutOfBoundsException` on input > 32 KB.

I used TPCH-9 query from `pact-tests` to reliably reproduce the multiplexing deadlock (see issue #20) even on small scale (because of the relatively long processing pipeline).


StephanEwen:
Rebased and merged in 05401481550685bcaa3ba22c63590988dde7eaf1

PR #61, opened 2013-08-10
Reducer without grouping key ("ALL_GROUP") fails in optimizer

rmetzger:
This pull request contains a test case that validates a bug with the pact-optimizer.

(The test obviously fails, because I did not include a fix for the bug!)
I also fixed a little typo and made an error message a bit more understandable by giving an example.


StephanEwen:
Merged manually in b8d213da44259b8ed55cb10e01204fac4db8dbe0 and 88d7305a5267aac598949519275123208195daf7.

PR #62, opened 2013-08-12
Minor bugfix for TextInputFormat + Testcase

rmetzger:
I found this bug during the development of customer use-cases.

I guess its an easy fix but I want to focus on the use-cases right now (and I have a workaround).

The integration test will obviously fail for this pull request because it contains a test case for an unfixed bug.


uce:
Swapping the failing line with the next one in `TextInputFormat` fixed it for me. The problem was that the current buffer `position` is not allowed to be greater than the current `limit`. I've sent you a patch for this to push to your branch.

PR #69, opened 2013-08-26
Removed Nephele plugins

uce:
As discussed last Friday, I removed the plugins feature from Nephele.


uce:
My comment on this was too short. (We also discussed this offline which we probably shouldn't have in the first place.)

Nephele plugins allow to rewrite the `ExecutionGraph` at runtime. This kind of low level interaction with the system is imho not useful (and if I remember correctly the consensus was that it is not used by anybody or just a single person?!).

If we still agree on this issue, we should finally merge it. It increases the "complexity" of the Nephele layer and I don't think that we want to support this kind of job manipulation in the future.

rmetzger:
I'm going to discuss the pull request tomorrow with Stephan. I think we are going to merge it.

sscdotopen:
In general I'd always vote in favor of removing stuff that is rarely/never used. Less code means less bugs and less maintenance.

StephanEwen:
Manually merged in 921278e322eaae008dae75e20a505f53aaeed1ca

PR #70, opened 2013-08-28
Remove pact-examples dependency in pact-clients.

carabolic:
pact-clients was dependent on pact-examples in order to run
LocalExecuterTest. The problem with that was, that this prevented the
use of the LocalExecuter within the pact-examples module (circular
dependency). Hence the LocalExecuterTest was moved to the pact-test
module and the pact-examples depedency was removed from pact-clients.

### DIFF
--- a/pact/pact-clients/pom.xml
+++ b/pact/pact-clients/pom.xml
@@ -30,13 +30,6 @@
 		</dependency>
 
 		<dependency>
-			<groupId>eu.stratosphere</groupId>
-			<artifactId>pact-examples</artifactId>
-			<version>${project.version}</version>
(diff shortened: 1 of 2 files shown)


carabolic:
Hi,

the problem with the old approach (having pact-examples as an dependency for pact-clients) was that you could not locally test your new examples, if they were part of the pact-examples module, with *.pact.clients.LocalExecutor. Doing so would have produced a circular dependency (pact-clients -> pact-examples -> pact-clients -> Call of Cthulhu). With my change you can now safely write and test new pact-examples without summoning the demons.

Regards,
Christoph

aljoscha:
Merge please

PR #73, opened 2013-09-03
Switch to using objects instead of classes for usercode

aljoscha:
Hi Everyone,
I'm opening this pull request to start a discussion on this new feature. I changed the whole stratosphere stack to use objects instead of classes to pass usercode (i.e. Stubs) around. This means that one now uses
MapContract mapper = MapContract.builder(new TokenizeLine())
            .input(source)
            .name("Tokenize Lines")
            .build();
instead of
MapContract mapper = MapContract.builder(TokenizeLine.class)
            .input(source)
            .name("Tokenize Lines")
            .build();

The initial motivation for this is that I require it for the scala-frontend. I cannot generate classes that would be instantiable by the stratosphere runtime so I need to pass objects to the PACTs. I noticed though that this could also make regular java usercode more concise. The Configuration would not be necessary any more because one can just store the configuration in the stub object and for example have a stub with a constructor where you pass in stuff, i.e.:
MapContract mapper = MapContract.builder(new TokenizeLine("<some-line-ending>"))
            .input(source)
            .name("Tokenize Lines")
            .build();
The line ending would then be stored in the stub, the runtime serializes everything and then when the job is run the runtime deserializes the job and the information is available in every instance of the stub.

On thing that is affected by the change is that all usercode stubs now must be serializable. For this I made base class Stub "Serializable" and also other things such as Value and the derived classes.

Now, discuss away... :D

Aljoscha

P.S. We could then also have
        FileDataSource source = new FileDataSource(new TextInputFormat("ASCII"), dataInput, "Input Lines");
instead of
    FileDataSource source = new FileDataSource(TextInputFormat.class, dataInput, "Input Lines");
    source.setParameter(TextInputFormat.CHARSET_NAME, "ASCII");

### DIFF
--- a/nephele/nephele-common/src/main/java/eu/stratosphere/nephele/fs/Path.java
+++ b/nephele/nephele-common/src/main/java/eu/stratosphere/nephele/fs/Path.java
@@ -24,6 +24,7 @@ package eu.stratosphere.nephele.fs;
 import java.io.DataInput;
 import java.io.DataOutput;
 import java.io.IOException;
+import java.io.Serializable;
 import java.net.URI;
 import java.net.URISyntaxException;
 
@@ -35,7 +36,8 @@ import eu.stratosphere.nephele.util.StringUtils;
  * Names a file or directory in a {@link FileSystem}. Path strings use slash as
  * the directory separator. A path string is absolute if it begins with a slash.
  */
-public class Path implements IOReadableWritable {
+public class Path implements IOReadableWritable, Serializable {
+	private static final long serialVersionUID = 1L;
 
 	/**
 	 * The directory separator, a slash.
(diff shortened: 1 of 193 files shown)


aljoscha:
I now changed it so that the old interface stays intact while adding the possibility to pass objects.

uce:
:+1:

rmetzger:
I tested the pull request successfully with my telecom-usecase stuff. Worked very well.

rmetzger:
Cascading is also passing objects instead of classes. So this approach is not as uncommon in the Hadoop world, as we probably thought.

uce:
Users implementing a stub have to also implement the Serializable interface when passing objects, e.g.:
`public static class TokenizeLines extends MapStub implements Serializable { ... }`

Can't we have the stubs directly implement Serializable instead of the user?

aljoscha:
No, see issue #147

uce:
I totally forgot this... sorry. :sob:

PR #89, opened 2013-09-12
Pact Compiler fix+testcase for SinkJoinerPlanNode

rmetzger:
Hopefully fixed a bug in the pact compiler for branching plans:
    - added a test case provided by Anja Kunkel from HU Berlin
    - my understanding of the bug is the following: During a check if additional pipeline breakers are necessary,
    the compiler checks for the LocalStrategy of a SinkJoinerPlan. The SinkJoinerPlan is a "fake" plan element to group multiple output sinks into one sink (the compiler assumes one root node, this is why a "fake" plan node is required).
    The local strategy of SinkJoinerPlan is initialized as NONE, which means that we just pass through the elements (which makes sence since we are talking about a sink) The initialization of a NONE-local strategy assumes only one input, not two.
    I replaced the LocalStrategy with BINARY_NO_OP, which is a NONE for two input.
    The compiler is now able to do its pipline breaker checks.

### DIFF
--- a/pact/pact-compiler/src/main/java/eu/stratosphere/pact/compiler/PactCompiler.java
+++ b/pact/pact-compiler/src/main/java/eu/stratosphere/pact/compiler/PactCompiler.java
@@ -677,7 +677,7 @@ public class PactCompiler {
 
 			while (iter.hasNext()) {
 				rootNode = new SinkJoiner(rootNode, iter.next());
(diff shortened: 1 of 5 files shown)


aljoscha:
Seems to fix a bug I had with several sinks in a plan. Made my day. :dancers:

PR #90, opened 2013-09-12
Pact Compiler Bug: "Cannot subtract more cost then there is" (no fix, just a testcase)

rmetzger:
I ran into this bug while trying to create a test case for another compiler bug!

I want to focus onto the t-labs usecases right now, so I open this as an unfixed bug.


StephanEwen:
Fixed in 6dcf74f022f4a127260b0e8d52b772fb28fa2249

PR #91, opened 2013-09-12
Pact Compiler: Potential fix for a NullPointerException in SinkJoiner

rmetzger:
I got the following exception during my work on the tlabs usecases.

I'm not sure if this fix is a suitable solution. Since the original plan is quite large, it is not very easy for me to reproduce the bug in a dedicated test case.

```
Exception in thread "main" java.lang.NullPointerException
    at java.util.ArrayList.addAll(ArrayList.java:559)
    at eu.stratosphere.pact.compiler.plan.SinkJoiner.computeUnclosedBranchStack(SinkJoiner.java:81)
    at eu.stratosphere.pact.compiler.PactCompiler$BranchesVisitor.postVisit(PactCompiler.java:1233)
    at eu.stratosphere.pact.compiler.PactCompiler$BranchesVisitor.postVisit(PactCompiler.java:1)
    at eu.stratosphere.pact.compiler.plan.TwoInputNode.accept(TwoInputNode.java:809)
    at eu.stratosphere.pact.compiler.plan.TwoInputNode.accept(TwoInputNode.java:807)
    at eu.stratosphere.pact.compiler.PactCompiler.compile(PactCompiler.java:703)
    at eu.stratosphere.pact.compiler.PactCompiler.compile(PactCompiler.java:547)
    at eu.stratosphere.pact.client.LocalExecutor.executePlan(LocalExecutor.java:79)
```

This PR also contains a little change to show an meaningful exception if a null-input is given into a contract.

### DIFF
--- a/pact/pact-common/src/main/java/eu/stratosphere/pact/generic/contract/SingleInputContract.java
+++ b/pact/pact-common/src/main/java/eu/stratosphere/pact/generic/contract/SingleInputContract.java
@@ -149,6 +149,9 @@ public abstract class SingleInputContract<T extends Stub> extends AbstractPact<T
 	public void accept(Visitor<Contract> visitor) {
 		if (visitor.preVisit(this)) {
 			for(Contract c : this.input) {
+				if(c == null) {
+					throw new IllegalArgumentException("The contract '"+this.getName()+"' has a illegal input (null)");
+				}
 				c.accept(visitor);
 			}
 			visitor.postVisit(this);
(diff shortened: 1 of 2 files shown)


StephanEwen:
I have gone through the code and the only way I can reproduce that error is with graphs that are disconnected, something like

<pre>
 sink        sink
   ^          ^
   |          |
source      source
</pre>

In that sense, this mechanism might actually be used as a way to detect such situations.

StephanEwen:
The logic catches many situations of disjoint data flow graphs, but not all.

We should be able to extend the branch-handling logic to detect it in all cases. Committed a case that is covered and one that is not covered (test deactivated for now).

StephanEwen:
Part one fixed in 21bcc0acd0ffe413f3df60fd6683debe19da2f93

Issue subsumed by #173

PR #92, opened 2013-09-12
Fixed commons-io missing in lib directory + fixed wrong groupId of commons-io

andrehacker:
This fixes the bug that commons-io was missing in the lib directory.

Furthermore, a wrong (outdated) groupId for commons-io was specified in pact-clients (the [old pom file](http://repo1.maven.org/maven2/org/apache/commons/commons-io/1.3.2/commons-io-1.3.2.pom) specifies the relocation). I also switched to the commons-io versions that is used by hadoop (2.1).

### DIFF
--- a/stratosphere-dist/src/main/assemblies/bin.xml
+++ b/stratosphere-dist/src/main/assemblies/bin.xml
@@ -34,12 +34,7 @@
           The previous command did not work because it also excludes libraries that should be in lib, such as commons-io.
 				-->
         <exclude>commons-fileupload:commons-fileupload</exclude>
-        <exclude>org.eclipse.jetty:jetty-continuation</exclude>
-        <exclude>org.eclipse.jetty:jetty-http</exclude>
-        <exclude>org.eclipse.jetty:jetty-io</exclude>
-        <exclude>org.eclipse.jetty:jetty-security</exclude>
-        <exclude>org.eclipse.jetty:jetty-server</exclude>
-        <exclude>org.eclipse.jetty:jetty-servlet</exclude>
+        <exclude>org.eclipse.jetty:jetty-*</exclude>
 			</excludes>
 		</dependencySet>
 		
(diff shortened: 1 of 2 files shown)


rmetzger:
I should note that the bug is not covered by any testcase. 
Since the TaskManagers are started using bash-scripts, we can not test for those bugs using maven (at least not right now. It is possible: http://stackoverflow.com/questions/3491937/i-want-to-execute-shell-commands-from-mavens-pom-xml).

stratosphere-dist/src/main/assemblies/bin.xml
```diff
@@ -28,7 +28,18 @@
...
+				  This is a hardcoded exclude-list containing all libraries that are exclusively used in pact-clients.
+          The previous command did not work because it also excludes libraries that should be in lib, such as commons-io.
+				-->
+        <exclude>commons-fileupload:commons-fileupload</exclude>
+        <exclude>org.eclipse.jetty:jetty-continuation</exclude>
+        <exclude>org.eclipse.jetty:jetty-http</exclude>
+        <exclude>org.eclipse.jetty:jetty-io</exclude>
+        <exclude>org.eclipse.jetty:jetty-security</exclude>
+        <exclude>org.eclipse.jetty:jetty-server</exclude>
+        <exclude>org.eclipse.jetty:jetty-servlet</exclude>
```
rmetzger: Can't we use wildcards here? (org.eclipse.jetty:*) ?

andrehacker: I had the same idea, but I saw that two jetty artifacts are used from hadoop-common, which is used in nephele-hdfs and pact-runtime: jetty and jetty-util.

However, now I see that these two have a different groupId (org.mortbay.jetty) - I will look into this and try to use the wildcards. Good that you asked again!

andrehacker:
Now I use wildcards to exclude all jetty artifacts.

PR #96, opened 2013-09-13
Fixed nephele-config.sh substr command calls on OS X (closes #88)

### LINKED ISSUE #88: Start scripts not working on Mac
The start scripts `bin/start-local.sh` etc. are not working correctly on Mac OS X. 

```
ozone  ./bin/start-local.sh 
expr: syntax error
expr: syntax error
expr: syntax error
expr: syntax error
expr: syntax error
expr: syntax error
expr: syntax error
Starting Nephele job manager
```

I think this was introduced by Issue #60.

The problem seems to be the `expr substr` calls in `nephele-config.sh`. On OS X the manpage for `expr` says the following:

> According to the POSIX standard, the use of string arguments length, substr, index, or match produces undefined results. In this version of expr, these arguments are treated just as their respective string values."

On Linux systems `expr substr` works fine.

uce:
This is issue #88.

On OS X the `expr substr` (sub-)command is not supported. I replaced it with bash's (extended) syntax for string expansions.

If this is not OK an alternative would be to use `uname -s | cut -c1-6`.

### DIFF
--- a/stratosphere-dist/src/main/stratosphere-bin/bin/nephele-config.sh
+++ b/stratosphere-dist/src/main/stratosphere-bin/bin/nephele-config.sh
@@ -19,7 +19,8 @@
 # but the windows java version expects them in Windows Format, i.e. C:\bla\blub.
 # "cygpath" can do the conversion.
 manglePath() {
-    if [ "$(expr substr $(uname -s) 1 6)" == "CYGWIN" ]; then
+    UNAME=$(uname -s)
+    if [ "${UNAME:0:6}" == "CYGWIN" ]; then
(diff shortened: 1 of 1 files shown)


rmetzger:
Does the fixed version also work on cygwin and linux?

uce:
Yes, it works on Linux. Didn't test it on Cygwin though. But it should work, right? Does anybody run Windows and can test it with Cygwin?

rmetzger:
Let's ask Aljoscha tomorrow, he created the cygwin version.

rmetzger:
I did test the script with Windows/cygwin.

PR #102, opened 2013-09-18
Remove TeraSort from examples

uce:
The TeraSort job (one of the example jobs) is currently not working, because of some issues with the range partitioner (see issue #7).

I think we shouldn't have it in the build until this is fixed.


rmetzger:
Thanks for adding the comment!

Issue #76, opened 2013-09-06
Redesign stratosphere.eu website


dimalabs:

rmetzger:
Preview of @aalexandrov's work: http://rmetzger.github.io/ozone/

aalexandrov:
New version of the website is available, minus the input from @ktzoumas , at http://aalexandrov.github.io/ozone. I encourage everybody to check for errors (dead links, inappropriate text, etc.). The homepage will probably be reworked in the coming weeks but this is not time critical.

Once we have the full text for the Stratosphere workshop I'll merge the gh-pages branch with the main repository.

@ktzoumas: Can you do a pull request with your modification of the `events/2013/workshop.html` page to my fork of the ozone repository?

aalexandrov:
Closed by pull request #104.

PR #104, opened 2013-09-19
New version of the Stratosphere website

aalexandrov:
New version of the Stratosphere website using a gh-pages branch.


aalexandrov:
Closes #76.

rmetzger:
The wiki is broken (seems like the css is not loading)

PR #107, opened 2013-09-20
AllReducer does not work in iterations with dop>1 (added test case)

andrehacker:
Compiler throws the following error when I use the AllReducer inside iterations with dop>1:
`eu.stratosphere.pact.compiler.CompilerException: Error: All functions that are part of an iteration must have the same degree-of-parallelism as that iteration.`
The dop is automatically set to 1 for the AllReducer (which absolutely makes sense). I extended Roberts test case to catch this bug.

Maybe anyone can explain why we have this restriction? Due to this restrictions I also had to remove some other optimizations where I defined the dop to be 1 for some contracts manually.

BTW: It would be nice to give a hint in the error message which node has a dop different from iteration (e.g. the name and type).

### DIFF
--- a/pact/pact-compiler/src/test/java/eu/stratosphere/pact/compiler/ReduceAllTest.java
+++ b/pact/pact-compiler/src/test/java/eu/stratosphere/pact/compiler/ReduceAllTest.java
@@ -15,7 +15,7 @@
 
 package eu.stratosphere.pact.compiler;
 
-import static org.junit.Assert.*;
+import static org.junit.Assert.fail;
 
 import org.junit.Test;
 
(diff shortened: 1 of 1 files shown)


StephanEwen:
Th iterations currently have the restriction that all operators have to have the same degree-of-parallelism as the iteration operator. Since All_Reduce produces a final DOP=1 operator, it is rejected.

The rational behind all nodes being of the same DOP is the relationship between nodes and iteration heads, for the distributed aggregators. I assume that an operator with a DOP < Iteration DOP should actually work, so we might be able to relax that condition to allow ALL_REDUCE.

StephanEwen:
I looked through the code and it seems like we can run all operators with DOP < Iteration-DOP.

@andrehacker Can you create a simple test case out of this issue?

andrehacker:
Good to hear! 
The test case is already done - see above (if this is what you mean)

StephanEwen:
The above code only tests whether the compiler complains, not whether it actually works at runtime. We would need a test like https://github.com/stratosphere/stratosphere/blob/master/pact/pact-tests/src/test/java/eu/stratosphere/pact/test/pactPrograms/ComputeEdgeDegreesITCase.java

Is that feasible?

andrehacker:
this makes sense, should be no problem, will do it asap.

andrehacker:
I wrote the test case, but there is a strange behaviour:
If I specify N iterations, the UDF will be called N+1 times. Is this normal?
The last time it is called with an empty iteration. The result of the last call is ignored.

See the code here:
https://github.com/andrehacker/ozone/blob/all-reduce-iteration-test/pact/pact-tests/src/test/java/eu/stratosphere/pact/test/iterative/IterationWithAllReducer.java#L120

uce:
Yes, this is normal.

After the n-th iteration, the iteration head will propagate the termination event to the intermediate and tail tasks. If I'm not mistaken, this results in the n+1th UDF call.

StephanEwen:
Subsumed by #280

PR #108, opened 2013-09-21
Ignore system files when using FileInputFormat

rmetzger:
The FileInputFormat adds files that start with an underscore ( _ ) when the given input file is a directory during the creation of InputSplits.

MapReduce jobs create an empty `_SUCCESS` file and a `_logs` directory. These are not relevant when reading the files.

This error also caused empty InputSplits.
FileInputFormat is the base class for most InputFormats.
Hadoop has a smiliar mechanism, which is more advanced (they have a Filter-system that supports multiple filters to be registered at the file system). My implementation does use the same filter as their FileInputFormat. This approach also allows to chain filters.


fhueske:
Why not make the filter fully configurable instead of static?
Filter on _\* and .\* could be default setting.

rmetzger:
You are right. It would be actually better to make it configurable, so that it is not necessary to extend the existing FileInputFormat. I'm just too lazy to implement it and I guess the feature is not super relevant, since nobody needed it so far.

PR #110, opened 2013-09-23
Upgrade to newer maven surefire (integration tests) version + Fix for failing testcase in ExecutionGraphTest

rmetzger:
I upgraded to the most recent maven surefire version. It shows more detailed error messages / summaries, which is especially useful for travis builds (because we cannot access the surefire-reports on the test machines)
The new version marks errored tests as failed.

The pull request contains a fix for a failing testcase in the ExecutionGraphTest (that was most probably introduced by https://github.com/dimalabs/ozone/pull/53) which was not detected because it errored (NullPtr)


StephanEwen:
Partially manually merged and subsumed by #151

PR #121, opened 2013-10-01
Add an example of using LocalExecutor in WordCount example

aljoscha:
Asterios pointed out that nowhere in the examples do we show that the LocalExecutor exists. So I added a static main method that shows it's usage.


rmetzger:
Great! That is a very good idea.

Once I've finished the "LocalDistributedExecutor", we could also write a comment about it

PR #126, opened 2013-10-04
Generate source attachments for each subproject

rmetzger:
Maven did not generate source attachments for the jars.

This change should activate it for all subprojects. This is an requirement for stable releases in maven central. It is also nice to have, if you want to look up our sources in eclipse.


StephanEwen:
Is this issue fixed with a8228ceb96bfba2b4b9d63ace8c77077a0689a7b ?

PR #135, opened 2013-10-09
Add scala code to mainline

### LINKED ISSUE #118: Add scala-frontend to stratosphere
@aljoscha is about to finish the Scala frontend-implementation.

I'd like to discuss here if we are going to add the frontend into the main ozone-codebase or if we are going to let it stay in the ozone-scala repository.

It is not possible to (git) merge the two repositories, so if we add scala to the main code, we have to copy the files into the repo.

I vote for adding scala into the main repository. This is a major feature of our system. Having the code here would also include it into the testing system (travis).
It is easier for users to check out only one repository.

aljoscha:
This closes #118

### DIFF
--- a/pact-scala/pact-scala-core/pom.xml
+++ b/pact-scala/pact-scala-core/pom.xml
@@ -1,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <artifactId>pact-scala</artifactId>
+    <groupId>eu.stratosphere</groupId>
+    <version>0.4-ozone-SNAPSHOT</version>
+    <relativePath>..</relativePath>
+  </parent>
+
+  <artifactId>pact-scala-core</artifactId>
+  <name>pact-scala-core</name>
+
+  <packaging>jar</packaging>
+
+  <dependencies>
+    <dependency>
+      <groupId>eu.stratosphere</groupId>
+      <artifactId>pact-common</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>eu.stratosphere</groupId>
+      <artifactId>pact-compiler</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+  </dependencies>
+
+  <reporting>
+    <plugins>
+    </plugins>
+  </reporting>
+
+  <build>
+    <plugins>
+    </plugins>
+  </build>
+  
+</project>
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ target
 .project
 .settings
 .cache
+.metadata/*
 *.class
 filter.properties
 tmp/*
--- a/README.md
+++ b/README.md
@@ -91,6 +91,13 @@ Get some test data:
 
 To contribute back to the project or develop your own jobs for Stratosphere, you need a working development environment. We use Eclipse and IntelliJ for development. Here we focus on Eclipse.
 
+If you want to work on the scala code you will need the following plugins:
+  * scala-ide: http://download.scala-ide.org/sdk/e38/scala210/stable/site
+  * m2eclipse-scala: http://alchim31.free.fr/m2e-scala/update-site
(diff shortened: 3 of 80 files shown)


rmetzger:
Very nice. But the tests are disabled, right?

aljoscha:
Yes, they are are disabled. I just wanted to get it out so that people can look at it and experiment with it. The examples will all be ported soon (KMeans and WordCount are converted already) and then I will also add integration tests similar to KMeansITCase and friends.

pact-scala/pact-scala-core/pom.xml
```diff
@@ -0,0 +1,48 @@
...
+      <version>${project.version}</version>
+    </dependency>
+    
+    <dependency>
+      <groupId>eu.stratosphere</groupId>
+      <artifactId>pact-compiler</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    
+    <dependency>
```
rmetzger: This is a duplicate dependency

rmetzger:
Travis found the following:

```
[WARNING] /home/travis/build/dimalabs/ozone/pact-scala/pact-scala-core/src/main/scala/eu/stratosphere/scala/analysis/postPass/GlobalSchemaCompactor.scala:62: warning: match may not be exhaustive.
[WARNING] It would fail on the following input: Some((x: eu.stratosphere.scala.analysis.UDF[?] forSome x not in (eu.stratosphere.scala.analysis.UDF0[?], eu.stratosphere.scala.analysis.UDF1[?,?], eu.stratosphere.scala.analysis.UDF2[?,?,?])))
[WARNING]         val (forwardPos, outputFields) = node.getUDF match {
[WARNING]                                               ^
[WARNING] /home/travis/build/dimalabs/ozone/pact-scala/pact-scala-core/src/main/scala/eu/stratosphere/scala/codegen/UDTGen.scala:46: warning: match may not be exhaustive.
[WARNING] It would fail on the following inputs: BaseClassDescriptor(_, _, _, _), CaseClassDescriptor(_, _, _, _, _), UnsupportedDescriptor(_, _, _)
[WARNING]       val fieldTypes = getIndexFields(desc).toList map {
[WARNING]                                                        ^
[WARNING] /home/travis/build/dimalabs/ozone/pact-scala/pact-scala-core/src/main/scala/eu/stratosphere/scala/codegen/UDTGen.scala:61: warning: match may not be exhaustive.
[WARNING] It would fail on the following inputs: BaseClassDescriptor(_, _, _, _), CaseClassDescriptor(_, _, _, _, _), UnsupportedDescriptor(_, _, _)
[WARNING]       val fieldIds = getIndexFields(desc).toList map {
[WARNING]                                                      ^
[WARNING] warning: there were 36 deprecation warning(s); re-run with -deprecation for details
[WARNING] warning: there were 67 feature warning(s); re-run with -feature for details
[WARNING] 5 warnings found
```

aljoscha:
Ok, I'll remove the duplicate and fix the warnings. One of the travis builds failed with a mysterious error, is this another one for your collection?

rmetzger:
Great!
The error is known, it happens sometimes. I'll add it to the list ;)

aljoscha:
Alright, I fixed the duplicate dependency and got rid of the warnings. :checkered_flag:

rmetzger:
Why did you put pact-scala into the root directory?
Shouldn't it be in /pact/pact-scala or so?

It seems that, except for stratosphere-dist, the directory structure is similar to maven.

rmetzger:
Looks like your Scala code made it successfully into maven's snapshot repo: https://oss.sonatype.org/content/repositories/snapshots/eu/stratosphere/pact-scala-core/0.4-ozone-SNAPSHOT/

aljoscha:
It should probably not be named pact-scala in the first place. Users don't ever see that they are working with PACTs internally so I put it in the top-level.

rmetzger:
why not calling it "scala", with "scala-core", "scala-examples" as sub-projects ?

aljoscha:
Ok, I'll do that tomorrow.

fhueske:
Isn't a maven artifact named 'scala' a bit to unspecific? 
What's wrong with scala-pact? It's a frontend for the Pact programming model, right?

Just my 2 cents...

Btw. I'm having trouble with the scala code in Eclipse (ScalaIDE for Scala 2.10 and scala-maven plugins are installed).

rmetzger:
Well, we have the group id as well. So people hopefully understand that it is only our scala interface. An entry in a pom would look like this:

``` xml
<dependency>
    <artifactId>scala</artifactId>
    <version>0.4</version>
</dependency>
```

Concerning your troubles: you probably need an additional plugin (see README)

fhueske:
OK, that did the trick. Thanks!

aljoscha:
Yes, what @fhueske said was my reason for naming it pact-scala and not just scala. I still would find it strange to have a package called scala...

PR #137, opened 2013-10-09
Ozone-YARN-Integration 

TobiasHerb:
## Description:
- added the 'nephele-yarn' module
  - implemented the 'InstanceManager' interface
- added the 'nephele-common-yarn' module
  - implemented the client API, 'YarnJobClient.java'
  - implemented simple demo job, 'eu.stratosphere.nephele.example.yarn'
- rewritten the 'DiscoveryService.java' (by Daniel)
- added in the 'JobManager.java' an additional execution mode 'YARN'
- TaskManager can now determine IPC/DATA ports, if it is bootstrapped in a yarn container
  - multiple TMs can exist on a single cluster node
- a lot of minor changes
## YARN Integration:

Two resource managers, in concrete MESOS and YARN, were examined that provide the capability to run multiple instances of the data parallel execution framework Stratosphere alongside other applications (e.g. Hadoop, Hyracks) in a cluster/cloud infrastructure to efficiently share common computing resources. After a detailed analysis of both frameworks we believe that YARN has several significant advantages over MESOS which are listed in the following:
- YARN is primarily written in Java. MESOS is written in C++.
- YARN uses simple unix processes as computing resource abstraction. MESOS uses Linux containers that imply a high overhead.
- YARN integrates pluggable resource schedulers.
- YARN is able to directly handle rack and machine locality in resource requests.
- YARN is going to be the basis for Hadoop MapReduce going forward and other frameworks (e.g. Hyracks).

Due to its essential advantages we decided to integrate the YARN cluster manager in the Stratosphere scheduling kernel, instead of the MESOS framework. To understand the integration mechanism of YARN in Stratosphere do we need to depict the core components of both frameworks
## YARN core components:
- ResourceManager (RM): The central component in YARN. The RM exists as singleton instance in the infrastructure and manages all available computing resources. It acts as pure resource negotiator for all concurrent running systems.
- NodeManager (NM): On every node in the infrastructure runs a NM instance that runs a liveness protocol and do the concrete container allocation with permission of the RM.
- ApplicationMaster (AM): The AM exists per running job and is responsible for allocating/scheduling worker containers (by requesting the RM) and monitoring the job lifecycle.
- YARN Client: The client submits the job in form of starting an AM instance and handing over the concrete executable job. It also performs monitoring of the running jobs.
## Stratosphere core components:
- JobManager (JM): The JM is the central component in Stratosphere’s execution layer Nephele and is responsible for allocating resources (part of a subcomponent called InstanceManager) and scheduling/monitoring submitted jobs.
- TaskManager (TM): The TM has similar responsibilities as the NodeManager in YARN and also executes jobs in cooperation with other TMs.
- Job Client: Submits a job and monitors the job execution.
## Nephele-YARN Interaction:

In the integration of YARN in Nephele we operate YARN as a a meta-level framework that encloses certain components of Nephele. This approach has the advantage that the current code base of Nephele is not dramatically changed and needs only to be extended. A further and more important advantage is the clean process isolation among multiple concurrent Stratosphere instances and other applications (e.g. Hadoop, Hyracks).

The sequence diagram depicts the basic interaction:
yarn_nephele_seqdia

Interaction Description: The Client establishes a YARN context and requests an ApplicationMaster container. Inside the allocated container, a new instance of Stratosphere’s JobManager is created. The client submits a Stratosphere Job to the JobManager instance, which delegates the resource requests to its own InstanceManager subcomponent, which is a YARN-based implementation of the Nephele InstanceManager interface. The resource requests are then delegated from there to the YARN ResourceManage node, that actually allocates the required containers (if available). Inside each allocated container a Stratosphere TaskManager is bootstrapped which receives and executes tasks of the submitted job from the the JobManager.

yarn_nephele_component_interaction

1.) Client sends a request to the YARN Resource Manager to create an Application Master (AM)
2.) ResourceManager (RM) creates the AM container. The JobManager is bootstrapped inside the AM. (via a shell command that was defined in the Client-RM request). The JobManager is started in YARN mode, with the YARN InstanceManager.
3.) Client sends a Job to the JobManager. The JobManager determines the resource needs.
4.) The JobManagers calls the RM (via InstanceManager.requestInstance) and allocated the determined number of containers.
5.) RM starts the YARN containers. The TaskManagers are bootstrapped inside these containers. The TaskManager have to determine their RPC and data ports (because containers can be scheduled to the same node).
6.) The boostrapped TaskManagers are registered by the JobManager via the first heart-beat signal.
7.) The JobManager can begin to schedule the execution verticies on the TaskManagers.

### DIFF
--- a/nephele/nephele-clustermanager/src/main/java/eu/stratosphere/nephele/instance/cluster/AllocatedSlice.java
+++ b/nephele/nephele-clustermanager/src/main/java/eu/stratosphere/nephele/instance/cluster/AllocatedSlice.java
@@ -68,7 +68,7 @@ class AllocatedSlice {
 	public AllocatedSlice(final ClusterInstance hostingInstance, final InstanceType type, final JobID jobID,
 			final long allocationTime) {
 
-		this.allocationID = new AllocationID();
+		this.allocationID = AllocationID.generate();
 		this.hostingInstance = hostingInstance;
 		this.type = type;
 		this.jobID = jobID;
--- a/nephele/nephele-clustermanager/src/main/java/eu/stratosphere/nephele/instance/cluster/ClusterManager.java
+++ b/nephele/nephele-clustermanager/src/main/java/eu/stratosphere/nephele/instance/cluster/ClusterManager.java
@@ -701,7 +701,7 @@ public class ClusterManager implements InstanceManager {
 	 */
 	@Override
 	public synchronized void reportHeartBeat(final InstanceConnectionInfo instanceConnectionInfo,
-			final HardwareDescription hardwareDescription) {
+			final HardwareDescription hardwareDescription, final String taskID) {
 
 		ClusterInstance host = registeredHosts.get(instanceConnectionInfo);
 
--- a/nephele/nephele-clustermanager/src/test/java/eu/stratosphere/nephele/instance/cluster/ClusterManagerTest.java
+++ b/nephele/nephele-clustermanager/src/test/java/eu/stratosphere/nephele/instance/cluster/ClusterManagerTest.java
@@ -165,7 +165,7 @@ public class ClusterManagerTest {
 
 			// Although the hardware description indicates an instance of type "small", the cluster manager is supposed
 			// to take the user-defined instance type "high"
-			cm.reportHeartBeat(ici, hardwareDescription);
+			cm.reportHeartBeat(ici, hardwareDescription, null);
 
 			instanceTypeDescriptions = cm.getMapOfAvailableInstanceTypes();
 			assertEquals(3, instanceTypeDescriptions.size());
@@ -195,7 +195,7 @@ public class ClusterManagerTest {
 
 			ipAddress = "192.168.198.3";
 			ici = new InstanceConnectionInfo(InetAddress.getByName(ipAddress), ipAddress, null, ipcPort, dataPort);
-			cm.reportHeartBeat(ici, hardwareDescription);
+			cm.reportHeartBeat(ici, hardwareDescription, null);
 
 			instanceTypeDescriptions = cm.getMapOfAvailableInstanceTypes();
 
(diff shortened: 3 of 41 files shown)


rmetzger:
Open Issues:
- Memory issues (@TobiasHerb)
- Integration into Pact job execution (@rmetzger)
- Integration tests issues.(@rmetzger)

Long-term enhancements:
- Caching of Yarn Containers

rmetzger:
If anyone is interested in continuing this work, you can find the code where I left of in this branch: https://github.com/rmetzger/stratosphere/tree/yarn_rebased_remove_discovery_rebased (I used this pull request as a basis for a "deep" yarn integration (changing Stratosphere's internals))

The code there is well-integrated into the overall stratosphere build system (separate maven packages, integration into build profiles, clean/well-integrated run-scripts); 
I changed the job submission so that one Container comes up first (the compiler has to wait until yarn deployed a container) and did quite some debugging on our cluster.

Current Status:
- I tested the code in normal operation (not on yarn) and it seems to work
- the code expects hadoop 2.0.5-alpha, not the stable (and officially released): 2.2.0
- Yarn works on my local machine (with only one container allocation)
- Yarn does not work on our cluster. (It is able to start the JobManager and TaskTracker and to start processing the job but then it fails because it somehow re-registers the TaskTracker. As I said, the whole TaskTracker registration mechanism is new.
  In order to get it to run on the cluster, it expects that there is a shared file system (I'm  sure it will not work if you just invoke it on one machine.)

Please refer to https://github.com/stratosphere/stratosphere/pull/312 for the new Yarn integration (which does basically not change Stratosphere itself, it just adds a control-layer to execute regular Stratosphere code). The code there takes care of all the deployment aspects (code distribution etc.).

I suggest to close this pull request for now. (The work here can still be used if someone wants a deeper integration and a more on-demand behavior of Stratosphere on yarn.)

StephanEwen:
Subsumed by #312

PR #152, opened 2013-10-11
Scala examples for Triangle Enumeration

fhueske:
Added first versions of the Scala examples for triangle enumeration.
The examples include:
- Annotation of edges with the corresponding vertex degrees (+ unique projection of vertexes and duplicate removal)
- Enumeration of triangles on edges with degrees.

The examples perform the same job as the corresponding Java versions. However, they use different strategies (joins instead of reduce in some places). Exactly following the Java strategies didn't work for a Scala novice ;-)
However, this might be changed later.


fhueske:
I did pay attention and included commits into the pull request that did not belong here.
A new pull request was opened #153.

PR #153, opened 2013-10-11
Scala examples for Triangle Enumeration

fhueske:
Added first versions of the Scala examples for triangle enumeration.
The examples include:

```
Annotation of edges with the corresponding vertex degrees (+ unique projection of vertexes and duplicate removal)
Enumeration of triangles on edges with degrees.
```

The examples perform the same job as the corresponding Java versions. However, they use different strategies (joins instead of reduce in some places). Exactly following the Java strategies didn't work for a Scala novice ;-)
However, this might be changed later.


aljoscha:
Looks good. Could you also add tests similar to the KMeansIterativeITCase and friends?

aljoscha:
Was it hard to figure out how to do stuff, by the way?

fhueske:
The solution I committed was easy to implement. I also tried to follow more closely the Java version but could not do it.

I have a couple remarks, that will share with you after the weekend. But overall a very nice work!

I suggest that the other examples are also done by people that haven't used the Scala frontend yet. That should get you more feedback.

aljoscha:
Yes, I agree. Plus it's less work for me ... :smile:

fhueske:
I prepared integration tests for the example jobs but they fail due to some missing Scala class. This seems to be some integration issues. The example jobs themselves are working.
I'm waiting for a fix/hint from @aljoscha to merge them together with the tests.

fhueske:
Integration tests should work now. 
Let's wait for Travis and merge if all tests pass.

aljoscha:
@fhueske You added the additional dependency on scala-reflect. Was that what fixed the weird TypeCreator error we saw?

fhueske:
@aljoscha Yes, that solved the problem. 
I did not further investigate the issue after the bug was fixed. Do you see any problems with the added dependency?

aljoscha:
Nah, I don't see a problem with that.

PR #155, opened 2013-10-11
Fix bug with user-code wrappers

aljoscha:
The problem was that the user code wrapper was not deserialized from the
config using the user-code class loader. Now we jave a custom
ObjectInputStream that uses the user-code class loader to resolve
classes.


aljoscha:
@rmetzger Could you please verify that this does indeed fix the bug you had.

rmetzger:
Yes. I'm going to do that today.

rmetzger:
Fix verified! Thank you @aljoscha!

PR #156, opened 2013-10-13
Improve pom-tooling scripts

rmetzger:
The pom-generate script now allows to specifiy a filename for the generated pom .This is useful if one wants to overwrite the existing pom with the other profile as default!


StephanEwen:
Manually merged in 02ff4805b845a1b5b20fb0a008304307cd600033

PR #157, opened 2013-10-13
Remove commons-io from pact-clients.

rmetzger:
- it does not seem to be used there
- It is in conflict with commons-io from hadoop yarn (which uses commons-io 2.1)
- The coordinates of the artifact changed (and Andre and I forgot to rename it)
- My Telecom use-cases were breaking because of this: My code expected a method from commons-io:2.1 but 1.3.2 was loaded. I had no troubles loading my code with pact-client.sh, I did, however, not check if the pact webinterface still works. But eclipse did not show any errors after removing commons-io.

### DIFF
--- a/pact/pact-clients/pom.xml
+++ b/pact/pact-clients/pom.xml
@@ -69,15 +69,15 @@
 			<type>jar</type>
 			<scope>compile</scope>
 		</dependency>
-
+		<!-- commons-io is required by commons-fileupload
+			See http://commons.apache.org/proper/commons-fileupload/dependencies.html
+			and https://github.com/dimalabs/ozone/pull/157 -->
 		<dependency>
-			<groupId>org.apache.commons</groupId>
+			<groupId>commons-io</groupId>
 			<artifactId>commons-io</artifactId>
-			<version>1.3.2</version>
-			<type>jar</type>
+			<version>2.1</version>
 			<scope>compile</scope>
 		</dependency>
-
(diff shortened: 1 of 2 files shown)


StephanEwen:
If I recall correctly, commons-io was indeed a dependency from the web-interface. I think either jetty or commons-fileupload required it.

aalexandrov:
I had to configure this for a different project last week, commons-fileupload seems to require commons-io: http://commons.apache.org/proper/commons-fileupload/dependencies.html.

We should check whether jar upload still works with the new configuration.

rmetzger:
Okay. You guys are right. The dependency is required.
The fileupload works fine even without the dependency, because other projects (hadoop) depend on commons-io 2.1, so the jar is in the classpath.
I'll see if everything works with commons-io 2.1, so that we have only one version of it for the whole project. The problem is that our yarn build probably pulls a different version of commons-io.

aalexandrov:
Then I suggest to define a `commons-io.property` with different values for hadoop_v1 and hadoop_yarn as part of the maven profiles and use it in `pact-clients`, or simply define an appropriate entry in the `<dependencyManagement>` section in the root pom profiles. 

It's a hack but it will solve the problem, and we should not spend too much time on such issues.

StephanEwen:
Manually merged in 7e0cebf36f0051ac1adf62abcbc1e171df2ea06d

PR #161, opened 2013-10-14
Serializer fixes

### LINKED ISSUE #160: Scala Frontend: PactList Serialization Bug
I tried to reimplement the Java version of the ComputeEdgeDegree example as a Scala Pact program.
The job requires to return multiple records from a reducer (one per input record). However, currently the Scala reducer can only return a single record. I tried to solve this issue by returning a Scala list from the Reducer that contains all emitted records and to flat out that list using a flatMap. The Scala list seems to be correctly translated into the PactList type, but during its serialization a NullPointerException is raised.

The code of the failing Scala job is here:

https://github.com/fhueske/ozone/blob/fc914eadc3507cc27a2eb4891b3f491df0cb0e7a/pact-scala/pact-scala-examples/src/main/scala/eu/stratosphere/scala/examples/graph/ComputeEdgeDegrees2.scala

aljoscha:
This fixes #160

### DIFF
--- a/pact-scala/pact-scala-core/src/main/scala/eu/stratosphere/scala/codegen/SerializeMethodGen.scala
+++ b/pact-scala/pact-scala-core/src/main/scala/eu/stratosphere/scala/codegen/SerializeMethodGen.scala
@@ -202,7 +202,7 @@ trait SerializeMethodGen[C <: Context] { this: MacroContextHolder[C] with UDTDes
       val add = Apply(Select(target, "add"), List(value))
       val addNull = Apply(Select(target, "add"), List(mkNull))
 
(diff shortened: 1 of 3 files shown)


aljoscha:
@fhueske I checked it with your code example in #160, it does indeed work now.

fhueske:
:-D Thanks for the fix!

aljoscha:
@rmetzger The travis build for this pull request is done and all tests passed as can be seen when clicking on the link here. Github, however, does not seem to notice. Could this be because of the move to stratosphere/stratosphere?

rmetzger:
Okay, I'm going to ask the Travis guys tomorrow evening. (I don't have time sooner)

rmetzger:
@aljoscha it seems that the Travis integration is working again.

PR #162, opened 2013-10-14
Multiple joins with solutionset

vasia:
A set of changes to allow more than one joins with the Solution Set.


StephanEwen:
Cleaned up and fixed as of f863fa345a08ee908e936a02995767fdf1405ebb and fdaadf41c4c6c14f56557aa601fbdd4e0944386c

Issue #113, opened 2013-09-24
Configuration setup


uce:
I find it confusing to have configuration parameters set in different places. Besides the XML files in `conf` we also have configuration parameters in `bin/nephele-config.sh`. We could unify this by providing a script `conf/nephele-env.sh` which sets the respective environment variables and is included by `bin/nephele-config.sh`.

Any ideas or comments on this?

I would also appreciate a general discussion about our configuration setup. My main question is if we want to keep it in XML? What could be alternatives?

fhueske:
+1 for configuration in a single place.

I think the XML config is fine. It is easy to understand by humans and can be parsed or generated by machines (given an XML parser).

asteriosk:
Yes, it should be kept in a single place - it's always preferable.

The XML file is ok though: the current configuration is pretty flat
(property that has key and value) isn;t this an overkill? Simple key=value
should also work. Unless the XML is there for a specific reason, I prefer
simple, .ini-style configuration files - easier to edit and add new values.
Plus, you don't need an XML parser to do this simple thing (less
dependencies in the project): you read the conf file by line and you split
on the first '='. Then you trim the two resulting strings to remove
possible space. Done.

What is more, .ini style config files are easy to parse from shell scripts.

So, if someone finds it easy to replace the whole XML story by simple files
(named .conf or smth), I would be happy to get rid of XML. If this is an
overkill, let's keep what's already working.

rmetzger:
+1 single file

I'd like to add support for other file formats, such as YAML or JSON. I think it should be easily possible to support multiple formats.

One has to add 4 lines for a single stupid key-value pair

``` xml
<property>
                <key>jobmanager.rpc.address</key>
                <value>localhost</value> 
</property>
```

compared to (YAML)

``` yaml
jobmanager.rpc.address: localhost
```

Yaml also supports nesting ( See https://github.com/dropwizard/dropwizard/blob/master/dropwizard-example/example.yml ) and everything else one wants ;)

See also this discussion: http://stackoverflow.com/questions/1925305/best-config-file-format

We could unify all configuration into ONE file. Even heapsize and other jvm related configuration. Since the format for yaml is so simple, it is possible to parse it using regular bash tools!

uce:
Although XML is indeed easy to parse and understand, I find it way to verbose and repetitive (for humans).

I would go with @rmetzger's YAML proposal. :+1:

andrehacker:
+1 singe config directory.
[...]

PR #163, opened 2013-10-15
Unified config and added YAML (syntax) support (#113)

uce:
## conf/stratosphere-conf.yaml

Unified config to `stratosphere-conf.yaml`, which replaces the old `nephele-user.xml` and `pact-user.xml` and is also read by `bin/nephele-config.sh`.

The new file is not a 1:1 copy of the old two. Instead I removed some stuff I thought was not that important:
- instance configuration: I don't think anyone was using this other than the default instance? 
- multicast for broadcast: The comments for this feature said "use at own risk" :cake: 
- profiler impl classes: We only have two implementation for the profilers, didn't make sense to point to them in the config
- jobclient.polling.interval: Adjusted the system default from 5 to 2 (which was our config default before)
- pact.runtime.fs_timeout: Default is 0, which means wait forever (which should be ok as long as there are no problems)

@asteriosk, @ktzoumas I think we should add a configuration overview on the website (like http://spark.incubator.apache.org/docs/latest/configuration.html)

@StephanEwen: What about the multicasting (for broadcast) feature? Is it stable? Does anybody ever use it? Do we want to remove it?

The old XML files still work and could also be combined with the YAML file (where YAML overwrites keys in XML), but I think it is ok this way. Also note that the YAML config reader in `GlobalConfiguration` only uses YAML syntax for `key: value` mappings and nothing more.
## bin/nephele-config.sh

The bash script can also parse key-value pairs from the config, for example:
- JM heap size with `jobmanager.heap.mb`
- TM heap size with `taskmanager.heap.mb`

There are also some other things, which were supposed to be directly edited in the bash scripts before (see `KEYS_*` in `bin/nephele-config.sh`).

The defaults and environment variables are still the same, so there should not be any bad surprises.


rmetzger:
Very nice. Even with Testcases. The new configs are nice to read.

StephanEwen:
Manually merged in 7e0cebf36f0051ac1adf62abcbc1e171df2ea06d

PR #164, opened 2013-10-15
Added local strategy SORT to DataSinkTask and extended DataSinkTaskTest

### LINKED ISSUE #146: Sorted output not working
The Pact API allows to specify sorted output. 
This feature is currently not working due to open issues with range partitioning.
However, the user is not notified that this is not working.

I suggest that we address this issue as follows:
- As a quick fix, add a compiler exception for jobs that request output sorting.
- As a partial solution, enable locally sorted output (and fully sorted on sinks with DOP 1).
- Finally, enable sorted output via range partitioning.

fhueske:
Fixes #146 and adds support for locally sorted output. Globally sorted output is not supported yet.
This pull request only adds the local strategy SORT to the DataSinkTask and extends the DataSinkTaskTest.
The Pact optimizer had already set all required task configurations.

@StephanEwen Please check that the canceling logic and tests are OK.


StephanEwen:
Manually merged in e51d6a3a3dc3a026e276c1df9d386ab411861b46

PR #167, opened 2013-10-16
Added getUserCodeClass to UserCodeWrapper

mleich:
This pull request adds getUserCodeClass () to UserCodeWrapper to allow for quick access to the class.

Sopremo and Meteor make heavy use of reflection and often require access to the class of the user code.
Currently, the UserCodeWrapper only allows to return instances of the user code, not the class.
It would be possible to just call getClass() on the user code object, however this leads to unnecessary instantiations of the user code if it is provided as a class, not as object.


mleich:
This change would also make it easier to adapt ozone-testing to stratosphere 0.4-ozone

fhueske:
Looks good to me. 
Shouldn't be a problem as it only extends the interface and does not interfere with any existing code.
Will merge your request.

PR #175, opened 2013-10-18
Change name in base pom and version

aljoscha:
End the version name craziness and go back to having 0.4-SNAPSHOT as
version number. Also change the name of the toplevel pom back to
stratosphere.


rmetzger:
Looks good.

StephanEwen:
Manually merged in 2d44c7d1239e581360405435455a5ecdd36c920b

PR #179, opened 2013-10-18
Fix stratosphere-dist to include the scala jars

aljoscha:
Also fix an annoying bug that made tests fail because a dependency on
scala-reflect was wrongly in the "test" scope and also add an example of
how to create a PlanAssembler for a scala job in WordCount

### DIFF
--- a/pact-scala/pact-scala-examples/src/main/scala/eu/stratosphere/scala/examples/wordcount/WordCount.scala
+++ b/pact-scala/pact-scala-examples/src/main/scala/eu/stratosphere/scala/examples/wordcount/WordCount.scala
@@ -15,35 +15,39 @@ package eu.stratosphere.scala.examples.wordcount
 
 import scala.Array.canBuildFrom
 import eu.stratosphere.pact.client.LocalExecutor
-import eu.stratosphere.pact.common.`type`.base.PactInteger
-import eu.stratosphere.pact.common.`type`.base.PactString
-import eu.stratosphere.scala.Args
 import eu.stratosphere.scala.DataSource
 import eu.stratosphere.scala.ScalaPlan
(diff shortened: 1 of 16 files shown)


rmetzger:
Looks good

pact-scala/pact-scala-examples/src/main/scala/eu/stratosphere/scala/examples/wordcount/WordCount.scala
```diff
@@ -39,6 +40,12 @@ object RunWordCount {
   }
 }
 
+class WordCountDescriptor extends PlanAssembler {
+  override def getPlan(args: String*) = {
+    new WordCount().getPlan(args(0), args(1))
```
StephanEwen: Does WordCount actually execute with a DOP > 1? It seems that the Args are parsed, but a specified DOP option is never used.

stratosphere-dist/src/main/assemblies/bin.xml
```diff
@@ -138,6 +138,19 @@
...
+		<fileSet>
+			<!-- copy jar files of pact job examples -->
+			<directory>../pact-scala/pact-scala-examples/target</directory>
+			<outputDirectory>examples/pact-scala/</outputDirectory>
+			<fileMode>0644</fileMode>
+			<includes>
+				<include>*.jar</include>
+			</includes>
+			<excludes>
+				<exclude>pact-examples-${project.version}.jar</exclude>
```
StephanEwen: Does that exclude the wrong project jar file? Seems it should be called "pact-scala-examples-${project.version}.jar"

aljoscha:
Right as always. :D I'll add the correct exclude and add DOP as a parameter.

aljoscha:
I think this badboy should be ready for merge now. Care to have another look?

Issue #143, opened 2013-10-10
Simple Webinterface for Jobmanager


rmetzger:
assigned to Markus Holzemer
Simple jetty server running inside the Jobmanager.

("nice to have" - for release)

rmetzger:
Hello guys,

Markus started to work on this task today.
We need to specify how we would like to have the webinterface.
I suggest that we orient us on the JobTracker interface of Hadoop. 
So a list of waiting, running and completed jobs.
An interface to view logfiles.
A very basic progress overview.

StephanEwen:
There is no notion of waiting jobs right now, so I guess we should go with the running ones (most likely only one most of the time) and then the completed ones.

I think that hadoop is using a dedicated server for job histories, because in YARN, the jobmanager lives only for a single job typically.

andrehacker:
Regarding the logfiles I think it is very important to be able to view the logs for a specific job only.
Logs include the things written to log (from the system or UDF) and std-out (from UDF).

I can confirm what stephan writes about yarn, there is a separate job history daemon. They even added a log aggregation feature called "yarn.log-aggregation-enable" to copy all aggregated logs for a job to hdfs, since log analysis is such a common task.
Although yarn has a different architecture, they do offer such log-per-job web interface and it is a good orientation.

fhueske:
I see two main use cases for the web interface:
1) Progress monitoring of running job
- Show status of currently running tasks
- How many subtasks of each task are in which status (scheduled, started, running, finishing, finished, ...)
- Dependencies of tasks (Which tasks is input of which task (should be displayed as a graph in the end, but right not some text format might be sufficient)

2) Performance debugging (of finished jobs)
- Provide log files
- Provide timestamps / timeline visualization of subtasks (when did which subtask finish, which tasks a stragglers, on which node did tasks run, etc.)
- this requires to save all information, either in a tmp file or some database like Derby or SQLite
- can be later extended to work on currently running job

In my opinion, progress monitoring is easier implement and more important right now. Performance debugging can be added later but should be considered when designing the infrastructure of the web interface.

Btw. AFAIK Nephele is actually queuing jobs and gives out at least the job name and job ID of waiting jobs. But I am not sure which additional information is available (e.g., JobGraph or ExecutionGraph)

markus-h:
This issue can be closed I think.

fhueske:
Agreed.

PR #185, opened 2013-10-21
Information Frontend for JobManager

markus-h:
First draft of a webinterface (info server) for the JobManager. At the moment the info server is only started when using the shell script for starting the job manager (i.e. start-local.sh).
The default URL to access the info server is http://localhost:8081

Currently the info server gives an overview over all running jobs and tasks. Dependend Tasks are connected. For sake of human readability I restricted the shown states in the overview to:
started = CREATED || SCHEDULED || ASSIGNED || READY || STARTING
running = RUNNING
finished = FINISHING || FINISHED || CANCELING || CANCELED
failed = FAILED

Since this is only a first outline I'd appreciate your feedback!


uce:
Some screenshots: 

screen shot 2013-10-21 at 2 40 52 pm
screen shot 2013-10-21 at 2 43 59 pm
screen shot 2013-10-21 at 2 44 05 pm

Can you say what is further on the roadmap? I couldn't find an issue where we have a list of what we want. Seeing what is running right now is definetly nice to have.

I would like:
- List of finished job with runtimes. I think this will need some further changes in the JobManager to keep track of finished jobs, right?
- Log file viewer for debugging

fhueske:
Very nice first draft! I like the edges to show the task dependencies!

Here are my comments:
1. The ID of tasks and subtasks is not necessary, in my opinion.
2. Another column with the total number of subtasks would be nice (= sum of all for other columns). 
3. Auto reloading the page on event changes or in regular intervals (configurable?)
4. Having a graphical bar in each cell that shows how many of the total number of tasks are in this state would be easier to parse than reading numbers. Of course, the number should be displayed as well.
5. Showing the history of executed jobs as Ufuk suggested. This requires to store the history of all tasks which will be necessary in any case if the interface should also help debugging jobs.
6. Show for each subtask the timestamps of the state changes.
7. Displaying the log and std-out files of JobManager und TaskManagers would be very nice. Should be easy for JobManager logs if we assume that the webserver runs on the same node. For TaskManager logs and std-outs this will be more complicated, since these need to be transferred to the JM node.
8. The edges between tasks could encode the type of channel (network, in-memory) and the distribution pattern (forward, bipartite) maybe by color and dashing? Not sure if this information is available through the JM management interface. 

Order by importance and/or assumed complexity ;-)

rmetzger:
Thank you for your pull request!
@uce This is the issue that defines the task: https://github.com/stratosphere/stratosphere/issues/143

I think we should keep it simple first and show only the JobManager log. Transferring logs from the TaskManager with the JobManager needs two additional threads (on each side) plus the network traffic.

markus-h:
Thanks for your comments! I think I'll do the easy changes and extensions in this pull request, so please don't accept it yet.

fhueske:
Two other small things, that would help to check the status of a running job with one look.
- Split the finished state into finished and canceled.
- Color the status columns according to the colors of the SWT frontend (Color codes are somewhere in the source code of nephele-visualization)
  - Starting -> Gray
  - Running -> Green
  - Finished -> Bĺue
  - Canceled -> Orange
  - Failed -> Red

physikerwelt:
I tried to make a screencast... setup stratosphere and start wordcount job... I managed to demonstrate on a macbook pro (using my own git repo that is faster than github) within <5 minutes. The major drawback is that after submitting the job the user is somehow lost and has to look into the filesystem manually.
I see this patch as one possible solution to make a screenscast that I'd like to publish. The other alternative would be to find a way to x-forward the nephele visualisation from the demo vm to the host computer. The latter solution would require a little more expertise from the end user, since he must manage to use xforward to display the visualisation.
Which mehtod woud you prefere for a screenscast?

rmetzger:
That's really cool!
I would prefer to use the new web-interface. @markus-h is going to change some design details tomorrow morning, so you should probably wait for 12 hours ;)

Concerning the nephele visualization: You don't necessary need to use x-forwarding.
a) you can start Stratosphere on your Mac Book! One of the core developers is using OS X (@uce), so I'm quite confident that it works on OS X as well (except that you have to manually add a dependency to get Nephele Visualization running on OSX :( )

b) you can connect from a local stratosphere binary into a remote (considering your VM as remote) stratosphere JobManager.
Set the jobManager address in the configuration on your local stratosphere to the ip:port of your VM's JobManager.

nephele/nephele-management/src/main/java/eu/stratosphere/nephele/managementgraph/ManagementGroupVertex.java
```diff
@@ -417,4 +420,76 @@ public void write(final DataOutput out) throws IOException {
...
+		json += "\"groupvertexid\": \"" + this.getID() + "\",";
+		json += "\"groupvertexname\": \"" + this.getName() + "\",";
+		json += "\"numberofgroupmembers\": \"" + this.getNumberOfGroupMembers() + "\",";
+		json += "\"groupmembers\": [";
+		
+		// Count state status of group members
+		Map<ExecutionState, Integer> stateCounts = new HashMap<ExecutionState, Integer>();
+		
+		// initialize with 0
+		for (ExecutionState state : ExecutionState.values())
```
rmetzger: use our loop-style 
for() {
//
}

physikerwelt:
@rmetzger I won't have a chance to start before 4 pm tomorrow. Most probably I'll care about the screencasts on monday....

nephele/nephele-server/resources/web-docs-infoserver/css/nephelefrontend.css
```diff
@@ -0,0 +1,223 @@
+*
```
rmetzger, line 1: I think the css, images etc. are all copied from the pact-webfrontend, right?

We should find a way to keep them only once so that we can modify the layout in one central point.

nephele/nephele-server/src/main/java/eu/stratosphere/nephele/jobmanager/JobManager.java
```diff
@@ -1169,5 +1180,25 @@ public InputSplitWrapper requestNextInputSplit(final JobID jobID, final Executio
...
+	 * 
+	 * @param config
+	 */
+	public void startInfoServer(Configuration config) {
+		// Start InfoServer
+		try {
+			int port = config.getInteger(ConfigConstants.JOB_MANAGER_WEB_PORT_KEY, ConfigConstants.DEFAULT_WEB_FRONTEND_PORT);
+			server = new WebInfoServer(config, port);
+			server.start();
+		} catch (IOException ioe) {
```
rmetzger: I think one `catch (Exception e)` is enough since they do exactly the same and IO exception is a sublcass of Exception.

nephele/nephele-server/src/main/java/eu/stratosphere/nephele/jobmanager/web/JobmanagerInfoServlet.java
```diff
@@ -0,0 +1,141 @@
...
+				wrt.write("\"time\": " + jobEvent.getTimestamp()+",");
+				
+				// Serialize ManagementGraph to json
+				wrt.write("\"groupvertices\": [");
+				boolean first = true;
+				for(ManagementGroupVertex groupVertex : jobManagementGraph.getGroupVerticesInTopologicalOrder()) {
+					if(first) {
+						first = false;
+					} else {
+						wrt.write(","); }
```
rmetzger: style ;)

nephele/nephele-server/src/main/java/eu/stratosphere/nephele/jobmanager/web/JobmanagerInfoServlet.java
```diff
@@ -0,0 +1,141 @@
...
+			resp.getWriter().print(t.getMessage());
+			if (LOG.isWarnEnabled()) {
+				LOG.warn(StringUtils.stringifyException(t));
+			}
+		} finally {
+			if (jmConn != null) {
+				try {
+					RPC.stopProxy(jmConn);
+				} catch (Throwable t) {
+					System.err.println("Could not cleanly shut down connection from compiler to job manager");
```
rmetzger: Use LOG instead of `System.err`.

physikerwelt:
just a very first alpha version of a screencast (we probably need text and music)
http://streaming.dima.tu-berlin.de/podcast/mschubotz/strat-inst-01.mov (unfortunately some debian repo seems to be inresposinve so I had to re trigger vagrant provision manually)

http://streaming.dima.tu-berlin.de/podcast/mschubotz/strat-wordcount.mov

nephele/nephele-common/src/main/java/eu/stratosphere/nephele/configuration/ConfigConstants.java
```diff
@@ -74,6 +74,29 @@
...
+
+	/**
+	 * The key for Stratosphere's base dir path
+	 */
+	public static final String STRATOSPHERE_BASE_DIR_PATH_KEY = "stratosphere.base.dir.path";
+	
+	/**
+	 * The key for the config parameter defining port for the pact web-frontend server.
+	 */
+	public static final String JOB_MANAGER_WEB_PORT_KEY = "jobmanager.web.port";
```
rmetzger, line 89: Can you add this port to the config file?

markus-h:
I think I'm ready for the pull. I incorporated most of the feedback. I`ll to further extensions (like archive or debugging function) in a new request.

rmetzger:
Great. Thank you. I think @StephanEwen will merge the request.

nephele/nephele-server/src/main/java/eu/stratosphere/nephele/jobmanager/web/JobmanagerInfoServlet.java
```diff
@@ -0,0 +1,108 @@
...
+	 */
+	private final JobManager jobmanager;
+	
+	public JobmanagerInfoServlet(JobManager jobmanager) {
+		this.jobmanager = jobmanager;
+	}
+	
+	/*
+	 * (non-Javadoc)
+	 * @see javax.servlet.http.HttpServlet#doGet(javax.servlet.http.HttpServletRequest,
```
rmetzger: We have the policy to remote all `(non-Javadoc)` if we run over them during development. (We also remove all `@author` tags).
So you should not introduce them ;)
Remove the whole comment. every good IDE is able to find the correct comment.
In line 25 is an empty comment.

nephele/nephele-server/src/main/java/eu/stratosphere/nephele/jobmanager/web/JobmanagerInfoServlet.java
```diff
@@ -0,0 +1,108 @@
...
+	 * javax.servlet.http.HttpServletResponse)
+	 */
+	@Override
+	protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws ServletException, IOException {
+		
+		try {
+			List<RecentJobEvent> recentJobs = jobmanager.getRecentJobs(); //jmConn.getRecentJobs();
+			ArrayList<RecentJobEvent> jobs = new ArrayList<RecentJobEvent>(recentJobs);
```
rmetzger: Why are you copying the the List from an Array list into an Array list?

nephele/nephele-server/src/main/java/eu/stratosphere/nephele/jobmanager/web/LogfileInfoServlet.java
```diff
@@ -0,0 +1,107 @@
...
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import eu.stratosphere.nephele.util.StringUtils;
+
+public class LogfileInfoServlet extends HttpServlet {
+	
+	/**
+	 * 
+	 */
```
rmetzger: empty comment. We use 1L for the serialVersionUID.

nephele/nephele-server/src/main/java/eu/stratosphere/nephele/jobmanager/web/LogfileInfoServlet.java
```diff
@@ -0,0 +1,107 @@
...
+	private static final Log LOG = LogFactory.getLog(LogfileInfoServlet.class);
+	
+	private File logDir;
+	
+	public LogfileInfoServlet(File logDir) {
+		this.logDir = logDir;
+	}
+	
+	/*
+	 * (non-Javadoc)
```
rmetzger: remove

nephele/nephele-server/src/main/java/eu/stratosphere/nephele/jobmanager/web/LogfileInfoServlet.java
```diff
@@ -0,0 +1,107 @@
...
+	}
+	
+	private static String readFile(File file) {
+		String out = "";
+		byte[] fileData;
+		
+		try {
+			FileInputStream input = new FileInputStream(file);
+	
+			fileData = new byte[input.available()];
```
rmetzger: You allocate the filesize in memory here.

markus-h:
I removed the wrong comments and changed the reading of the logfiles to a buffered reader

StephanEwen:
Manually merged in f3db2bc02121d668b64faaf759d4d62e4d984ebb

PR #186, opened 2013-10-21
Rename groupReduce to groupMap in scala frontend

aljoscha:
The title really says it all.

### DIFF
--- a/pact-scala/pact-scala-core/src/main/scala/eu/stratosphere/scala/operators/ReduceOperator.scala
+++ b/pact-scala/pact-scala-core/src/main/scala/eu/stratosphere/scala/operators/ReduceOperator.scala
@@ -36,17 +36,18 @@ import eu.stratosphere.scala.analysis.FieldSelector
 import eu.stratosphere.scala.OneInputKeyedScalaContract
 import eu.stratosphere.scala.DataStream
 import eu.stratosphere.scala.OneInputHintable
+import eu.stratosphere.scala.OneInputScalaContract
 
 class GroupByDataStream[In](val keySelection: List[Int], val input: DataStream[In]) {
-  def groupReduce[Out](fun: Iterator[In] => Out): DataStream[Out] with OneInputHintable[In, Out] = macro ReduceMacros.clunkyReduce[In, Out]
-  def combinableGroupReduce(fun: Iterator[In] => In): DataStream[In] with OneInputHintable[In, In] = macro ReduceMacros.combinableReduce[In]
+  def reduceGroup[Out](fun: Iterator[In] => Out): DataStream[Out] with OneInputHintable[In, Out] = macro ReduceMacros.reduceGroup[In, Out]
+  // def combinableReduceGroup(fun: Iterator[In] => In): DataStream[In] with OneInputHintable[In, In] = macro ReduceMacros.combinableReduce[In]
   
-  def reduce(fun: (In, In) => In): DataStream[In] with OneInputHintable[In, In] = macro ReduceMacros.properReduce[In]
+  def reduce(fun: (In, In) => In): DataStream[In] with OneInputHintable[In, In] = macro ReduceMacros.reduce[In]
 }
 
 object ReduceMacros {
   
-  def groupByImpl[In: c.WeakTypeTag, Key: c.WeakTypeTag](c: Context { type PrefixType = DataStream[In] })(keyFun: c.Expr[In => Key]): c.Expr[GroupByDataStream[In]] = {
+  def groupBy[In: c.WeakTypeTag, Key: c.WeakTypeTag](c: Context { type PrefixType = DataStream[In] })(keyFun: c.Expr[In => Key]): c.Expr[GroupByDataStream[In]] = {
     import c.universe._
 
     val slave = MacroContextHolder.newMacroHelper(c)
@@ -60,7 +61,7 @@ object ReduceMacros {
     return helper
   }
   
-  def properReduce[In: c.WeakTypeTag](c: Context { type PrefixType = GroupByDataStream[In] })(fun: c.Expr[(In, In) => In]): c.Expr[DataStream[In] with OneInputHintable[In, In]] = {
+  def reduce[In: c.WeakTypeTag](c: Context { type PrefixType = GroupByDataStream[In] })(fun: c.Expr[(In, In) => In]): c.Expr[DataStream[In] with OneInputHintable[In, In]] = {
     import c.universe._
 
     val slave = MacroContextHolder.newMacroHelper(c)
@@ -161,7 +162,7 @@ object ReduceMacros {
     return result
   }
 
-  def clunkyReduce[In: c.WeakTypeTag, Out: c.WeakTypeTag](c: Context { type PrefixType = GroupByDataStream[In] })(fun: c.Expr[Iterator[In] => Out]): c.Expr[DataStream[Out] with OneInputHintable[In, Out]] = {
+  def reduceGroup[In: c.WeakTypeTag, Out: c.WeakTypeTag](c: Context { type PrefixType = GroupByDataStream[In] })(fun: c.Expr[Iterator[In] => Out]): c.Expr[DataStream[Out] with OneInputHintable[In, Out]] = {
     import c.universe._
 
     val slave = MacroContextHolder.newMacroHelper(c)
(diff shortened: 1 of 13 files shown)


aljoscha:
@aalexandrov pointed this out today and I also think that it is technically not a reduce (it is a map from a list of records to a record (or something else)). I also think, however, that this might confuse people coming from hadoop.

So what do you think?

rmetzger:
Can't we have both names?

aljoscha:
Yes, we could. But why?

rmetzger:
To make everyone happy: The people coming from Hadoop and those with a background in functional programming

aljoscha:
I think there should me one name for things. If there are two people will start wondering what the difference is and might therefore not know what to use.

fhueske:
I agree with Aljoscha in having only one name for each operation. However, we should carefully think about the naming of these and all other functions.

GroupMap basically describes how individual groups are processed (mapped one by one) but does not describe what is happening within a group. One could also interpret this as a mapper over group elements (which would be basically a regular mapper). On the other hand, the function reduce does not indicate that it is applied on individual groups (which is an assumption from the Hadoop context but not true from a functional perspective).

So somebody should have some thoughts about consistent naming of operators.

StephanEwen:
I also agree we should have one name. However, our main audience are most likely not going to be functional programmers and we should be careful not to confuse them too badly. So groupMap() is not the best name, though it is technically correct in my opinion.

In term of the Hadoop community, they would expect names like
1. combinableReduce()
2. reduce();
3. reduceAll();

We could go for someting like
1. reduce() -> more in the function sense als combining always two elements, aka, the only really relevant case of a combinable reduce
2. reduceAsGroup() -> Iterator over entire group
3. reduceAll() -> like the first, but this function exists directly on the data stream, i.e., it is not grouped. We could also just name it reduce(), because the distinction is anyways that it is not called on the grouped stream.

Any thoughts?

aljoscha:
Technically, there are 6 versions:
Directly on DataStream:
1. reduce() the proper functional reduce
2. mapGroup(): groupwise reduce (functional map) over all records
3. combinableMapGroup(): same as 2. but is combinable

On a GroupByDataStream:
4. reduce() the proper functional reduce
5. mapGroup(): groupwise reduce (functional map) over all records in groups grouped by a key
6 combinableMapGroup(): same as 5. but is combinable

Names are still up for debate, right now they are:
1. 2. 3. not offered in scala frontend, is it sensible to offer these?
1. reduce()
2. groupReduce()
3. combinableGroupReduce()

fhueske:
mapGroup is not a functional Map over group elements, since the records are not individually processed and the function holds a state.
How about going with fold instead of reduce for the pure functional reduce?

Btw. is there a way to specify a combine function that is different from its associated reduce function?

aljoscha:
No, it is a functional map over groups (group => ?), that's what the name says. As to the second question: Not there is not right now.

fhueske:
That's exactly what I mean. ;-) It is hard to distinguish between the internal and external function, which might also be because of MapReduce's terminology. A map function is technically not mapping but applied by a map function...

Anyways, maybe calling the function mapGroup_s_() would help to make clear that we are mapping over all groupS and not over the elements of an individual group.

StephanEwen:
As to the six variants @aljoscha mentioned:

On Data Stream directly:
1.  reduce() the proper functional reduce  <-- **we need this one, also in Scala**
2.  mapGroup(): groupwise reduce (functional map) over all records <-- **Infeasible, does not parallelize**
3.  combinableMapGroup(): same as 2. but is combinable <-- **is a little more general than (1), but redundant in my opinion**

On a GroupByDataStream:
1.  reduce() the proper functional reduce <-- **most important one, should also be called like this**
2.  mapGroup(): <-- **The name groupReduce() seems more sensible to me, even though it is not correct in functional terms. Or reduceGroup(). Either should work.**
3.  combinableMapGroup(): same as 5. but is combinable <-- **again, somewhat redundant with (4), i would skip this**

So, we would end up with three forms of reduce, out of which two are called reduce and are really a functional reduce.

BTW: I think in Cascading, the case where we struggle with a name is "groupBy" and "every" to express going over all groups. (As opposed to "Each" which is going over all elements (map)).

aljoscha:
So I will just keep the names and add the global reduce() (which is also inherently combinable). Right?

fhueske:
I had a short discussion with @StephanEwen and we propose the following:
- reduce() for the functional reduce on DataStream and GroupByDataStream. This is an implicitly combinable reduce function on two records at a time.
- reduceGroup() on GroupByDataStream. This is a non-combinable reduce using an iterator over the records of a group.
- reduceAll() on DataStream. This is a non-combinable reduce using an iterator over all records of the stream.

The corner case of a combinable reduce where the combine function is different from the reduce function is currently missing here, but might be added later.

aljoscha:
Alright, I changed the names according to @fhueske's last comment and added reduce/reduceAll on DataStream.

pact-scala/pact-scala-examples/src/main/scala/eu/stratosphere/scala/examples/datamining/BatchGradientDescent.scala
```diff
@@ -61,7 +61,7 @@ abstract class BatchGradientDescent(eps: Double, eta: Double, lambda: Double, ex
     def gradientDescent = (s: DataStream[(Int, Array[Double])], ws: DataStream[(Int, Array[Double], Double)]) => {
 
       val lossesAndGradients = ws cross examples map { (w, ex) => new ValueAndGradient(w._1, computeGradient(ex._2, w._2)) }
-      val lossAndGradientSums = lossesAndGradients groupBy { _.id } combinableGroupReduce { _.reduce(_ + _) }
+      val lossAndGradientSums = lossesAndGradients groupBy { _.id } reduceGroup { _.reduce(_ + _) }
```
StephanEwen: The example line 

```
lossesAndGradients groupBy { _.id } reduceGroup { _.reduce(_ + _) }
```

creates a reducer that is not distributive (combinable) within the system. Can we not use directly

```
lossesAndGradients groupBy { _.id } reduce { _ + _ }
```

The second version is distributive to the system and looks nice as well.

aljoscha: Yes, I simply did a search and replace on the thing.

aljoscha:
Changed according to @StephanEwen's remark about the reduce.

pact-scala/pact-scala-examples/src/main/scala/eu/stratosphere/scala/examples/datamining/KMeans.scala
```diff
@@ -94,8 +94,8 @@ class KMeans extends PlanAssembler with PlanAssemblerDescription with Serializab
     def computeNewCenters(centers: DataStream[(Int, Point)]) = {
 
       val distances = dataPoints cross centers map computeDistance
-      val nearestCenters = distances groupBy { case (pid, _) => pid } combinableGroupReduce { ds => ds.minBy(_._2.distance) } map asPointSum.tupled
-      val newCenters = nearestCenters groupBy { case (cid, _) => cid } combinableGroupReduce sumPointSums map { case (cid, pSum) => cid -> pSum.toPoint() }
+      val nearestCenters = distances groupBy { case (pid, _) => pid } reduceGroup { ds => ds.minBy(_._2.distance) } map asPointSum.tupled
```
StephanEwen, line 97: This one is trickier. It would work efficiently like

```
distances groupBy { case (pid, _) => pid } reduce( (x, y) => { if (x._2.distance < y._2.distance) x else y } )
```

but it is not quite as nice. The minBy function on iterators is very convenient here. However, the reduce that has an iterator is not combinable. Even if we re-introduced the combinable version with the iterator, the system could not figure out that this is a function that can be applied on pairs and hence never use running aggregates (hash based) but only sorted aggregates.

Also not totally straightforward is:

```
distances groupBy { case (pid, _) => pid } reduce ( Seq(_, _).minBy(_._2.distance)  ) 
```

but it creates the undesired sequence objects.

We should solve that by introducing built-in aggregate functions that take key selectors. Something like 

```
distances groupBy { case (pid, _) => pid } aggregate( MIN, { _._2.distance } ) 
```

We can also translate this into very efficient code, as we can use the analyzed selector function to directly go into the runtime object layout, skipping the user code altogether.

pact-scala/pact-scala-examples/src/main/scala/eu/stratosphere/scala/examples/grabbag/Grabbag.scala
```diff
@@ -224,7 +228,7 @@ object ConnectedComponents {
     def propagateComponent = (s: DataStream[(Int, Int)], ws: DataStream[(Int, Int)]) => {
 
       val allNeighbors = ws join undirectedEdges where { case (v, _) => v } isEqualTo { case (from, _) => from } map { (w, e) => e._2 -> w._2 }
-      val minNeighbors = allNeighbors groupBy { case (to, _) => to } combinableGroupReduce { cs => cs minBy { _._2 } }
+      val minNeighbors = allNeighbors groupBy { case (to, _) => to } reduceGroup { cs => cs minBy { _._2 } }
```
StephanEwen, line 231: See above concerning non-distributive reduce.

pact-scala/pact-scala-examples/src/main/scala/eu/stratosphere/scala/examples/datamining/KMeans.scala
```diff
@@ -94,8 +94,8 @@ class KMeans extends PlanAssembler with PlanAssemblerDescription with Serializab
...
+      val newCenters = nearestCenters groupBy { case (cid, _) => cid } reduceGroup sumPointSums map { case (cid, pSum) => cid -> pSum.toPoint() }
```
StephanEwen, line 98: Can we do this as

```
nearestCenters groupBy { case (cid, _) => cid } reduce { (z, v) => (z._1, z._2 + v._2) } map { case (cid, pSum) => cid -> pSum.toPoint() }
```

This is also a candidate for a built-in aggregate function.

```
nearestCenters groupBy { case (cid, _) => cid } aggregate ( SUM, {_._2 } ) map { case (cid, pSum) => cid -> pSum.toPoint() }
```

pact-scala/pact-scala-examples/src/main/scala/eu/stratosphere/scala/examples/graph/ConnectedComponents.scala
```diff
@@ -47,7 +47,7 @@ class ConnectedComponents extends Serializable {
     def propagateComponent = (s: DataStream[(Int, Int)], ws: DataStream[(Int, Int)]) => {
 
       val allNeighbors = ws join undirectedEdges where { case (v, _) => v } isEqualTo { case (from, _) => from } map { (w, e) => e._2 -> w._2 }
-      val minNeighbors = allNeighbors groupBy { case (to, _) => to } combinableGroupReduce { cs => cs minBy { _._2 } }
+      val minNeighbors = allNeighbors groupBy { case (to, _) => to } reduceGroup { cs => cs minBy { _._2 } }
```
StephanEwen, line 50: See above concerning non-distributive reduce.

pact-scala/pact-scala-examples/src/main/scala/eu/stratosphere/scala/examples/graph/TransitiveClosureNaive.scala
```diff
@@ -46,7 +46,7 @@ class TransitiveClosureNaive extends Serializable {
 
       val allNewPaths = paths join edges where { p => p.to } isEqualTo { p => p.from } map joinPaths
       // TODO this fill fail due to bug in iterations runtime
-      val shortestPaths = allNewPaths union paths groupBy { p => (p.from, p.to) } combinableGroupReduce { _ minBy { _.dist } }
+      val shortestPaths = allNewPaths union paths groupBy { p => (p.from, p.to) } reduceGroup { _ minBy { _.dist } }
```
StephanEwen, line 49: See above concerning non-distributive reduce.

pact-scala/pact-scala-examples/src/main/scala/eu/stratosphere/scala/examples/relational/TPCHQuery3.scala
```diff
@@ -63,7 +63,7 @@ class TPCHQuery3 extends PlanAssembler with PlanAssemblerDescription with Serial
 
     val filteredOrders = orders filter { o => o.status == status && o.year > minYear && o.orderPriority.startsWith(priority) }
     val prioritizedItems = filteredOrders join lineItems where { _.orderId } isEqualTo { _.orderId } map { (o, li) => PrioritizedOrder(o.orderId, o.shipPriority, li.extendedPrice) }
-    val prioritizedOrders = prioritizedItems groupBy { pi => (pi.orderId, pi.shipPriority) } combinableGroupReduce { _ reduce addRevenues }
+    val prioritizedOrders = prioritizedItems groupBy { pi => (pi.orderId, pi.shipPriority) } reduceGroup { _ reduce addRevenues }
```
StephanEwen, line 66: See above concerning non-distributive reduce.

pact-scala/pact-scala-core/src/main/scala/eu/stratosphere/scala/operators/ReduceOperator.scala
```diff
@@ -326,4 +327,155 @@ object ReduceMacros {
...
+    val (udtIn, createUdtIn) = slave.mkUdtClass[In]
+    
+    val contract = reify {
+
+      val generatedStub = new ReduceStub with Serializable {
+        val inputUDT = c.Expr[UDT[In]](createUdtIn).splice
+        val outputUDT = inputUDT
+        val udf: UDF1[In, In] = new UDF1(inputUDT, outputUDT)
+        
+        private val combineRecord = new PactRecord()
```
StephanEwen, line 347: A general comment on the generated wrapping UDFs: If you create the PactRecord in open, you save the cost of serializing it into the object that is shipped to and through the cluster.

pact-scala/pact-scala-core/src/main/scala/eu/stratosphere/scala/operators/ReduceOperator.scala
```diff
@@ -326,4 +327,155 @@ object ReduceMacros {
...
+  def globalReduce[In: c.WeakTypeTag](c: Context { type PrefixType = DataStream[In] })(fun: c.Expr[(In, In) => In]): c.Expr[DataStream[In] with OneInputHintable[In, In]] = {
+    import c.universe._
+
+    val slave = MacroContextHolder.newMacroHelper(c)
+    
+//    val (paramName, udfBody) = slave.extractOneInputUdf(fun.tree)
+
+    val (udtIn, createUdtIn) = slave.mkUdtClass[In]
+    
+    val contract = reify {
```
StephanEwen, line 340: This reification with splicing in other code is just really beautiful!

aljoscha:
The comments are all valid, but can we just merge this for now to have the names in place. We can change the stuff that optimizes for performance later, can't we?

aljoscha:
Or should I add the combinable reduceGroup again?

fhueske:
If.a combinable Reduce is added, it might be a good idea to specify the reduce and combine functions separately.
I guess most cases where the combine function is identical to the reduce function can be expressed with the functional reduce.
Cases where this is not true, can simply provide the reduce function twice. That's not super nice, but that way we keep the interface concise and cover all cases.

What do you think?

aljoscha:
yes, that would be good, maybe
input groupBy {} combineGroup {} reduceGroup {} ?

fhueske:
@aljoscha Looks good to me.

I also have some comments on the build-in aggregation functions proposed by @StephanEwen. What would be the result of such a function? One tuple per group with the computed aggregate as single value?

That would be probably not enough. Computing more than one aggregate on a group and preserving the grouping key are very common. So, if we provide something like this, we need a more expressive way to construct the output tuple, similar to the SELECT clause in SQL. To ensure deterministic results, we need to reason about the grouping key. Otherwise, non-key fields might be included in the resulting tuple without being aggregated.

Any thoughts?

StephanEwen:
For the aggregator functions, the result would be one tuple per group. If Multiple aggrgegates are build in a tuple we can list them all

```
something groupBy { x => (x._1, x._3) } aggregate( ( min, { _._2 } ), ( avg { _._4 }), (count, {_._5}) ) 
```

We can either report an error if fields are neither grouped nor aggregated (ansi SQL style) or simply use the field from some tuple (MYSQL style)

StephanEwen:
I will merge this for now.

@aljoscha Please fix the performance issues then in a separate pull request.

StephanEwen:
Merged in b117cbf9c3aff87f37b7138286415704f83595ab

PR #193, opened 2013-10-22
PactRecord allocating too few memory when writing UTF encoded strings

mleich:
PactRecord serializes UTF strings by writing the size of the encoded string (2 bytes) and the actual data.
The current code correctly checks whether size+data actually fit in memory, but it only allocates the memory for the data, not the 2 size bytes.


StephanEwen:
Merged in 08a9e76b9fb3272fb228746a8106040c84e9047c

PR #195, opened 2013-10-22
Remote executor feature

aljoscha:
This pull request will in the end have a RemoteExecutor which will be used like:
val exec = new RemoteExecutor("localhost:666", "some.jar")
exec.execute(myPlan)

(Also usable from Java and written in Java)

For now I just have some refactoring of how Client.run() works and of PactProgram, what do you think of these?

Client.run() now accepts a PlanWithJars which is a wrapper of a Plan and a list of jars required for the job. I removed the jar extraction/deletion logic from Client.

### DIFF
--- a/pact/pact-clients/src/main/java/eu/stratosphere/pact/client/CliFrontend.java
+++ b/pact/pact-clients/src/main/java/eu/stratosphere/pact/client/CliFrontend.java
@@ -51,6 +51,7 @@ import eu.stratosphere.pact.client.nephele.api.Client;
 import eu.stratosphere.pact.client.nephele.api.ErrorInPlanAssemblerException;
 import eu.stratosphere.pact.client.nephele.api.PactProgram;
 import eu.stratosphere.pact.client.nephele.api.ProgramInvocationException;
+import eu.stratosphere.pact.compiler.CompilerException;
 
 /**
  * Implementation of a simple command line fronted for executing PACT programs.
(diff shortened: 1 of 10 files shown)


StephanEwen:
This pull request contains the unrelated commit to rename groupReduce. Can you resubmit this?

uce:
I really really like this. :+1: 

We could also think about merging this with the LocalExecutor. Something like Spark where we run locally if the JM address is "local" or so and otherwise remotly.

In any way it would be great to finish this before the next milestone. :dancers:

aljoscha:
@uce Yes, this is where I am going with this. :smile: I will have ExecutionContext(address, jars*) and this will internally either create a LocalExecutor or a RemoteExecutor.

@StephanEwen I will not have to remove the commit. I'm working on top of the rename-reduce-... branch since I hope that this will be merged soon. When I then rebase this work on top of master it will disappear here. Concerning the lists: I know, but I didn't care because the code is nowhere near performance critical. I will change the code, though.

aljoscha:
Also, don't merge this until it's done, please.

aljoscha:
Ok, I think this is ready to merge once travis gives the okay.

What do you think?

I have some plans regarding what @uce said but that will have to wait. At least this gives the basic infrastructure.

StephanEwen:
Did anyone actually check whether the web frontend and the command line client still work? There is no automated testing for that. Would be good to have that verified ;-)

aljoscha:
I ran the web-frontend today. It still worked :-)

Issue #192, opened 2013-10-22
SingleInputContract deadlock


lukacsg:
The problem usually occurs when I'm running a PACT plan on large input file and I use a single input contract (extending SingleInputContract<T>, eg. Map, Reduce) with two inputs (according to the documentation, public Builder input(Contract ...inputs) lets us to use the union of the multiple input parameters).

In many cases, but not always, stubs are running, but nothing happens - they reach a deadlock waiting for each other. No errors are reported, CPU is not used, no I/O activity. There is no output of the Reduce stub. We can observe output of the Map - Reduce must wait for all of the input data, but Map is able to produce output without collecting all of the input.

I set the DOP to 1, Stratosphere was running in local mode.
The word counter example code reproduces the problematic behavior on a multiplied "hamlet" input file:
for i in {1..100}; do  cat ./hamlet.txt>>hamlet100.txt ; done

 1) The original example code works well on the original hamlet.txt file both with Cogroup and Reduce.
 2) The code runs using Cogroup, no apparent problems.
 3) We usually reach the deadlock using Reduce with 2 inputs.
 4) We also tested Map with two inputs with a slightly different test code, and reached the deadlock.

We experienced the same behaviour with both Stratosphere 2.0 and 4.0 Ozone standalone installs, and also on many different machines (Ubuntu, Fedora; Sun Java 7, OpenJDK 7).

The source code is the variation of WordCounter example to demonstration the bug:
(In the comment: Cogroup and Map test.)

<pre><code>
package hu.sztaki.ilab.er.stratosphere;

import eu.stratosphere.pact.common.contract.CoGroupContract;
import java.util.Iterator;

import eu.stratosphere.pact.common.contract.FileDataSink;
import eu.stratosphere.pact.common.contract.FileDataSource;
import eu.stratosphere.pact.common.contract.GenericDataSink;
import eu.stratosphere.pact.common.contract.MapContract;
import eu.stratosphere.pact.common.contract.ReduceContract;
import eu.stratosphere.pact.common.io.RecordOutputFormat;
import eu.stratosphere.pact.common.io.TextInputFormat;
import eu.stratosphere.pact.common.plan.Plan;
import eu.stratosphere.pact.common.plan.PlanAssembler;
import eu.stratosphere.pact.common.plan.PlanAssemblerDescription;
import eu.stratosphere.pact.common.stubs.CoGroupStub;
import eu.stratosphere.pact.common.stubs.Collector;
[...]

PR #197, opened 2013-10-23
Add union input test (issue #192)

uce:
This test is an adaption of issue #192.

I checked that it deadlocks without @StephanEwen's fix and succeeds with it.


StephanEwen:
Merged in

Issue #189, opened 2013-10-21
Remove IPv4 Stack Hint


StephanEwen:
Nephele has in many scripts hints to Java to run an IPv4 stack. If I recall correctly, the hint was necessary for the discovery manager broadcast-message-discovery in cloud setups.

Since we do not use that mechanism any more, we should be able to remove these hints from the scripts and test setups.

uce:
OK, this was kind of a fail. First I didn't check out the new branch from master, which I noticed after doing the PR. Then I noticed that @StephanEwen had conflicting changes on our master branch in a pom file.

So that's why it says that I reference this issue three times. :zzz: :punch:

StephanEwen:
Fixed as of 361e469bb9ac742bd3c2ddaf930cdebe160300f1

Issue #18, opened 2013-05-21
Rework Nephele Discovery Service


StephanEwen:
Currently, the nephele discovery service adds extra startup time and port conflicts, but is never used. It was originally designed for certain cloud setups and is obsolete in this version.

StephanEwen:
The Discovery service is in some form required to detect the correct public addresses of the taskmanagers for data transfer. It might be possible to merge that with the jobmanager's ipc port, so we have one less port to worry about (for example in shared settings like YARN).

The JobManager discovery via UDP broadcast can be excluded.

rmetzger:
Since I'm working on the Yarn integration (https://github.com/stratosphere/stratosphere/pull/137) (which breaks the Discovery service anyways) it seems to be reasonable to remove it completely together with adding Yarn support.

StephanEwen:
The taskmanager needs to figure out its public IP. That part, the discovery service did well, in my opinion, although I am not sure how it would work if a computer was in two or more subnets at the same time.

Do we have a good alternative to that?

rmetzger:
We could just use the mechanism to detect the TMs own public IP.
I think we could try on each public IP to reach the JobManager (whose address we know for sure from the configuration file)

I need to look more closely into the code. If you are okay with removing the autodiscovery feature (which seems to be dead anyways), I'll see what I can do.

StephanEwen:
Auto-discovery should go. That one rarely works.

If you manage to detect the right IP Address without the UDP ping, that would be perfect. Saves us one more port that is currently not configurable.

PR #203, opened 2013-10-23
Remove IPv4 Stack Hint (issue #189)

uce:
See @StephanEwen's comments in #189.

These were all references to the IPv4 stack I found.

### DIFF
--- a/nephele/nephele-server/src/main/java/eu/stratosphere/nephele/discovery/DiscoveryService.java
+++ b/nephele/nephele-server/src/main/java/eu/stratosphere/nephele/discovery/DiscoveryService.java
@@ -93,9 +93,13 @@ public class DiscoveryService implements Runnable {
 
 	/**
 	 * Flag indicating whether to use IPv6 or not.
+	 * <p>
+	 * This flag used to be set depending on the "java.net.preferIPv4Stack" system property, which was enforced by the
+	 * startup scripts. The hint was necessary for cloud setups in combination with the now removed cloud manager.
+	 * 
+	 * @see {@link https://github.com/stratosphere/stratosphere/issues/189}
 	 */
-	private static final boolean USE_IPV6 = "true".equals(System.getProperty("java.net.preferIPv4Stack")) ? false
-		: true;
+	private static final boolean USE_IPV6 = false;
 
 	/**
 	 * ID for job manager lookup request packets.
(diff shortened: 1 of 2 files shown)


StephanEwen:
Before we can merge this, we need to test whether cluster setups are affected. The discovery manager is important for task managers to find their job-manager-facing ip address. We should also let the os or jvm determine the network stack (ip v4 or ip v6) and not force anything, because that may  prohibit the system to work in certain settings. If I recall correctly, the discovery manager needed ip v4 for broadcast announcements of the job managers address in cloud settings where no slave file was available, but we don't want to go that route.

uce:
OK, sure. Then we should keep the system property check in `DiscoveryService`, no? Both
- `java.net.preferIPv4Stack`, and
- `java.net.preferIPv6Addresses`

are set by the OS, right?

The [docs](http://docs.oracle.com/javase/7/docs/api/java/net/doc-files/net-properties.html) say:

> java.net.preferIPv4Stack (default: false)
> If IPv6 is available on the operating system the underlying native socket will be, by default, an IPv6 socket which lets applications connect to, and accept connections from, both IPv4 and IPv6 hosts. However, in the case an application would rather use IPv4 only sockets, then this property can be set to true. The implication is that it will not be possible for the application to communicate with IPv6 only hosts.
> 
> java.net.preferIPv6Addresses (default: false)
> When dealing with a host which has both IPv4 and IPv6 addresses, and if IPv6 is available on the operating system, the default behavior is to prefer using IPv4 addresses over IPv6 ones. This is to ensure backward compatibility, for example applications that depend on the representation of an IPv4 address (e.g. 192.168.1.1). This property can be set to true to change that preference and use IPv6 addresses over IPv4 ones where possible.

I think I would go with the `java.net.preferIPv6Addresses` check.

Or is there another way?

StephanEwen:
Do we need such flags at all? If we do not use that one specific feature of the discovery service, is there anything left that is IPv4 or IPv6 aware or dependent? I think we should be able to write the "get my public IP" functionality such that it works with both v4 and v6 transparently.

uce:
OK, now I see what you're getting at. I didn't have a close look :eyes: at the `DiscoveryService`, but what you describe (transparent "get my public IP") makes sense.

StephanEwen:
Since the code paths in the discovery service that require the flag are inactive anyways, I will merge this now.
We will follow up on this with issue #18

StephanEwen:
Merged in 361e469bb9ac742bd3c2ddaf930cdebe160300f1

PR #207, opened 2013-10-24
update default jdk to oracle java 7

physikerwelt:
w.r.t. l 130 I think to trailing / is needed

### DIFF
--- a/stratosphere-dist/src/main/stratosphere-bin/bin/nephele-config.sh
+++ b/stratosphere-dist/src/main/stratosphere-bin/bin/nephele-config.sh
@@ -62,7 +62,7 @@ DEFAULT_JOBM_HEAP_MB=256                            # Java heap size for the Job
 DEFAULT_TASKM_HEAP_MB=512                           # Java heap size for the TaskManager (in MB)
 DEFAULT_ENV_PID_DIR="/tmp"                          # Directory to store *.pid files to
 DEFAULT_ENV_LOG_MAX=5                               # Maximum number of old log files to keep
-DEFAULT_ENV_JAVA_HOME="/usr/lib/jvm/java-6-sun/"    # Java home
+DEFAULT_ENV_JAVA_HOME="/usr/lib/jvm/java-7-oracle"  # Java home
 DEFAULT_ENV_JAVA_OPTS=""                            # Optional JVM args
 DEFAULT_ENV_SSH_OPTS=""                             # Optional SSH parameters running in cluster mode
 


StephanEwen:
Thanks for suggesting that. The java path is normally determined from the system set environment variable JAVA_HOME. The configuration parameter is only relevant if that value is not set, which normally indicates that there is no proper Java installation.

The default value was a help for developers. We might want to set out to something that would match a system workout explicit Java installation. I guess that would actually be rather open-jdk-7.

Any thoughts?

physikerwelt:
After changing that variable I could start nephele-visualisation. I'd prefere open-jdk too if it's verified to work.

StephanEwen:
So the visualization Scripts do not properly read the java home variable?

physikerwelt:
That could be the case. Independent of that, the java-sun default should disappear in my opinion.

uce:
OK, I see the problem with the viz script.

I noticed it when I was changing stuff on the scripts before, but didn't think it was problematic.

The problem is the following.
- In `nephele-config.sh` JAVA_HOME is set to the default value if it is not set in the environment.
- JAVA_HOME is then used to set JAVA_RUN. If JAVA_HOME is not set (which will not be the case), it would fall back to setting it to `java` directly (in default path).
- In `nephele-visualization.sh` the JVM is started by `$JAVA_HOME/bin/java` instead of `$JAVA_RUN` like in the task mangager etc. Therefore, if JAVA_HOME is not set or the default doesn't make sense for the system, nepehle-visualization will not start.

I think @physikerwelt couldn't start nephele-viz, because the default path didn't make sense on his system.

I will change the nephele-viz start script to use JAVA_RUN. In general, I think a clean-up of the scripts would be worthwhile.

physikerwelt:
@uce Yes:

```
/b/bin$ ./nephele-visualization.sh 
./nephele-visualization.sh: line 46: /usr/lib/jvm/java-6-sun//bin/java: No such file or directory
```

physikerwelt:
what?

physikerwelt:
OK I understand... but why didn't you modify my pull request?

uce:
Because your pull requests title suggests something totally different than what I did in my pull request.

We still need to remove the trailing `/` though.

Since this issue and #214 are basically the same question, I assume that this here can also be closed?

rmetzger:
It is impossible to modify other people's pull requests (except you have access to their repositories). See also: https://help.github.com/articles/using-pull-requests

@uce did not add the the change you suggested because it does not solve the problem, but hides the symptoms.

physikerwelt:
What I don't understand is why did #211 did not add the slash. At least I could have rebased this pull request...even though I created it using the webui only.
(Beside that is's just a minor style issue: Try /bin//echo test)
Whatever JAVA version you want to use, please try to use one of the travis config:
jdk:
- oraclejdk7
- openjdk7
- openjdk6
  java6sun is not part of the list. So I'd strongly recommend to change or remove the default parameter.
  In order to avoid further conflicts that this pull request causes, I close it.

physikerwelt:
by the was it's still set to sun... and so fails to run if JAVA_HOME is unset and java-sun is not installed

fhueske:
I'm not sure if we should even set a default JAVA_HOME. This is very distribution specific...

Why not using the config as 1st choice, the system configured JAVA_HOME as 2nd choice, and if this does not give a valid home, simply show an error message?
In any case we should add it to the yaml config file (deactivated behind a comment).

physikerwelt:
Yes. That would be better. Otherwise the sun-error message is the first thinh the user sees if he tries Stratosphere. That could be misinterpreted.

Issue #212, opened 2013-10-25
Map after BulkIteration does not work


aljoscha:
This is the exception I get:
eu.stratosphere.pact.compiler.CompilerException: An error occurred while translating the optimized plan to a nephele JobGraph: null
    at eu.stratosphere.pact.compiler.plantranslate.NepheleJobGraphGenerator.postVisit(NepheleJobGraphGenerator.java:556)
    at eu.stratosphere.pact.compiler.plantranslate.NepheleJobGraphGenerator.postVisit(NepheleJobGraphGenerator.java:1)
    at eu.stratosphere.pact.compiler.plan.candidate.SingleInputPlanNode.accept(SingleInputPlanNode.java:142)
    at eu.stratosphere.pact.compiler.plan.candidate.SingleInputPlanNode.accept(SingleInputPlanNode.java:141)
    at eu.stratosphere.pact.compiler.plan.candidate.OptimizedPlan.accept(OptimizedPlan.java:161)
    at eu.stratosphere.pact.compiler.plantranslate.NepheleJobGraphGenerator.compileJobGraph(NepheleJobGraphGenerator.java:165)
    at eu.stratosphere.pact.test.util.TestBase2.getJobGraph(TestBase2.java:282)

@StephanEwen this could also have to to with the mapper chaining I suppose?

Also, when I use the output of an iteration in a file sink and also as one input of a match I get the same exception. Maybe thats related, maybe not.

StephanEwen:
We cannot chain to iterations nodes. This is a one line fix, I'll get onto it...

StephanEwen:
31dbfe65e46d8419d0da8d670b5958407adbdbf2 fixes the mapper chaining issue.

I'll look into the case where the iteration result has two consumers...

StephanEwen:
@aljoscha Do you have a minimal test case for that?

aljoscha:
I think we can close this, right? And the other thing is also fixed.

StephanEwen:
Okay, close it if your problem is solved. I am adding a test case for the case of an iteration followed by a multiple consumers.

StephanEwen:
Fixed in d1498502482d4e5045880c21a8e5b4fab2caa1b4

PR #213, opened 2013-10-25
Fix NepheleJobGraphGenerator bug with iterations

aljoscha:
When traversing it is not checked whether iterations where already
visited, this led to a bug when you used the output of an iteration
twice.


aljoscha:
@StephanEwen Is this how it is supposed to be now? This is unrelated to #212

StephanEwen:
That one looks good to me!

StephanEwen:
Merged in b08240d13a4208eda702539d0c47a068c67515b2

Issue #225, opened 2013-10-30
Make Stratosphere ready for deployment to Maven Central


rmetzger:
Maven Central has some requirements for packages.
These are listed here: https://docs.sonatype.org/display/Repository/Sonatype+OSS+Maven+Repository+Usage+Guide#SonatypeOSSMavenRepositoryUsageGuide-5.Prerequisites

While working at https://github.com/stratosphere/stratosphere/pull/219 I tried to release the artifact, which failed with the following unmet requirements:
- Signature Validation (e.g. we need to sign our artifacts)
- Missing javadoc (`Missing: no javadoc jar found in folder '/eu/stratosphere/nephele-common/0.4-alpha.0-hadoop2'`)

Both should be easy to fix.

PR #226, opened 2013-10-30
Cleanup of main pom; generate javadocs

rmetzger:
Step 1 for https://github.com/stratosphere/stratosphere/issues/225

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -226,100 +226,7 @@
 				</configuration>
 			</plugin>
 
-			<!-- measure and report source code complexity -->
-<!--			<plugin>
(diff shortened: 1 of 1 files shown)


physikerwelt:
+1

StephanEwen:
I agree that a test coverage check is good. But when we reconfigure this, we would probably move to a newer version anyways and check the configuration.

I would vote to remove the old garbage from the pom and setup test coverage checks properly when we will re-introduce them.

fhueske:
fair enough

StephanEwen:
Merged in a8228ceb96bfba2b4b9d63ace8c77077a0689a7b

PR #227, opened 2013-10-30
Additional CrossWithSmall and CrossWithLarge operations

### LINKED ISSUE #218: Replace Cross by CrossWithSmall and CrossWithLarge
The Cross operator is currently very dangerous. Until we have incremental plan rollout, the optimizer has to rely on fallback heuristics when estimates are not available. Unfortunately, there is no proper fallback heuristic for Cross: One side needs broadcast, which can kill any plan if it is the wrong side.

In the spirit of robustness and stability, I would like to take the Cross temporarily out of the equation by offering only a construct that explicitly declares which side is the smaller.

As soon as the heuristics model is improved, or we have implemented the incremental plan rollout, we could re-introduce the self-deciding Cross.

Comments are welcome!

StephanEwen:
Fixes #218


StephanEwen:
Merged in 4285102eb5cd9024c07a841c66240f8ade310d50

Issue #217, opened 2013-10-28
TestBase2 is not platform independent (WINDOWS)


rmetzger:
A student from IMPRO-3 who is using Windows 7 reported a bug with Testbase2.

The methods "getTemp*Path()" have the prefix "file://" hardcoded.

I assigned @markus-h to this issue, since he is a Windows user.
I think there is a way in java to find the platform's way to separate directories.

use "mvn verify" to trigger the bug.

markus-h:
I started working on the issue. Looks like not only file paths need to be adjusted in the testings but also most exec commands.
Moreover I tried to use the shell scripts to start the server (eg start-local.sh), but they seem not to work under cygwin for me. Would perhaps be a nice enhancement?

rmetzger:
We had once a PR that claimed to make Stratosphere Cygwin compatible.

Where do we have exec commands? Do you mean the scripts in /bin ?

uce:
What is the problem with the scripts?

I remember @aljoscha adding a function to change the filepaths on cygwin installations. The changes we introduced afterwards only affected the call to `uname` to figure out on what kind of system we are running. Strange.

StephanEwen:
I think the only problem we are facing here is that the file URI prefix is hard coded.

It should do to replace 

```
file:// + f.getAbsolutePath()
```

with

```
f.toURI().toString()
```

Se we need to externalize the path prefix which contains the scheme

uce:
Yes, sure. @rmetzger checked it yesterday on a Windows machine and indeed `file:\\` does the trick.

But @markus-h's problem with the scripts sounds independent of this to me. Afaik there are no hardcoded paths in the scripts.

markus-h:
The issue for the hard coded URI prefixes is quite easy to resolve, like StephanEwen mentioned. 

But there are multiple exec calls out of Java in the testcases (Runtime.getRuntime().exec(cmd)). For example in ExternalProcessFixedLengthInputFormatTest, ExternalProcessInputFormatTest and even in the ExternalProcessInputFormat class. These seem to be a little bit harder to resolve in windows.

fhueske:
As the name suggests, an ExternalProcessInputFormat starts an external process and parses its output into records.
The process is of course not OS independent. Not sure if the tests can be adapted in a way that the call to Runtime.exec() call or Process that represents the external process is mocked.

Is there a way in Maven to provide platform dependent tests or disable certain tests for different platforms?

StephanEwen:
[...]

PR #229, opened 2013-10-31
Resolved issue with hardcoded "file://" URIs under windows

markus-h:
I implemented the proposed solution for Issue #217. I exchanged every hardcoded file:// path to a platform independent expression.
I tested it with mvn verify on Windows and on my Ubuntu VM in local mode. More tests on different Linux systems and on a cluster would probably be nice.

### DIFF
--- a/nephele/nephele-common/src/main/java/eu/stratosphere/nephele/fs/FileSystem.java
+++ b/nephele/nephele-common/src/main/java/eu/stratosphere/nephele/fs/FileSystem.java
@@ -164,7 +164,7 @@ public abstract class FileSystem {
 		URI localUri;
 
 		try {
-			localUri = new URI("file:///");
+			localUri = isWindows() ?  new URI("file:/") : new URI("file:///");
 		} catch (URISyntaxException e) {
 			throw new IOException("Cannot create URI for local file system");
 		}
@@ -442,5 +442,11 @@ public abstract class FileSystem {
 		} else {
 			return 1;
 		}
+	}
+	
+	// ------------------------------------------------------------------------
+	
+	public static final boolean isWindows() {
+		return System.getProperty("os.name").startsWith("Windows");
 	}
 }
--- a/nephele/nephele-common/src/main/java/eu/stratosphere/nephele/fs/FileSystem.java
+++ b/nephele/nephele-common/src/main/java/eu/stratosphere/nephele/fs/FileSystem.java
@@ -50,6 +50,8 @@ public abstract class FileSystem {
 	 * Object used to protect calls to specific methods.
 	 */
 	private static final Object SYNCHRONIZATION_OBJECT = new Object();
+	
+	private static final boolean IS_WINDOWS =  System.getProperty("os.name").startsWith("Windows");
 
 	/**
 	 * An auxiliary class to identify a file system by its scheme
@@ -447,6 +449,6 @@ public abstract class FileSystem {
 	// ------------------------------------------------------------------------
 	
 	public static final boolean isWindows() {
-		return System.getProperty("os.name").startsWith("Windows");
+		return IS_WINDOWS;
 	}
 }
--- a/nephele/nephele-examples/src/main/java/eu/stratosphere/nephele/example/broadcast/BroadcastJob.java
+++ b/nephele/nephele-examples/src/main/java/eu/stratosphere/nephele/example/broadcast/BroadcastJob.java
@@ -229,7 +229,7 @@ public class BroadcastJob {
 		producer.connectTo(consumer, ChannelType.NETWORK);
 
 		// Attach jar file
-		jobGraph.addJar(new Path("file://" + JAR_FILE.getAbsolutePath()));
+		jobGraph.addJar(new Path(JAR_FILE.toURI().toString()));
 
 		// Submit job
 		Configuration conf = new Configuration();
--- a/nephele/nephele-examples/src/main/java/eu/stratosphere/nephele/example/broadcast/BroadcastJob.java
+++ b/nephele/nephele-examples/src/main/java/eu/stratosphere/nephele/example/broadcast/BroadcastJob.java
@@ -146,7 +146,8 @@ public class BroadcastJob {
 		OUTPUT_PATH = args[6];
 
 		// Prepare jar file
-		JAR_FILE = new File("/tmp/broadcastJob.jar");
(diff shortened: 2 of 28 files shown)


pact/pact-runtime/src/test/java/eu/stratosphere/pact/runtime/task/DataSinkTaskTest.java
```diff
@@ -404,5 +404,14 @@ public int serializeRecord(PactRecord rec, byte[] target) throws Exception
 			return super.serializeRecord(rec, target);
 		}
 	}
+
```
rmetzger, line 408: Can you put this method into one central place, like TestBase2 ?

pact/pact-tests/src/test/java/eu/stratosphere/pact/test/iterative/nephele/customdanglingpagerank/CustomCompensatableDanglingPageRank.java
```diff
@@ -97,7 +98,7 @@ public static JobGraph getJobGraph(String[] args) throws Exception {
 		String pageWithRankInputPath = ""; //"file://" + PlayConstants.PLAY_DIR + "test-inputs/danglingpagerank/pageWithRank";
 		String adjacencyListInputPath = ""; //"file://" + PlayConstants.PLAY_DIR +
 //			"test-inputs/danglingpagerank/adjacencylists";
-		String outputPath = "file:///tmp/stratosphere/iterations";
+		String outputPath = FileSystem.isWindows() ? "file:/c:/tmp/stratosphere/iterations" : "file:///tmp/stratosphere/iterations";
```
rmetzger: We should not hard code the `tmp` directory. It seems that c:/tmp/ is not a Windows-maintained temporary folder.
http://en.wikipedia.org/wiki/Temporary_folder
This is behavior was already wrong before your change. How much work would it be for you to fix that?

sscdotopen:
It's not a bug that this code breaks on windows, its a feature :D

pact/pact-tests/src/test/java/eu/stratosphere/pact/test/iterative/nephele/customdanglingpagerank/CustomCompensatableDanglingPageRank.java
markus-h: Not much work. We could use the constructTestPath method you marked for this path as well, just specifying a different output directory inside of the java.io.tmpdir folder.

pact/pact-runtime/src/test/java/eu/stratosphere/pact/runtime/task/DataSinkTaskTest.java
markus-h, line 408: TestBase2 seems not to be in the Classpath for this class. At least eclipse does not permit the import. Have I overlooked something?

fhueske, line 408: No, that's correct. 
TestBase2 part of the pact-tests Maven module which is not imported (and should not be imported) by pact-runtime.

nephele/nephele-examples/src/main/java/eu/stratosphere/nephele/example/broadcast/BroadcastJob.java
```diff
@@ -229,7 +229,7 @@ private static void runJob(final int run, final BufferedWriter throughputWriter,
 		producer.connectTo(consumer, ChannelType.NETWORK);
 
 		// Attach jar file
-		jobGraph.addJar(new Path("file://" + JAR_FILE.getAbsolutePath()));
+		jobGraph.addJar(new Path(JAR_FILE.toURI().toString()));
```
StephanEwen, line 233: This file has still further platform dependent paths at other places.

nephele/nephele-common/src/main/java/eu/stratosphere/nephele/fs/FileSystem.java
```diff
@@ -443,4 +443,10 @@ private int getNumberOfBlocks(final long length, final long blocksize) {
 			return 1;
 		}
 	}
+	
+	// ------------------------------------------------------------------------
+	
+	public static final boolean isWindows() {
+		return System.getProperty("os.name").startsWith("Windows");
```
StephanEwen: if you want to make this code more efficient, you can go for something like

```
private static final boolean IS_WINDOWS =  System.getProperty("os.name").startsWith("Windows");

public static final boolean isWindows() {
    return IS_WINDOWS;
}
```

That way the check happens only once at class loading and initialization time. Since we do not assume that we change the OS underneath a running JVM, it should be fine ;-)

nephele/nephele-examples/src/main/java/eu/stratosphere/nephele/example/broadcast/BroadcastJob.java
rmetzger, line 233: @markus-h  Is this issue resolved?

markus-h, line 233: I think so

pact/pact-tests/src/test/java/eu/stratosphere/pact/test/util/TestBase2.java
```diff
@@ -246,9 +248,9 @@ public void compareResultsByLinesInMemory(String expectedResultStr, String resul
 	}
 	
 	public File asFile(String path) {
-		if (path.startsWith("file://")) {
-			return new File(path.substring(7));
-		} else {
+		try {
+			return new File(new URI(path).getPath());
```
rmetzger, line 252: I'm not sure if this test is correct anymore. We want to make sure that the path starts with file://

rmetzger:
Sorry for all the critical comments, I'm currently working on `TestBase` so I see some things.

This line won't produce a correct "file://" path: https://github.com/markus-h/stratosphere/blob/Windows_Testbase2-pr/pact/pact-tests/src/test/java/eu/stratosphere/pact/test/util/filesystem/LocalFSProvider.java#L105

StephanEwen:
Merged in 385c0be988eef4fe93602e0d13e526c22179c37e

Issue #208, opened 2013-10-24
Enhancement: Expand test coverage


physikerwelt:
I propose to configure travis in a way that changes in stratosphere/stratosphere trigger re-running the tests in in other project that depend on stratosphere i.e. stratosphere-sopremo and other essential projects.
By that we could expand the "test coverage" without much effort and would have a chance to discover errors that are hard to detect manually, especially if there are dependencies to the -SNAPSHOT version.

StephanEwen:
Can you change that independently in your build server setup? 

I think the crux of the issue is actually that meteor and other projects depend on a snapshot version. These version are bound to change on ways that cause problems for dependent projects.

My feeling is that a correct software engineering solution would be to push a version, say 0.3.0 or so, that represents the current snapshot and will not change. You could use that as a stable dependency. At some point you could move forward to 0.4, which is then stable, while we work on the snapshot version of 0.4.1

What do you think?

physikerwelt:
Yes. I completely agree with you about the version numbers... and I think I'm not the only one.
However, I think a retrigger all tests would be nice for fixes and pre release checks.
I think we have to expect hotfixes for released versions and should think about a process how to deal with them. (Maybe in terms of version numbers?)

physikerwelt:
I found that with a random work approach http://mojo.codehaus.org/buildnumber-maven-plugin/usage.html
I'm not yet quite sure if it's helpful or distracting but it would at least allow to set unique references.

rmetzger:
Hey,

so I think in general, its a good idea to somehow store the version (e.g. the git revision id) in the snapshots, so that we can see which version a user has.
The best solution in my opinion is to write the id into the README.txt file.
I'm really against changing the name of the build again. (We would have to update all the documentation and examples etc.)
I don't see that the buildnumber-maven-plugin does support this (it can only write into .property file).

I think this plugin does everything we need (we need to try if it also can write into non .property files, but it looks like (since the're also writing into xml): https://github.com/ktoso/maven-git-commit-id-plugin

fhueske:
I agree with @rmetzger. 
Having a VERSION file with an ID that can be linked to the last commit of the build is a very good idea and will help to identify bugs in user settings.
A new version number for each commit sounds a bit heavy weight to me.

physikerwelt:
OK. We are moving a little bit into the "How to do releases?" Question.

In a former thread Robert explained, that the master branch moves from a stable state to the next stable state in terms of pull requests.
[...]

PR #231, opened 2013-11-01
Show version and git revision on JobManager startup (in log)

rmetzger:
Hi, 

as a follow up of this discussion https://github.com/stratosphere/stratosphere/issues/208 I propose to include the version and git revision into the JobManager log file like this:

```
00:07:27,186 INFO  eu.stratosphere.nephele.jobmanager.JobManager                 - Starting Stratosphere JobManager (Version: 0.4-SNAPSHOT, Rev:55bfa06)
00:07:27,195 INFO  eu.stratosphere.nephele.discovery.DiscoveryService            - Discovery service socket is bound to /127.0.0.1:7001
00:07:27,209 INFO  eu.stratosphere.nephele.ipc.Server                            - IPC Server Responder: starting
```

There will be also a file in the root directory of the build called `.version.properties` with the following contents:

```
#Generated by Git-Commit-Id-Plugin
#Sat Nov 02 00:06:43 CET 2013
git.commit.id.abbrev=55bfa06
git.commit.user.email=metzgerr@web.de
git.commit.message.full=Hotfixing version detection\nAdd renaming utility to master branch\n
git.commit.id=55bfa06e8971957b3c2896855069b121c7c650df
git.commit.message.short=Hotfixing version detection Add renaming utility to master branch
git.commit.user.name=Robert Metzger
git.build.user.name=Robert Metzger
git.commit.id.describe=55bfa06
git.build.user.email=metzgerr@web.de
git.branch=version_information
git.commit.time=30.10.2013 @ 11\:38\:27 CET
git.build.time=02.11.2013 @ 00\:06\:43 CET
git.remote.origin.url=https\://github.com/stratosphere/stratosphere.git
```

If you want, we can also add the `.version.properties` file into each jar file we deploy. Currently, its only included into `nephele-server`.

### DIFF
--- a/nephele/nephele-server/src/main/java/eu/stratosphere/nephele/jobmanager/JobManager.java
+++ b/nephele/nephele-server/src/main/java/eu/stratosphere/nephele/jobmanager/JobManager.java
@@ -169,7 +169,6 @@ public class JobManager implements DeploymentManager, ExtendedManagementProtocol
 	private volatile boolean isShutDown = false;
 	
 	public JobManager(ExecutionMode executionMode) {
-		logVersionInformation();
 		
 		final String ipcAddressString = GlobalConfiguration
 			.getString(ConfigConstants.JOB_MANAGER_IPC_ADDRESS_KEY, null);
@@ -345,12 +344,12 @@ public class JobManager implements DeploymentManager, ExtendedManagementProtocol
 	/**
 	 * Log Stratosphere version information.
 	 */
-	private void logVersionInformation() {
-		String version = getClass().getPackage().getImplementationVersion();
+	private static void logVersionInformation() {
+		String version = JobManager.class.getPackage().getImplementationVersion();
 		String revision = null;
 		try {
 	    	Properties properties = new Properties();
-	    	InputStream propFile = getClass().getClassLoader().getResourceAsStream(".version.properties");
+	    	InputStream propFile = JobManager.class.getClassLoader().getResourceAsStream(".version.properties");
 			if(propFile != null) {
 				properties.load(propFile);
 				revision = properties.getProperty("git.commit.id.abbrev");
@@ -370,7 +369,8 @@ public class JobManager implements DeploymentManager, ExtendedManagementProtocol
 	 */
 	@SuppressWarnings("static-access")
 	public static void main(final String[] args) {
-
+		logVersionInformation();
+		
 		final Option configDirOpt = OptionBuilder.withArgName("config directory").hasArg()
 			.withDescription("Specify configuration directory.").create("configDir");
 
--- a/nephele/nephele-server/src/main/java/eu/stratosphere/nephele/jobmanager/JobManager.java
+++ b/nephele/nephele-server/src/main/java/eu/stratosphere/nephele/jobmanager/JobManager.java
@@ -34,6 +34,7 @@
 package eu.stratosphere.nephele.jobmanager;
 
 import java.io.IOException;
+import java.io.InputStream;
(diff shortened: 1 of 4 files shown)


nephele/nephele-server/src/main/java/eu/stratosphere/nephele/jobmanager/JobManager.java
```diff
@@ -167,7 +169,8 @@
 	private volatile boolean isShutDown = false;
 	
 	public JobManager(ExecutionMode executionMode) {
-
+		logVersionInformation();
```
StephanEwen: Would it be good to move this to the main(String[] args) method? That method initializes the context (such as also the global configuration)

StephanEwen:
Quick question: Why do we need two version property files? One in the nephele-server jar and one in the distribution?

rmetzger:
I'm moving it to the main method.

The version file for the distribution is the one we certainly need.
The one in the nephele-server.jar file is probably not necessary if we know the path to the root-directory of the stratosphere installation at runtime. (Which should be the case if stratosphere has been started with our scripts)
The nephele-server.jar method is a bit more reliable because we read the file directly from the same jar as the JobManager itself (using the classloader).
There is currently no case in which users start stratosphere without our scripts. But I'm trying to change our Yarn implementation to be independent of the scripts, and then, we probably need to read it from the jar. (Currently, Yarn uses the nephele-jobmanager.sh script to start the JM. So the script and everything must be available on each machine (e.g. HDFS). I hope that there is way with Yarn to start an application from a jar-file. I have not really looked into the issue, so might be totally different in the end)

rmetzger:
I moved the method into main(). 
I'm also in contact with travis to analyze the build issues we're experiencing right now.

aljoscha:
The f*ck, this makes my eclipse complain about some plugin that it cannot find when I try to import. Average Joe is not gonna like this.

rmetzger:
Okay, I have to admit its my fault. I did not test this with eclipse.
Is there a way to ignore plugins in eclipse out of maven (profiles, etc.)

Seems like there is a solution to the problem: https://github.com/ktoso/maven-git-commit-id-plugin/issues/22

aljoscha:
You can add this in the pom on the same level as <plugins> and it will ignore it in eclipse:

``` xml
<pluginManagement>
    <plugins>
        <!--This plugin's configuration is used to store Eclipse m2e settings 
            only. It has no influence on the Maven build itself. -->
        <plugin>
            <groupId>org.eclipse.m2e</groupId>
            <artifactId>lifecycle-mapping</artifactId>
            <version>1.0.0</version>
            <configuration>
                <lifecycleMappingMetadata>
                    <pluginExecutions>
                        <pluginExecution>
                            <pluginExecutionFilter>
                                <groupId>
                                    pl.project13.maven
                                </groupId>
                                <artifactId>
                                    git-commit-id-plugin
                                </artifactId>
                                <versionRange>
                                    [2.1.5,)
                                </versionRange>
                                <goals>
                                    <goal>revision</goal>
                                </goals>
                            </pluginExecutionFilter>
                            <action>
                                <ignore></ignore>
                            </action>
                        </pluginExecution>
                    </pluginExecutions>
                </lifecycleMappingMetadata>
            </configuration>
        </plugin>
    </plugins>
</pluginManagement>
```

StephanEwen:
Can you prepare a merge request for that?

rmetzger:
Fixed in https://github.com/stratosphere/stratosphere/commit/24f658177eb2f3272dfe0e9791b774bd86d52861

PR #232, opened 2013-11-02
Dump PACT plan as json, using the local executor

rmetzger:
I needed this feature for generating graphs with graphviz

This python script converts a JSON Graph into a dot file: https://github.com/rmetzger/pact-plan-to-dot

Example: (I can't show it in a higher resolution, due to legal restrictions)
plan

### DIFF
--- a/pact/pact-clients/src/main/java/eu/stratosphere/pact/client/LocalExecutor.java
+++ b/pact/pact-clients/src/main/java/eu/stratosphere/pact/client/LocalExecutor.java
@@ -15,6 +15,8 @@
 
 package eu.stratosphere.pact.client;
 
+import java.util.List;
+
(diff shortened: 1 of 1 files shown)


fhueske:
Converting the Json to Dot for plotting the plan DAG is a cool idea.

I think we might even push that further and use Dot for the web interface #185, #143.
I found a JavaScript based GraphViz (Dot) compiler. Canviz: http://code.google.com/p/canviz/
The online demo looks quite promising IMHO: http://www.ryandesign.com/canviz/

Canviz is published under MIT license though. Not sure about the legal aspects here, but from a first glimpse it might be OK.

What do you think?

fhueske:
Arbor.js https://github.com/samizdatco/arbor might be another candidate... Again MIT license.

PR #234, opened 2013-11-04
Exclude jruby-complete from HBase.

rmetzger:
We don't need jruby-complete with hbase. It is used for the interactive HBase shell and causes troubles for Stratosphere users.


StephanEwen:
Merged in c92f7900abc7ca6b38b4c562cacca26e6320ad4a

PR #238, opened 2013-11-04
Fixed CVS Parsers

StephanEwen:
This pull request fixes several issues with the CVS parsers:
- Handling of quoted strings
- Precision when parsing floating point values (#6)


aljoscha:
Could this be the cause for #245 ? When the input does not match the fields specified for a RecordInputFormat not records are generated but there is no warning or exception or anything.

PR #239, opened 2013-11-05
Change how DataSourceFormats are handled internally.

aljoscha:
This and the addition of implicit UDT generation makes it easy to implement custom DataSourceFormats. I noticed the need for this from @rmetzger's avro input format.

### DIFF
--- a/pact-scala/pact-scala-core/pom.xml
+++ b/pact-scala/pact-scala-core/pom.xml
@@ -29,12 +29,12 @@
 	    <dependency>
 		    <groupId>org.scala-lang</groupId>
 		    <artifactId>scala-reflect</artifactId>
-		    <version>2.10.1</version>
+		    <version>2.10.3</version>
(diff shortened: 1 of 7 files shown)


aljoscha:
So should we merge this without the OK from travis? It don't think this could break anything. @rmetzger, are you still planning to run scala with the avro input format for the telko guys?

rmetzger:
Sure, I need this PR to be merged. But I will continue to work on this tomorrow evening, so there is some time left.

aljoscha:
This contains some other commits that were also necessary for this. I ran mvn verify on my machine, so this should be good to merge.

StephanEwen:
I'll take a quick look and then merge it...

aljoscha:
Travis seems to be working again ...

StephanEwen:
Merged in a02f070305e768a7976ad4cb503f091113a4a1bc

PR #250, opened 2013-11-06
Fixes #242. Excludes example artefacts from distribution lib folder

### LINKED ISSUE #242: Exclude the pact-examples and pact-scala-examples from being copied to lib
Curretly, the maven project jars are copied to lib. That means the classes are part of the bootstrap classpath. When one wants to run a modified example, it fails becasuse the classes are already in the class path (in an older version).

fhueske:
Excludes nephele-examples, pact-examples, and pact-scala-examples from distribution lib folder.


StephanEwen:
Merged in cfb20f9362d832fd2f813e9a8f8575aa25f1ca70

PR #253, opened 2013-11-08
Manually configured JAVA_HOME is preferred over system configuration. 

fhueske:
Manually configured JAVA_HOME is preferred over system configuration. 
No default JAVA_HOME is used. Instead an error message is shown if neither config nor system JAVA_HOME is defined. Addresses #207


rmetzger:
With this pull request, Stratosphere does not work on our cluster out of the box.

fhueske:
I don't think that should be an issue since our specific setup should not influence the default settings.
We should either properly set JAVA_HOME or remove one character ('#') in stratosphere-conf.yaml everytime we do a new setup.

I think having a default JVM specified is a bad idea since it is distribution specific. If JAVA_HOME is not set and the default configuration does not work, no process is started and the only error message is hidden in the ./log/jobmanager.out file.
Furthermore, manual configuration should override system configuration and not the other way round.

rmetzger:
If I see it right, we need the `JAVA_HOME` **only** for 
`JAVA_RUN=$JAVA_HOME/bin/java`

Your pull request disables our default strategy to just use "java"  (resolved by the PATH variable) if JAVA_HOME is not set.
In almost all cases invoking "java" will suffice

fhueske:
OK, I see. So we need another check if "java" is available.

The right order is IMHO:
1) stratosphere-conf.yaml
2) JAVA_HOME
3) java
4) error message.

rmetzger:
Yes, I think this is the best order.

fhueske:
OK, I adapted the script to use java as third option before showing an error message.

rmetzger:
Okay, works on the cluster. 

I hereby approve this pull request.

StephanEwen:
Merged in 1edf660e85a19772771f4aa50407f7e5dd860385

PR #254, opened 2013-11-08
Add support for using arbitrary Pact Value classes in scala frontend

aljoscha:
Also add checks in key selector extraction to check whether the selected
fields can actually be part of a key.

@StephanEwen, btw, PactDouble implements Key.


pact-scala/pact-scala-examples/src/main/scala/eu/stratosphere/scala/examples/grabbag/Grabbag2.scala
```diff
@@ -0,0 +1,54 @@
...
+import eu.stratosphere.scala.analysis.UDT
+
+import eu.stratosphere.scala.operators._
+
+
+object Main2 {
+  
+  def main(args: Array[String]) {
+
+    val input = TextFile("file:///home/aljoscha/dummy-input")
```
StephanEwen: Can you "de-aljoscha-fy" this example, or make it a proper test case?

aljoscha:
Ok, I removed the Grabbag2 think, added WordCountPactValue, which is the WordCount example but uses PactString and PactInteger. There is also an integration test for that.

StephanEwen:
Merged in cbfc5c2097f2286f6e1b27beb695d3d2292c9570

PR #258, opened 2013-11-08
Scala renames and cleanups

aljoscha:


aalexandrov:
Can we use DataCollection instead of DataSet? Since the system operators do not eliminate duplicates, DataCollection seems seems to be the more appropriate than DataSet. If we add duplicate elimination and duplicate eliminating operators to the runtime in the future, we can use DataSet as a specialisation of DataCollection that implicitly enforces this behaviour.

aljoscha:
But in my mind DataSet is not associated with a mathematical Set. Other frameworks also use some variation of DataSet (see spark; Resilient Distributed DataSet).

sscdotopen:
I agree with Aljoscha, I think dataset is the common term used by machine learners and data scientists

StephanEwen:
Merged in fd9520a571a5722774782ffdebf79237b6cf87c2

rmetzger:
Does the change break bigdataclass ?

aljoscha:
I changed bigdataclass to work with the new version.

aalexandrov:
The following Scala classes still need to be renamed:
- CoGroupDataStream -> CoGroupDataSet
- JoinDataStream -> JoinDataSet
- CrossDataStream -> CrossDataSet

If I see more I'll add them here.

PR #269, opened 2013-11-11
Added count() function on grouped data

StephanEwen:
Added a function count() to count elements in a group and return a two tuple with original group value and count.

Allows to write code like:

```
val words = input flatMap { _.toLowerCase().split("""\W+""") filter { _ != "" } }
val counts = words groupBy { x => x } count()
```


aljoscha:
Looks good to me but could you please also add an IT Test for the example.

PR #272, opened 2013-11-12
Fix workset iterations in scala frontend

aljoscha:
We used the globalPos to create WorksetIteration, these are not
available before the post pass has run, therefore use the localPos for
now.

Also add maxIterations parameter to iterateWithWorkset


aljoscha:
Do not merge until I give the go ahead.

aljoscha:
Alright, I think this badboy is ready for merging.

PR #273, opened 2013-11-12
throw exception instead of return max dop of 1 on IOException (from file system/hdfs)

rmetzger:


rmetzger:
merged in https://github.com/stratosphere/stratosphere/commit/3fff141844347ebea413d4223f4e42d4d46924a2

PR #274, opened 2013-11-15
Add new way of specifying scala jobs: query {}

aljoscha:
This is just to let you know what I'm working on. Do not merge yet.

I want to get rid of ScalaPlan and let people specify queries like this:

``` scala
def getScalaPlan(textInput: String, wordsOutput: String) = {
  query {
    val input = TextFile(textInput)

    val words = input flatMap { _.toLowerCase().split(" ") }
    val counts = words groupBy { case (word, _) => word } count()

    val output = counts.write(wordsOutput, CsvOutputFormat)

    Seq(output)
  }
}
```

I want to introduce this now before the new release so that people get to know it. In the future this will allow looking at the AST of the whole query, not just at single anonymous functions. Also, we could have something like this instead of the code mentioned above:

``` scala

val ex = ExecutionContext("local") // ExecutionContext("localhost:666", <jars>)
ex.query {
  val input = TextFile(textInput)
  val words = input flatMap { _.toLowerCase().split(" ") }
  val counts = words groupBy { case (word, _) => word } count()
  val output = counts.write(wordsOutput, CsvOutputFormat)
  Seq(output)
}
```

Where the query is immediately executed on the executor.

### DIFF
--- a/pact-scala/pact-scala-core/src/main/scala/eu/stratosphere/scala/Query.scala
+++ b/pact-scala/pact-scala-core/src/main/scala/eu/stratosphere/scala/Query.scala
@@ -1,0 +1,15 @@
+package eu.stratosphere.scala
+
+import eu.stratosphere.pact.generic.contract.Contract
+import eu.stratosphere.scala.analysis.NewGlobalSchemaGenerator
+import eu.stratosphere.pact.common.contract.GenericDataSink
+import eu.stratosphere.pact.common.plan.Plan
+import scala.collection.JavaConversions._
+import java.util.Calendar
+
+object query {
+  def apply(block: => Seq[ScalaSink[_]]): ScalaPlan = {
+    val sinks = block
+    new ScalaPlan(sinks)
+  }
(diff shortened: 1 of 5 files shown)


uce:
Big :+1: on the single `ExecutionContext`.

I think we should support both things you proposed.

---

Why do you have the `query { ... }` block in the `getScalaPlan`? Can't you just look at the AST of the function or do you need the block for this?

aljoscha:
I cannot just look at any ASTs. I need some kind of hook, so in some future version query could become a macro which would then allow me to look at the ast of the passed-in function.

uce:
I would also be OK with your proposals w/o the ability to analyze the complete query (i.e. skipping the query block in proposal 1). What do we expect to gain from analyzing the AST in Scala at this level of "query" granularty?

aalexandrov:
I'm drooling...

StephanEwen:
This is not a real pull request, so I would like to close this for now.

Issue #261, opened 2013-11-09
JDBC Input and Output format for Stratosphere


emrehasan:
Hi,

I would like to contribute to Stratosphere too. On your page https://github.com/stratosphere/stratosphere/wiki/Starter-Jobs i found the task 'JDBC Input and Output format for Stratosphere'  and would like to get more information on that.

rmetzger:
Hi,

Cool, we are very happy that you want to contribute to the project!

I would recommend you to start with implementing the JDBC Input Format. Create a class `JDBCInputFormat` that extends from `eu.stratosphere.pact.common.io.GenericInputFormat`.  The overall idea of the InputFormat is being able to read from a database with a JDBC driver.
So I would recommend you to set up a MySQL database (or mariadb, or anything that's open source and has JDBC support) and populate it with some data.
Then, write a little test stratosphere job that reads the result of a query into a CSV file.
Implement your `JDBCInputFormat` until your job runs successfully. 

You have to override some methods for the input format to work:
- `nextRecord()` is called by the framework to get the nextRecord from the input. So in your case, this method should return the next row from the query result.
- `reachedEnd()` the loop that calls `nextRecord()` checks with `reachedEnd()` if there are records left.
- The GenericInputFormat has a mechanism to create "input splits". This is the way Stratosphere reads files in parallel. I would recommend you to circumvent this mechanism at first and only use one input split for the input format. (so the input format does not run in parallel, but that's okay since I would not recommend loading "big data" scale data)
- I would suggest to create a constructor for the `JDBCInputFormat` with two arguments, one for the JDBC connection and one for a SQL Query.

Please ask here if you have questions or if you need help with anything. (I can offer you to review your code. We can also setup a Skype call (german or english) if you need further assistance)

Please also report any issues, inconveniences or missing information. This is also very valuable information for us, since we want to make the start with our system as easy as possible.

emrehasan:
Should the PactRecord be filled with just Strings like it is done in TextInputFormat or should the original datatypes be considered? If yes which sql-types shall be considered?

rmetzger:
It would be nice to map the SQL types to the respective Pact\* types, but I think the most basic types such as int, long, string, bool etc. should be sufficient for a first approach.

If you have types that are absolutely not mappable, map them to string or throw an exception.

emrehasan:
[...]

Issue #122, opened 2013-10-01
Remove generated comments


sscdotopen:
I suggest to remove all Eclipse-generated comments from the source like

```
/* (non-Javadoc)
         * @see eu.stratosphere...
*/

/**
     * {@inheritDoc}
     */
```

Such comments offer zero value to a reader.

aljoscha:
I agree, we just need the resident regex wizard to come up with the all-powerful regex to remove all that stuff.

rmetzger:
+1

I think that we also wanted to remove all `@author` tags.
We should probably do this with as few pull requests as possible (@uce has a large pull request pending).

aljoscha:
We should first merge in the big stuff because after changing basically every file a merge (or rebase) can get quite nasty.

PR #279, opened 2013-11-21
Implemented JDBCInputFormat for MariaDB, MySQL, Postgres and Derby. Code is covered by unit-test

emrehasan:
The first part (input-format) of issue #261 is implemented now and is also covered by a junit-test.


pact/pact-common/pom.xml
```diff
@@ -23,10 +23,25 @@
...
+		<dependency>
+			<groupId>postgresql</groupId>
+			<artifactId>postgresql</artifactId>
+			<version>9.1-901.jdbc4</version>
+		</dependency>
+		<dependency>
+			<groupId>org.apache.derby</groupId>
+			<artifactId>derby</artifactId>
+			<version>10.10.1.1</version>
+		</dependency>
```
rmetzger: I think we don't want to add all these databases as dependencies.

Does your Input format also run without the dependencies?
So that users that want to use a JDBC driver can add it for their own stratosphere job.

pact/pact-common/src/main/java/eu/stratosphere/pact/common/io/JDBCInputFormat.java
```diff
@@ -0,0 +1,411 @@
...
+import eu.stratosphere.pact.common.type.base.PactDouble;
+import eu.stratosphere.pact.common.type.base.PactFloat;
+import eu.stratosphere.pact.common.type.base.PactInteger;
+import eu.stratosphere.pact.common.type.base.PactLong;
+import eu.stratosphere.pact.common.type.base.PactNull;
+import eu.stratosphere.pact.common.type.base.PactShort;
+import eu.stratosphere.pact.common.type.base.PactString;
+
+/**
+ * InputFormat to read data from a database an generate PactReords.
```
rmetzger: typo

pact/pact-common/pom.xml
emrehasan: Yes it should work without the dependencies but the derby-dependency is necessary to automatically build and test

pact/pact-common/src/main/java/eu/stratosphere/pact/common/io/JDBCInputFormat.java
```diff
@@ -0,0 +1,411 @@
...
+ *
+ * The position of a value inside a PactRecord is determined by the table
+ * returned.
+ *
+ * @see Configuration
+ * @see PactRecord
+ * @see DriverManager
+ */
+public class JDBCInputFormat extends GenericInputFormat {
+
```
rmetzger: Concerning code style and guidelines: Could you remove the newlines after the class header, the enums etc.

rmetzger:
Okay, some more general feedback:
Can you make the inputFormat more generic?
It has the driver classes for some databases hardcoded. It would be nicer to allow the InputFormat to work with ALL JDBC drivers/databases.
I think the user has to supply the JDBC driver name as a string (remember the input format runs in a distributed fashion, so you need to make sure that the JDBC driver is loaded on the node that is running the input format. (So we need to test your InputFormat on a cluster) (We can do that here at the TU).

I'm not sure if it is neccessary to store all these information as class properties:

```
 +    private String dbTypeStr;
 +    private String host;
 +    private Integer port;
 +    private String dbName;
 +    private String username;
 +    private String password;
 +    private String derbyDBPath;
```

They should all be encoded in the connection string. I think you can still keep the constructor that gets a Configuration object, but it is probably better to construct the connection string there.

I'm not sure if you really need the `need_configure` variable. You should be able to determine if configuration is required based on the availability of a valid connection string (the variable name is not conform to Java variable naming!)

We don't want to have these comments

```
 /*
 +     * (non-Javadoc)
 +     * @see eu.stratosphere.pact.generic.io.InputFormat
 +     */
```

because they don't add any value (any smart IDE will point you to the right documentation).

I think it is okay to have only one example file. Write about the different variants in a comment!

Please ask me if you have any questions regarding my comments.
It is not my intention to discourage you ;) Your overall work is quite good!

pact/pact-common/src/test/java/eu/stratosphere/pact/common/io/JDBCInputFormatTest.java
```diff
@@ -0,0 +1,239 @@
...
+public class JDBCInputFormatTest {
+
+    JDBCInputFormat jdbcInputFormat;
+    Configuration config;
+    static Connection conn;
+
+    @BeforeClass
+    public static void setUpClass() {
+        try {
+            System.out.println("Before class");
```
rmetzger: remove debugging messages.

pact/pact-common/src/test/java/eu/stratosphere/pact/common/io/JDBCInputFormatTest.java
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    @After
+    public void tearDown() {
+        jdbcInputFormat = null;
+    }
+
+    public void configureEmbeddedDerbyConfig() {
+        jdbcInputFormat = new JDBCInputFormat("jdbc:derby:memory:ebookshop", "select * from books");
+        jdbcInputFormat.configure(null);
```
rmetzger: Why is it required to call the InputFormat with a null argument?
Does the user have to do this? or is it for testing purposes

pact/pact-common/src/test/java/eu/stratosphere/pact/common/io/JDBCInputFormatTest.java
```diff
@@ -0,0 +1,239 @@
...
+    @Test(expected = IllegalArgumentException.class)
+    public void testInvalidQuery(){
+        JDBCInputFormat conformat = new JDBCInputFormat("jdbc:derby:memory:ebookshop","abc");
+        conformat.configure(null);
+    }
+    
+    @Test
+    public void testJDBCInputFormatSettingPactRecordNormally() throws IOException {
+        PactRecord buffRecord;
+        List<PactRecord> records = new ArrayList<PactRecord>();
```
rmetzger: You don't need to store the PactRecords in a list!

Use only one PactRecord and pass it into the InputFormat over and over again (this is supported by the framework).
You can use a integer variable to count the number of records. Lift the checks on the fields to the while loop, then you have exactly the same validation without storing many PactRecords in a list.

pact/pact-common/pom.xml
rmetzger: You can add dependencies with a `test` scope. Then, they are only loaded for the execution of the tests.

pact/pact-common/src/main/java/eu/stratosphere/pact/common/io/JDBCInputFormat.java
```diff
@@ -0,0 +1,411 @@
...
+     */
+    @Override
+    public boolean nextRecord(PactRecord record) {
+        try {
+            resultSet.next();
+            ResultSetMetaData rsmd = resultSet.getMetaData();
+            int column_count = rsmd.getColumnCount();
+            record.setNumFields(column_count);
+
+            for (int pos = 0; pos < column_count; pos++) {
```
rmetzger: If the loop starts from `pos = 1` you don't need the `pos+1` everywhere. You always add 1 to `pos`.

pact/pact-common/src/test/java/eu/stratosphere/pact/common/io/JDBCInputFormatTest.java
zentol: if the constructor is called with a string right away it doesn't matter what is passed to configure afterwards, as it is ignored. (We assume that username/password is encoded in the string)
passing null was supposed to show that.

pact/pact-common/src/main/java/eu/stratosphere/pact/common/io/JDBCInputFormat.java
zentol: we always add 1 to pos because resutlSet and PactRecord use different indices.
the first field of a PactRecord has index 0, the first field in a resultSet index 1.
see retrieteTypeAndFillRecord:
 record.setField(pos, new PactDouble(resultSet.getBigDecimal(pos + 1).doubleValue()));

zentol:
"Can you make the inputFormat more generic?"

if we let the user supply the driver url it should work. we could make it even more generic if we always require the user to pass the database url aswell, since that is hardcoded to the db type right now aswell.

pact/pact-common/src/test/java/eu/stratosphere/pact/common/io/JDBCInputFormatTest.java
rmetzger: Okay, got it. So its a test to ensure that the configuration from the constructor is not overwritten. Then keep it.

pact/pact-common/src/main/java/eu/stratosphere/pact/common/io/JDBCInputFormat.java
rmetzger: okay, you're right.
Seems that JDBC is a bit stupid here. Nobody starts counting from 1 ;)

rmetzger:
Okay, I would suggest to do it this way. Hardcoding database names is not a good idea, since JDBC is designed to be database independent.

You can, if you want, add the code you've already written to a static method that allows to build the database url.

Like this

```
GenericDataSource source = new GenericDataSource(new JDBCInputFormat(JDBCInputFormat.getURL("mysql", "username", "secret", "table"), "SELECT blabl"), "Data Source"); 
```

emrehasan:
I think we removed comments with (non-Javadoc) prefix, but i assume my partner and me got this from your classes like you can see in GenericDataSource[85]

rmetzger:
Thanks for removing the comments.
You are right, our code is full of those comments, but we want to get rid of them: https://github.com/stratosphere/stratosphere/issues/122

pact/pact-common/src/main/java/eu/stratosphere/pact/common/io/JDBCInputFormat.java
```diff
@@ -0,0 +1,351 @@
...
+     *
+     * @param pos PactRecord position to be set.
+     * @param type SQL type of the resultSet value.
+     * @param record Target PactRecord.
+     *
+     */
+    private void retrieveTypeAndFillRecord(int pos, int type, PactRecord record) throws SQLException, NotTransformableSQLFieldException {
+        switch (type) {
+            case java.sql.Types.NULL:
+                record.setField(pos, new PactNull());
```
StephanEwen: Save an Object by going for PactNull.INSTANCE

pact/pact-common/src/main/java/eu/stratosphere/pact/common/io/JDBCInputFormat.java
```diff
@@ -0,0 +1,351 @@
...
+    }
+
+    /**
+     * Creates a JDBCInputFormat and configures it.
+     *
+     * @param parameters Configuration with all connection parameters.
+     * @param query Query to execute.
+     */
+    public JDBCInputFormat(Configuration parameters, String query) {
+        this.driverName = parameters.getString("driver", "");
```
StephanEwen: Can you make these Strings constants?

pact/pact-common/src/main/java/eu/stratosphere/pact/common/io/JDBCInputFormat.java
```diff
@@ -0,0 +1,351 @@
...
+            //			case java.sql.Types.DISTINCT:
+            //			case java.sql.Types.OTHER:
+            //			case java.sql.Types.REF:
+            //			case java.sql.Types.ROWID:
+            //			case java.sql.Types.STRUCT:
+        }
+    }
+
+    private boolean isFieldNullOrEmpty(String field) {
+        return (field == null || field.equals(""));
```
StephanEwen: fiel.length() == 0 is faster

pact/pact-common/src/main/java/eu/stratosphere/pact/common/io/JDBCInputFormat.java
```diff
@@ -0,0 +1,351 @@
...
+     * parameters (if necessary), establishing the connection and executing the
+     * query.
+     *
+     * @param parameters Configuration containing all or no parameters.
+     */
+    @Override
+    public void configure(Configuration parameters) {
+        boolean needConfigure = isFieldNullOrEmpty(this.query) || isFieldNullOrEmpty(this.dbURL);
+        if (needConfigure) {
+            this.driverName = parameters.getString("drivername", null);
```
StephanEwen: Use String constants please, as suggested above.

pact/pact-common/src/main/java/eu/stratosphere/pact/common/io/JDBCInputFormat.java
```diff
@@ -0,0 +1,351 @@
...
+            this.username = parameters.getString("username", null);
+            this.password = parameters.getString("password", null);
+            this.query = parameters.getString("query", null);
+            this.dbURL = parameters.getString("dburl", null);
+        }
+
+        try {
+            prepareQueryExecution();
+        } catch (SQLException e) {
+            throw new IllegalArgumentException("Configure failed:\t!" + e.getMessage());
```
StephanEwen: Please always chain exception, ie, pass the cause such that it does not get lost.

pact/pact-common/src/test/java/eu/stratosphere/pact/common/io/JDBCInputFormatTest.java
```diff
@@ -0,0 +1,194 @@
...
+     */
+    private static void createConnection(String dbURL) {
+        try {
+            Class.forName("org.apache.derby.jdbc.EmbeddedDriver");
+            conn = DriverManager.getConnection(dbURL);
+            createTable();
+            insertDataToSQLTables();
+            conn.close();
+        } catch (ClassNotFoundException except) {
+            except.printStackTrace();
```
StephanEwen: You need to call Assert.fail() to mark the test case as failed at this position. Otherwise the test will go on and either sucecdd, or produce an error later, which one will have to track back to the setup problem.

pact/pact-common/src/test/java/eu/stratosphere/pact/common/io/JDBCInputFormatTest.java
```diff
@@ -0,0 +1,194 @@
...
+        } catch (SQLException except) {
+            except.printStackTrace();
```
StephanEwen: Same as above.

pact/pact-common/src/test/java/eu/stratosphere/pact/common/io/JDBCInputFormatTest.java
```diff
@@ -0,0 +1,194 @@
...
+    JDBCInputFormat jdbcInputFormat;
+    Configuration config;
+    static Connection conn;
+
+    @BeforeClass
+    public static void setUpClass() {
+        try {
+            prepareDerbyDatabase();
+        } catch (ClassNotFoundException e) {
+            e.printStackTrace();
```
StephanEwen: Call Assert.fail() here to mark the setup as failed.

pact/pact-common/src/test/java/eu/stratosphere/pact/common/io/JDBCInputFormatTest.java
```diff
@@ -0,0 +1,194 @@
...
+    }
+    
+    @Test(expected = IllegalArgumentException.class)
+    public void testInvalidDBType(){
+        jdbcInputFormat = new JDBCInputFormat("idontexist.Driver", "jdbc:derby:memory:ebookshop", "select * from books;");
+        jdbcInputFormat.configure(null);
+    }
+
+    @Test
+    public void testJDBCInputFormatSettingPactRecordNormally() throws IOException {
```
StephanEwen: Can you check here for expected contents? The type check is good, but the contents should also match ;-)

If you make the contents a constant String[][] or Object[][], then you can have a loop that takes each row and compares it to the constant.

pact/pact-common/src/main/java/eu/stratosphere/pact/common/io/JDBCInputFormat.java
```diff
@@ -0,0 +1,351 @@
...
+     * Stores the next resultSet row in a PactRecord
+     *
+     * @param record target PactRecord
+     * @return boolean value indicating that the operation was successful
+     */
+    @Override
+    public boolean nextRecord(PactRecord record) {
+        try {
+            resultSet.next();
+            ResultSetMetaData rsmd = resultSet.getMetaData();
```
StephanEwen: I think you can skip the three lines here. You do not need to set the number of fields, the record will adjust as you set fields.

StephanEwen:
Nice work!

I added a few comments, but otherwise it looks good to me!

pact/pact-common/src/main/java/eu/stratosphere/pact/common/io/JDBCInputFormat.java
zentol: yup i checked, the record adjusts automatically, we could skip line 335.
but we still need column_count as an upper bound for the loop.

we should rename it to columnCount aswell...

rmetzger:
Hi guys,

once https://github.com/stratosphere/stratosphere/issues/289 is merged, can you create a new maven project in `stratosphere-addons` for the jdbc input (/output) format? Call it `stratosphere-jdbc`.

@StephanEwen: I guess this implementation of the  `JDBCInputFormat` will run with multiple input splits. Is there a way to set the DOP of the `GenericDataSource` from inside the InputFormat?
Or is it better to overwrite `createInputSplits()` ignoring the minSplits argument?
Or is a `SingleInstanceDataSource` or so required?

rmetzger:
Hey guys,

we now merged the pull request for "stratosphere-addons" (https://github.com/stratosphere/stratosphere/tree/master/stratosphere-addons). 
Could you create a new maven project under `stratosphere-addons` called `stratosphere-jdbc`.  Move the example and JDBC Input format into the new maven project.

zentol:
we will create the maven project before the end of this week.

StephanEwen:
I assume that the data base input is a task that can not be parallelized, i.e., there is a single instance only of the data source that issues the query, and which then distributes the data among all parallel tasks of the next stage.

In order to tell the system to not parallelize an input task, you can mark the input format as unsplittable by implementing the marker interface https://github.com/stratosphere/stratosphere/blob/master/pact/pact-common/src/main/java/eu/stratosphere/pact/generic/io/UnsplittableInput.java

If that is the case for your input format, please add the UnsplittableInput interface to the list of implemented interfaces.

ankurcha:
Pardon my ignorance, but I think it should be possible to parallize data load on a jdbc input. The idea would be to define the correct shard key(s), which may be left for the user to specify(abstract method). 

That being said, by default we could make it a single partition or maybe based off the primary key (if any). But allowing the user to specify the column to use to split the dataset may be very powerful that can have real scalability and concurrency advantages.

The UnsplittableInput route is definitely more simple but this would also mean that if for some reason any part of the dataload fails, we would end up needing to reload the full dataset, whereas in a parallelizable case, only that partition /split would need to be retried.

fhueske:
You are right. That would work for parallel database systems, that can efficiently run queries on individual data partitions without having a central master involved. 
If however: 
- The database system is not running in parallel, the same machine is asked many queries from different clients. A single query that asks for all data is probably more efficient than multiple queries asking for subsets. (Seq. IO vs. random IO, query compilation, locking, etc. overhead).
- The database system is parallel but with a master coordinator, this master needs to serve all clients the results (even though it does not need to compute the result). So the master might become the bottleneck here.
  Please correct me if you think that's wrong ;-)

I think, this really depends on the database setup and the feature of the DBMS. Furthermore, the programmer needs to know the setup quite well (partition keys and ranges).
But I agree, for parallel DBMS, it would be good to have the option to shard queries over data partitions.

zentol:
Would it be possible to let the user supply multiple queries, and use each of them in a separate split?

Given a user who knows about shard keys, i feel like it would be really inconvenient for him to pass the query and keys separately, und hope that the correct query is generated.

rmetzger:
Regarding your question: Yes. You could overwrite the getInputSplits() method (that would for example return two splits, one for each query).
I think we should first merge the current JDBC input format into the main line. Its nicely written and covers most use-cases.

If one wants to do more advanced stuff, there is also the way to create multiple DataSources with JDBCInputFormats. Like this.

```
+---------------+
| MySQL:Shard1  |
+---------------+
                  \
                   (Union (Stratosphere))--(Further processing)--Data Sink
                  /
+---------------+
| MySQL:Shard2  |
+---------------+
```

ankurcha:
That seems for to me.

zentol:
We ran into a problem when using the UnsplittableInput interface: the example stops working.
"Cannot compute input splits for Data Source: java.lang.NegativeArraySizeException" was the error message we got.

I was able to resolve it by overriding the createInputSplits method, with one that returns a list of a single GenericInputSplit().

rmetzger:
Hi @zentol;
I'm quite sure that this is a bug inside the system, so you don't have to do any workarounds. @StephanEwen can probably fix it within seconds ;)
I'll ask him tomorrow to fix it.

StephanEwen:
Manually merged (because of conflicts and renaming) in 4d73ef004199fb933bf4669da63e4254ad115c3f

Thanks guys, that was good work!

PR #280, opened 2013-11-21
Added integration test case for AllReducer and iterations

andrehacker:
This is the test case for issue [107](https://github.com/stratosphere/stratosphere/pull/107)
It is a simple scenario with iterations, dop>1 and with an keyless reducer. Should I check for anything else?

This test case will fail until (DOP < Iteration DOP) is allowed.


StephanEwen:
Merged in a4b1dbdb17c1df3348db362cafd185c326652158

PR #283, opened 2013-11-22
Extend website front page

rmetzger:
Hi,

This is the preview of the website changes I suggest here: http://robertmetzger.de/stratosphere/

Please do not merge this pull request! I think we need to rework some texts. 

I'm also thinking about adding some more pictures  and linking to some website subsections (quickstart)


uce:
Perfect!

PS: I also think that we have to change some texts (and maybe adapt some graphics from the Big Data Beers slides)

PPS: Lets incorparate the changes by Sunday

rmetzger:
Thanks. (But honestly: You created all the nice pictures, so most of the credits are for you)

I added a new commit (the preview also changed)
I also added more graphics from the slides (they are only visible if the user explicitly opens them ;) ).
Remember that I won't be available tomorrow (=sunday ;) )

We certainly need to rework some texts (but we should keep them as simple as possible).

Once we have new performance results, we need to add a performance section on the front page.

fhueske:
I like the Features section a lot. All important aspects are covered. Nice work!

I also have a few comments:
- The layout could be improved in my opinion. The two column layout made me first read the first column and then the second. However, the content is separated by rows not by columns. But maybe that's just me...
- The font size should be unified (Features in Optimizer section).
- I would skip the sentence "Its functional nature allows to parallelize the execution". In the context of the Scala Pact interface,  the parallelization comes from the programming model not Scala's functional nature (Java is parallelized as well without being functional...).
- In the optimizer section, I would refer to relational query optimizers. Also highlight, that the optimizer relieves the programmer from thinking about the efficient execution of the data flow. The features should cover: 
  - Cost-based choice of operators and shipping strategies. 
  - Reduction of shipped and written data volume. 
  - Pipelining of operations. 
- Instead of "Known from Hadoop" and "New in Stratosphere", I would simply make it "Hadoop" and "Stratosphere" and give all operators (incl. Map and Reduce) in the "More Operators" section. That would make the difference more clear.

rmetzger:
Thank you for your feedback! I Incorporated all your suggestions (please tell me if you're still unhappy with the Optimizer section!)

uce:
- **More Operators**: I would vote for only presenting our operators with the graphics and keep the comparison text-only. In the text I would say _the known MapReduce programming model_ instead of paradigm.
- **Advanced Data Flow Graphs**: "directed acyclic graphs (DAGs)" is a direct contradiction to the figure, which includes an iteration.
  - Capitalization of "Complex data flows in Hadoop"
  - The more detailed explanation says "stored to HDFS after each job". Isn't it stored _in_?
  - Also "around 30 seconds just for setup tasks". Shouldn't it be "just for setup"?
- **Powerful programming interfaces**: Capitalization and I would skip _Scalable Language_ and remove the link to scala-lang.org.
- **Support for Iterative Algorithms**: "distributed database" is confusing. Let's say "inside the system"?
  - Iterations with Hadoop: the only in "Data is only" is wrong placed?
- **Build-In Optimizer**: "For example the "Join" operator has two strategies available: Sort-merge-join and hybrid hash join." is suddenly in-depth compared to all the Hadoop stuff. Let's just say that we choose join strategies?
- **Open Source Community and Support**: The link "Open an Issue on GitHub" should link to the issues page.
- No new line before "We also have binaries that are ready to run on your server."

Then I would add some emphasis to the text by making keywords bold.

Other than that: P E R F E C T! :+1:

PS: The detailed explanations are a really good idea. :8ball:

rmetzger:
Operators: what do you say about this approach?
Advanced Data Flow: What is the "official" answer to that contradiction? I also noticed it. We can state that an an iteration is just another iterator, that internally has cycles? Or we say "DAG (except for the explicit definition of iterations" ?
Scala. Removed Scalable Language. Why removing the link? I made the link open on a new site, so that people hopefully find our tab still open after learning Scala ;)
Optimizer: the example is specific but that's why I called it an example. I just wanted to have the two join strategies somewhere on the front page. If you have a better idea: go for it.

I made some keywords bold. I like it for the lower part, in the upper part, I did not find highlight-able keywords. 

Do you think I should add the JobManager/TaskManager image as an additional (toggleable) explanation to the stack?
I asked @aalexandrov for performance results.

uce:
Operators: Really nice!

Data Flow: I think it would be best to just use _advanced data flow graphs_ as a replacements for DAGs.

Scala: Because of the sending people away issue, which might be OK with the new window. For the future: it will be nice to send them to a Scala guide on our page.

Optmizer: OK.

JM/TM: Yes, sure.

Emphasis: OK, I see. Maybe keyword is too specific. We could also try whole (first) sentences. Bold headlines would also be nice I think.

GitHub issue link: let's just link to the issues list and not directly to the "new issue" form. So people will see the current issues and open one if they want to.

aalexandrov:
I want to suggest some modifications directly as a push request. @rmetzger: Is the latest version pushed to your repository?

aalexandrov:
I'm pushing my changes right now, should be visible in 10-15 minutes. I did some minor improvements on the presentation and the HTML and went through the text up until System Stack.

My only remaining concern is that towards the end the descriptions become a bit heavy. I'll try to see if something can be shortened tomorrow. 

From the presentation perspective, I suggest to 
1. use a pop-up for the detailed explanations and 
2. re-format the feature descriptions either as either
   - a rolling slider with fixed structure (title + 3 line explanation + image) or 
   - text flow with h3 titles, paragraph-ed text and "float: left" pictures (similar to what we have now).

Other then that everything looks and reads nice, good job Robert.

StephanEwen:
Hey! Can we have a look together over the page on Monday before we push
anything?
Am 23.11.2013 21:02 schrieb "Alexander Alexandrov" <notifications@github.com

rmetzger:
Do you see major points?
I would like this to be pushed asap ;)
(Alex did not push to our website, he pushed to
http://aalexandrov.github.io/stratosphere/ )

On Sat, Nov 23, 2013 at 9:09 PM, Stephan Ewen notifications@github.comwrote:

> Hey! Can we have a look together over the page on Monday before we push
> anything?
> Am 23.11.2013 21:02 schrieb "Alexander Alexandrov" <
> notifications@github.com
> 
> > :
> > 
> > I'm pushing my changes right now, should be visible in 10-15 minutes. I
> > did some minor improvements on the presentation and the HTML and went
> > through the text up until System Stack.
> > 
> > My only remaining concern is that towards the end the descriptions
> > become
> > a bit heavy. I'll try to see if something can be shortened tomorrow.
> > 
> > From the presentation perspective, I suggest to
> > 1. use a pop-up for the detailed explanations and
> > 2. re-format the feature descriptions either as either
> > 3. a rolling slider with fixed structure (title + 3 line explanation
> > 4. image) or
> > 5. text flow with h3 titles, paragraph-ed text and "float: left"
> >    pictures (similar to what we have now).
> > 
> > Other then that everything looks and reads nice, good job Robert.
> > 
> > —
> > Reply to this email directly or view it on GitHub<
> > https://github.com/stratosphere/stratosphere/pull/283#issuecomment-29140375>
> > 
> > .
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/stratosphere/stratosphere/pull/283#issuecomment-29140500
> .

## 

rmetzger:
Okay, your version is online.
Here are my points: (sorry for the long list)

I'm also a bit unhappy with the current texts of the 6 overview features .. but there are some politics involved with that. 
- Beyond MR: People don't know what higher-order operators are. I want to keep it simple. (Especially in the first sentence)
-  Efficient Complex Data Flows: We don't want to be complex, we want to be advanced! (Note that I always write advanced when it comes to Stratosphere and complex when Hadoop ). I removed "arbitrary directed acyclic graphs" because the respective image is a cyclic graph! (Concerning arbitrary: I was told that "arbitrary" is not a nice word)
- I really liked the horizontal lines and the vertical spaces (including the headline (even though we do not only list features) ) (See also Fabian's concerns)

I like the idea of directly linking the hadoop solution in the text.
You are right about the text heaviness in the end. We need to change that.

What do you mean by "text flow with h3 titles, paragraph-ed text and "float: left" pictures (similar to what we have now)." ? 
I don't want to have a compact text on the website. People should not be scared by long texts. (That's why I used these blocks, switched the sides and the large spacings.)

I'm against using a slider because users have to click (and to understand the interface). I'm also against a pop up because it won't work on mobiles? Can google read popup's contents?. My approach only hides the advanced explanation if Javascript is enabled. So google's spider and users without JS will always see everything.

We work really hard to get people on the website, so we must do everything possible that they stay there and understand it!

aalexandrov:
"Can google read popup's contents?" Depends whether they are loaded from a hidden area of the HTML code or via AJAX.

"My approach only hides the advanced explanation if Javascript is enabled." Mine will do the same, its just a question of presentation to agents with JS support. 

Overall, do we really want to have all the detailed descriptions on the homepage? I would prefer a dedicated page with corresponding subsections explaining why Hadoop is bad and Stratosphere better with respect to dataflow graphs, iterations, etc., and brief feature descriptions linking to the corresponding technical subsection with an anchor.

Last, regarding sliding features vs text flow. The problem with a linear presentation is that it implicitly makes features placed higher more visible. A [carousel](http://getbootstrap.com/2.3.2/javascript.html#carousel) which changes the contents every 30 seconds time-share the visible space in the bottom half of page between all features (but can still degrade gracefully for non JS-users and phones).

Either way, we should try to have a predefined "template" with dedicated area for a picture, a title, and text, and have feature description text with similar weight across all features.

rmetzger:
Okay, lets meet on Monday to discuss the new front page.

ktzoumas:
Some changes in the text:

Very few use cases and real world application fit that model. --> There are several applications that do not naturally fit in this programming model.

Stratosphere is the only runtime that natively supports iterative algorithms inside the system. --> Stratosphere natively supports iterative algorithms in the runtime and optimizer.

Have the 6 grid elements be clickable and point to the corresponding explanations. 

As far as I am concerned, you can merge this, and then we can make a separate pull request for text changes.

rmetzger:
I added your changes.

Some of the 6 features now link to different sections below or somewhere else (I updated the demo page: http://robertmetzger.de/stratosphere/)

PR #284, opened 2013-11-25
Refactored Csv Input Format to be configurable via instance parameters

StephanEwen:


StephanEwen:
This pull request refactors the CsvInputFormats to:
-  Allow passing types through parameters, rather than the config
-  You pass types instead of parsers
-  It is generified such that the functionality is available to models other than pact record
  
  Please have a look and comment if you like the way it is used:
  
  https://github.com/stratosphere/stratosphere/pull/284/files#diff-c5813e5503590883623a27e0a1dc4b40R73 (MergeOnlyJoin.java, line 73);
  
  @aljoscha I had to adjust the Scala DataSource Macros. Can you double check that I this has no undesired side effects? It actually looks a lot nicer now ;-)
  
  @aljoscha The generation of the types filters by isUsed. If that prunes away fields, will the other fields not be in the wrong CSV column position? There is an alternate version (commented out) that replaces unused fields by null. It seems that one would do properly, as null type fields are skipped in the parser (keeping the other csv columns in place)

aljoscha:
I don't like that the field delimiter and the record delimiter parameters have differing types. The rest seems nicer though.

aljoscha:
Also, we should have a way to skip fields arbitrarily, i.e.

``` scala
CsvInputFormat[(Int, String, Int)](Seq(1,17,42))
```

to only read csv fields 1, 17, and 42.

In the java front end we need something similar for the constructor-style. Using the old style this was possible, right?

uce:
@aljoscha :+1:

I think (have to check) I've noticed a problem yesterday: in "real-world" CVS files you will usually find escaped commas, which don't act as delimiters, e.g. `field1,field2,"field,3",field4`. We should provide a simple solution for this by looking for opening escape symbols like " or ' and don't count the comma as a delimiter if the escape is still open.

aljoscha:
And also, I think the second way, you mentioned in the Scala CsvInputFormat is correct, since otherwise fields would wrongly be shifted around.

StephanEwen:
The parsers were written such that the field delimiter may only be a char.
If you pass a String with more than one char, it complains. So I changed
the type to prevent those situations. If we change the parsers to introduce
multi-char delimiters, we can change that...

On Mon, Nov 25, 2013 at 10:17 AM, Aljoscha Krettek <notifications@github.com

StephanEwen:
We can skip fields currently by passing nulls, e.g. (PactInteger, null,
PactString, null, null, PactDouble).

The old config style still works.

@ufuk; Strings in quotes work in order to skip delimiters. If you still
have cases where it did not work, please check the code in the var-length
string parser and adjust it.

StephanEwen:
Merged in 9dd4635ad69bf107ab9f81c8e7ffd495475471e0

PR #299, opened 2013-11-26
Add SingleInstanceDataSource for inputs that are not splittable

rmetzger:
This DataSource is required for https://github.com/stratosphere/stratosphere/pull/279 and the upcoming MongoDB connector.

### DIFF
--- a/pact/pact-common/src/main/java/eu/stratosphere/pact/common/contract/GenericDataSource.java
+++ b/pact/pact-common/src/main/java/eu/stratosphere/pact/common/contract/GenericDataSource.java
@@ -82,9 +82,6 @@ public class GenericDataSource<T extends InputFormat<?, ?>> extends Contract imp
 
 	// --------------------------------------------------------------------------------------------
 	
-	/* (non-Javadoc)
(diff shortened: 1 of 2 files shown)


rmetzger:
I had a little chat about this with @StephanEwen. We decided to add a method to the InputFormat `getMaxDOP()` to tell the framework the max dop.

I had an additional idea for this: We could also add a method to tell the framework a locality preference! it would really make sense to locate the DataSource on the same machine where the database etc. runs.
I guess the `SingleInstanceDataSource` is more suited for this since locality information should not be hardcoded into the input format.

StephanEwen:
c61bc993f9dd4825199c922086c5ed6c63e4c791 introduces a marker for input formats to declare their input as unsplittable (such as for database queries in most cases).

The long term alternative is to use the number of input splits as an upper bound for the degree of parallelism (we need to move the optimizer to the master server process before that though). Since a database query input format would typically create only a single input split, it would thereby declare that its input is unsplittable.

Should we still have the dedicated data source node suggested by this pull request?

rmetzger:
Not working

StephanEwen:
Fixed in 3e0d97416f761516f67e487ef997f4ec7a1d8fb5

PR #298, opened 2013-11-26
Extend csv format ctors

aljoscha:
This builds on #293, so only merge after that one is merged.

PR #301, opened 2013-11-27
Fix logging

aljoscha:
Make the scala schema printers use log4j

Fix the additional syserr appender in LocalExecutor: add a threshold so
that it does only output error messages

Make JobClient use log4j for the event output that was annoying me

This sits on #298, so only merge after that one's in.


StephanEwen:
Merged in 14a6cb93296ce1ae03425adafe918ed8d4ee473a

PR #302, opened 2013-11-27
fixed TableInputFormat returns no records when re-opened

mleich:
Currently TableInputFormat can only scan a single split.

The endReached flag is not reset when a new split is opened, so once the end of any split is reached, an instance cannot be used to scan another split (which is what Stratosphere tries) to do.

This sophisticated fix should resolve the issue.


StephanEwen:
Merged in b44349e8de1363077d4be2b96f558eccb26169e1

PR #290, opened 2013-11-26
Unified frontpage figures

fhueske:
All figures adapted to width of 800px. 
Uniform size of font (approx.) and data flow nodes.

PR #304, opened 2013-11-27
Fix typo on startpage and make sub headlines bold

uce:
1. @cboden noticed a typo on the front page: **constrainT MapReduce interface** should be **constrainED MapReduce interface**.
2. I've also put more emphasis on the headlines, which makes it imho easier to navigate the page.

### DIFF
--- a/index.html
+++ b/index.html
@@ -82,16 +82,18 @@ $( document ).ready(function() {
 
 
 <div class="container features-detail" style="padding-top:40px;padding-bottom:40px;">
-   <div class="row">
+  <div class="row">
     <div class="col-md-12">
-     <h1 class="text-center">Features</h1>
+     <h1 class="text-center"><strong>Features</strong></h1>
     </div>
-</div>
+  </div>
 
 <div class="container features-detail">
   <div class="row" id="feature_operators">
     <div class="col-md-12">
-      <h3>More Operators</h3>
+      <div class="page-header">
+        <h3><strong>More Operators</strong></h3>
+      </div>
     </div>
   </div>
   <div class="row">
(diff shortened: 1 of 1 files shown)


rmetzger:
Preview http://robertmetzger.de/stratosphere/ (with https://github.com/stratosphere/stratosphere/pull/290)

I don't like the `<hr>` (the line) below the headlines.?
You are separating the headline from its content? That does not make much sense to me.

uce:
Ah, sorry. I also had a preview. Forgot to post it.

This is the default CSS element for page headers, e.g. css class `.page-header`.

http://getbootstrap.com/components/#page-header

PS: putting an element like `hr` _by hand_ (to indicatae separate content) wouldn't make sense, but the line now actually is just layout and has no semantics (like separating content).

rmetzger:
Yes, but this is for the page header. Here, we have only a subheading.

And still, I don't like it ;)

On Wed, Nov 27, 2013 at 7:00 PM, Ufuk Celebi notifications@github.comwrote:

> Ah, sorry. I also had a preview. Forgot to post it.
> 
> This is the default CSS element for page headers, e.g. css class
> .page-header.
> 
> http://getbootstrap.com/components/#page-header
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/stratosphere/stratosphere/pull/304#issuecomment-29406001
> .

## 

fhueske:
I agree with Robert here.
I would have hr above the headline to have the individual sections clearly separated.
Otherwise a section's content and the headline of the next section form a visual block.

uce:
I agree (in this case :bowtie:). The preview at http://uce.github.io/ should be up to date any minute.

StephanEwen:
Are you merging this?

PR #306, opened 2013-11-28
Refactor scala input output

aljoscha:
This sits on #301, so only merge after that one's in.


StephanEwen:
Merged in cad8b4e717927b86f7b3499cc6cda8a8d28f2804

Issue #112, opened 2013-09-24
Union on DataSinks does not work


aljoscha:
When you have a DataSink with several inputs you get this error message:
13/09/24 13:48:09 ERROR client.JobClient: ERROR: eu.stratosphere.nephele.executiongraph.GraphConversionException: Job and execution vertex <Unnamed File Data Sink> have different number of inputs
    at eu.stratosphere.nephele.executiongraph.ExecutionGraph.createInitialGroupEdges(ExecutionGraph.java:409)
    at eu.stratosphere.nephele.executiongraph.ExecutionGraph.constructExecutionGraph(ExecutionGraph.java:282)
    at eu.stratosphere.nephele.executiongraph.ExecutionGraph.<init>(ExecutionGraph.java:176)
    at eu.stratosphere.nephele.jobmanager.JobManager.submitJob(JobManager.java:532)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:606)
    at eu.stratosphere.nephele.ipc.RPC$Server.call(RPC.java:417)
    at eu.stratosphere.nephele.ipc.Server$Handler.run(Server.java:946)

Fabian mentioned that having the compiler add a dummy mapper before the sink could be the solution.

fhueske:
Yes, that should be a quick fix. But AFAIK this would result in one additional thread and serialization overhead, since [DataSinkTask](https://github.com/stratosphere/stratosphere/blob/master/pact/pact-runtime/src/main/java/eu/stratosphere/pact/runtime/task/DataSinkTask.java)s cannot be chained behind [RegularPactTask](https://github.com/stratosphere/stratosphere/blob/master/pact/pact-runtime/src/main/java/eu/stratosphere/pact/runtime/task/RegularPactTask.java)s, due to the different abstract Nephele interfaces for processing and output tasks.

[...]

PR #307, opened 2013-11-28
DataSinkTask processes union input

fhueske:
Extended DataSinkTask to handle union input. Fixes issue #112.


StephanEwen:
Merged in 1a527afaf3bcda8099f96285884452f87186b1ca

Issue #275, opened 2013-11-18
BulkIteration member function is not implemented


lukacsg:
The implementation of the "public void setTerminationCriterion(Contract criterion)" member function in BulkIteration class is missing.

<code>
public void setTerminationCriterion(Contract criterion) {
                throw new UnsupportedOperationException("Termination criterion support is currently not implemented.");
        }
</code>

https://github.com/stratosphere/stratosphere/blob/master/pact/pact-common/src/main/java/eu/stratosphere/pact/generic/contract/BulkIteration.java

StephanEwen:
It is on the list for the coming release ;-) Until then, you can use
-  An absolute number of iterations, OR
-  The iteration aggregators to define a convergence criterion.

For the second option, have a look at the PageRank Example on how to use it:
-  An example aggregator (that collects many statistics about evolution of the solution): https://github.com/stratosphere/stratosphere/blob/master/pact/pact-examples/src/main/java/eu/stratosphere/pact/example/pagerank/PageRankStatsAggregator.java
-  An example convergence criterion: https://github.com/stratosphere/stratosphere/blob/master/pact/pact-examples/src/main/java/eu/stratosphere/pact/example/pagerank/DiffL1NormConvergenceCriterion.java
-  The creation of such an aggregator can be seen in the page rank plan construction (line 78) https://github.com/stratosphere/stratosphere/blob/master/pact/pact-examples/src/main/java/eu/stratosphere/pact/example/pagerank/DanglingPageRank.java#L78
-  The usage of the aggregator can be seen in the CoGroup that computes the new rank (lines 64 and lines 97-100)
  https://github.com/stratosphere/stratosphere/blob/master/pact/pact-examples/src/main/java/eu/stratosphere/pact/example/pagerank/DotProductCoGroup.java#L64

lukacsg:
Thanks! We will wait for the next release, while we can't define a convergence criterion or give an absolute number of iterations this time implementing our entity resolution algorithms. Should we leave the issue open?

StephanEwen:
Lets keep the issue.

I assume using the aggregators for convergence checking is not applicable to your algorithm?

sidlo:
We will see; at first we thought that our loop invariant cannot be implemented as a convergence criterion, but checking it again, it might work!

(We first thought that as a ConvergenceCriterion we must provide a measure that decreases in every step, and in our case, there is no such a measure. However, we have a simple condition of termination.)

markus-h:
[...]

Issue #123, opened 2013-10-03
Strange (and faulty) Iterations Behaviour


aljoscha:
In KMeansIterative, when you add an identity mapper after the reduceClusterCenter reducer you get:
java.lang.RuntimeException: : BUG: The task must have at least one output
    at eu.stratosphere.pact.runtime.task.RegularPactTask.registerInputOutput(RegularPactTask.java:248)
    at eu.stratosphere.nephele.execution.RuntimeEnvironment.<init>(RuntimeEnvironment.java:194)
    at eu.stratosphere.nephele.executiongraph.ExecutionGroupVertex.<init>(ExecutionGroupVertex.java:237)
    at eu.stratosphere.nephele.executiongraph.ExecutionGraph.createVertex(ExecutionGraph.java:497)
    at eu.stratosphere.nephele.executiongraph.ExecutionGraph.constructExecutionGraph(ExecutionGraph.java:275)
    at eu.stratosphere.nephele.executiongraph.ExecutionGraph.<init>(ExecutionGraph.java:176)
    at eu.stratosphere.nephele.jobmanager.JobManager.submitJob(JobManager.java:532)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:606)
    at eu.stratosphere.nephele.ipc.RPC$Server.call(RPC.java:417)
    at eu.stratosphere.nephele.ipc.Server$Handler.run(Server.java:946)
I discovered this one from a plan in the scala frontend where I also have a final mapper that does some real work after a reducer in an iteration (BulkIteration).

sscdotopen:
I suggest you open another issue for the second error as it seems to be a different thing. Could you also add a stacktrace from one of the machines for the case where you saw the stalling? I've seen stalling two, it can either be caused by deadlocks in the network layer or by the compiler failing to insert dams at the correct points.

aljoscha:
@StephanEwen I have added a failing test case in https://github.com/aljoscha/ozone/tree/chained-mapper-bug

StephanEwen:
Fixed in eec2840d831637e05a55a16bdd5a3e039a7d1ce0

aljoscha:
[...]

Issue #21, opened 2013-05-21
Implement deferred index updates Workset Iterations


StephanEwen:
Currently, the solution set must be updated for a certain key directly after it was accessed for that key. This is the fastest case.

While this covers a large portion of the cases, deferred updates (i.e. when affecting other partitions) may be required and increase the expressiveness of the model.

StephanEwen:
Implemented as of f73c583516aa042a5d92df3fc6c80649d1068a12

PR #315, opened 2013-12-03
Add deferred updates to workset iterations (issue #21) and fix chaining of iteration state updates (issue #123)

uce:
This pull request enables deferred updates for workset iterations (#21) and iteration state updates by chained tasks (#123).

---

**Deferred updates**: Before, workset iterations updated both the workset and solution set in the tail task. For connected components, this meant that the iteration tail (update components join) created both the workset for the next iteration _and_ updated the solution set.

  With this pull request, it will be possible to do the following:
- Intermediate workset update
- Intermediate solution set update
- (Separate) solution set tail update
- (Separate) workset tail update
- Unified solution set/workset tail update (as before)
  
  I extended `ConnectedComponentsNepheleITCase` with all possible variants. @vasia has a use case, where she further wants to test this (sorry, for letting you wait this long :bowtie:).

@StephanEwen: does it also make sense to have an intermediate task, which updates both sets and then has a tail later? It didn't think so and intermediate updates are exclusively workset or solution set updates. While posting this, I'm actually not so sure anymore. ;-)

@StephanEwen: I think it is correct that with the reprobing solution set update, both the build and probe side have to have the same type. Is this true?

---

**Chained iteration state updates**: If an iteration task has chained tasks, the last operator in the chain will do the state update.

  `IterationsWithChainingNepheleITCase` is a test case for this.

@StephanEwen: does it also make sense to have an intermediate chained driver to do the update and not just the last?

@StephanEwen: is it correct to assume that the last ChainedDriver has to have the same output type as the RegularPactTask it is used in?

---

When creating a Nephele JobGraph, you have to explicitly configure the type of update for the tail (before this the tail was always doing a workset update).

The relevant options are:
- setIsWorksetIteration()
- setIsWorksetUpdate()
- setIsSolutionSetUpdate()
- setIsSolutionSetUpdateWithoutReprobe()

Also the head has to wait for the solution set tail, if there is a separate one:
- setWaitForSolutionSetUpdate()

---

At some point, the overengineering bug bit me and I actually spent quite some time not doing anything productive with the code (aka this was basically finished a couple of weeks ago). There are quite some overlappings between the tail and intermediate task, which could be refactored. But the resulting code might be harder to understand. **Feedback wanted as I feel like not seeing the forest for the trees.**


uce:
@StephanEwen, I've addressed the issues we discussed last week:
- Tab formatting of `AbstractIterativePactTask`
- Removed IntelliJ's `.*` import optimization, which I didn't notice
- Call `config.getWaitForSolutionSetUpdate()` only once in the head
- Undo my renaming from `INSTANCE` to `SINGLETON` in the `*Broker` classes
- The answer to whether it makes sense to have a delegate for `FastSolutionSetUpdateCollector` is imho yes: an intermediate solution set join, which updates the solution set, but then has a separate workset tail

I've also merged your changes from 53bb5bda61b55b1a38806f0a60dd3ce0590c3fd7 (superstep abstraction in Nephele readers).

PS: I've force pushed the changes to my branch, so there are no new commits for the changes.

rmetzger:
Does the PR also fix https://github.com/stratosphere/stratosphere/issues/275 ?

StephanEwen:
It does not. The PR is purely about workset iterations...

StephanEwen:
Merged in 3ec7cae5561f12096fb23e1d6f54c41af5cfdaff

PR #317, opened 2013-12-03
Remove unserializable fields in UserCodeObjectWrapper.java

aljoscha:
Completely change this. Add new ClosureCleaner that is based on the spark ClosureCleaner but does things a bit differently.

This one now checks whether fields it removes are actually accessed. In UserCodeObjectWrapper now throw an exception if the user code object contains non-serializable fields.

Does someone know what to do about the code-copying? Because I took this from spark.

### DIFF
--- a/nephele/nephele-hdfs/pom.xml
+++ b/nephele/nephele-hdfs/pom.xml
@@ -41,6 +41,12 @@
 				<dependency>
 					<groupId>org.apache.hadoop</groupId>
 					<artifactId>hadoop-core</artifactId>
+                    <exclusions>
+                        <exclusion>
+                            <groupId>asm</groupId>
+                            <artifactId>asm</artifactId>
+                        </exclusion>
+                    </exclusions>
 				</dependency>
 			</dependencies>
 		</profile>
--- a/pact-scala/pact-scala-core/pom.xml
+++ b/pact-scala/pact-scala-core/pom.xml
@@ -36,6 +36,16 @@
 			<artifactId>scala-library</artifactId>
 			<version>2.10.3</version>
 		</dependency>
+        <dependency>
+            <groupId>org.scala-lang</groupId>
+            <artifactId>scala-compiler</artifactId>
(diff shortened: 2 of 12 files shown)


StephanEwen:
What do we actually gain here? Unnecessary serializable fields stay. Necessary non-serializable fields still crash the code and in a rather intransparent way, because users assume they have set the variable.

A clean solution would be to to an check about what fields are actually referenced and clean all others. If non-serializable fields remain, report an error.

aljoscha:
Yes but I have bigger fish to fry right now, so I wanted this as a quick fix for some cases.

StephanEwen:
I am a little worried that this makes problems hard to debug. If someone writes a program with a non-serializable data type in the closure, he does not get a nonserializable exception, but simply a null pointer at runtime. That is hard to trace back.

What do you think?

aljoscha:
Completely new, read updated description.

StephanEwen:
That is fine, if we handle the license correctly. Please make sure that:
- You write in the source file that it is in parts based on code from the spark project.
- You add to the NOTICE.txt file an entry for spark, similar to the ones that are in there.

StephanEwen:
Please remember to update the NOTICE file.

aljoscha:
Yes, yes, and also a notice in the code.

StephanEwen:
Merged in 039c0829cfd0b7741feb881e865b91958032d9dc

rmetzger:
I don't see a update in the NOTICE file: https://github.com/stratosphere/stratosphere/blob/master/stratosphere-dist/src/main/stratosphere-bin/NOTICE.txt

rmetzger:
Why did you close the PR without any comment?
I still think my comment regarding the missing entry in the NOTICE file is valid.

aljoscha:
I don't think I closed it on purpose. @StephanEwen said he wanted to merge anyways because of the restructuring. He could have just added the notice, though...

I didn't come around to adding it myself yet.

PR #319, opened 2013-12-04
Rich scala stubs

aljoscha:
Change Operators to accept stubs with open/close etc. You might call them rich stubs, but in the code it's just MapStub, FlatMapStub, FilterStub, ReduceStub, etc. They derive from MapStub etc. (the java one), so you can do things like getting the iteration context or what you might have...

This allows something like:

``` scala
.map( new MapStub[(String, Int), (String, Int)] {
  override def open(config: Configuration) = {
    println("Opening up this badboy.")
  }
  override def map(in: (String, Int)) = {
    println("IN RICH MAPPER: " + in)
    val (w, c) = in
    (w, c + 1)
  }
})
```

and likewise for flatMap and filter, for example:

``` scala
.map( new FlatMapStub[(String, Int), (String, Int)] {
  override def open(config: Configuration) = {
    println("Opening up this badboy.")
  }
  override def flatMap(in: (String, Int)) = {
    println("IN RICH MAPPER: " + in)
    in
  }
})
```

If people like this I will also make this available for all the other operators.

### DIFF
--- a/pact-scala/pact-scala-core/src/main/scala/eu/stratosphere/scala/stubs/MapStub.scala
+++ b/pact-scala/pact-scala-core/src/main/scala/eu/stratosphere/scala/stubs/MapStub.scala
@@ -21,7 +21,7 @@ import eu.stratosphere.pact.common.`type`.PactRecord
 abstract class MapStubBase[In: UDT, Out: UDT] extends JMapStub with Serializable with Function1[In, Out] {
   val inputUDT: UDT[In] = implicitly[UDT[In]]
   val outputUDT: UDT[Out] = implicitly[UDT[Out]]
-  lazy val udf: UDF1[In, Out] = new UDF1(inputUDT, outputUDT)
+  val udf: UDF1[In, Out] = new UDF1(inputUDT, outputUDT)
 
   protected lazy val deserializer: UDTSerializer[In] = udf.getInputDeserializer
   protected lazy val serializer: UDTSerializer[Out] = udf.getOutputSerializer
--- a/pact-scala/pact-scala-core/src/main/scala/eu/stratosphere/scala/stubs/MapStub.scala
+++ b/pact-scala/pact-scala-core/src/main/scala/eu/stratosphere/scala/stubs/MapStub.scala
@@ -1,0 +1,84 @@
+/**
+ * Copyright (C) 2010 by the Stratosphere project (http://stratosphere.eu)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+
+package eu.stratosphere.scala.stubs
+
+import eu.stratosphere.scala.analysis.{UDTSerializer, UDT, UDF1}
+import eu.stratosphere.pact.common.stubs.{Collector, MapStub => JMapStub}
+import eu.stratosphere.nephele.configuration.Configuration
+import eu.stratosphere.pact.common.`type`.PactRecord
+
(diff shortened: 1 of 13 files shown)


pact-scala/pact-scala-examples/src/main/scala/eu/stratosphere/scala/examples/grabbag/Grabbag.scala
```diff
@@ -62,12 +70,22 @@ object Main1 extends Serializable {
...
       .flatMap { l => l }
       .groupBy { case (word, _) => word }
       .combinableReduceGroup { _.reduce { (w1, w2) => (w1._1, w1._2 + w2._2) } }
       .map(fun)
+      .map( new MapStub[(String, Int), (String, Int)] {
+        override def open(config: Configuration) = {
+          println("Opening up this badboy.")
+        }
+        override def map(in: (String, Int)) = {
+          println("IN RICH MAPPER: " + in)
```
rmetzger: this looks like debugging code. Is it intended to have this in the master?

aljoscha: Have you seen the rest of the file? :smile:

rmetzger: Okay, I took a look :poop:
Why is it in the code then?

aljoscha: Because someone merged it and no one complained so far... :smirk:

fhueske:
I think we need these rich interfaces. Otherwise, the Scala API is less expressive than the Java API which should not be the case.

pact-scala/pact-scala-core/src/main/scala/eu/stratosphere/scala/stubs/MapStub.scala
```diff
@@ -0,0 +1,71 @@
+package eu.stratosphere.scala.stubs
+
+import eu.stratosphere.scala.analysis.{UDTSerializer, UDT, UDF1}
+import eu.stratosphere.pact.common.stubs.{Collector, MapStub => JMapStub}
+import eu.stratosphere.nephele.configuration.Configuration
+import eu.stratosphere.pact.common.`type`.PactRecord
+
+abstract class MapStubBase[In: UDT, Out: UDT] extends JMapStub with Serializable with Function1[In, Out] {
+  val inputUDT: UDT[In] = implicitly[UDT[In]]
```
StephanEwen: Does it make sense to make these private, to an implementing user does not mess with them?

pact-scala/pact-scala-core/src/main/scala/eu/stratosphere/scala/stubs/MapStub.scala
```diff
@@ -0,0 +1,71 @@
...
+  val outputUDT: UDT[Out] = implicitly[UDT[Out]]
+  lazy val udf: UDF1[In, Out] = new UDF1(inputUDT, outputUDT)
+
+  lazy val deserializer: UDTSerializer[In] = udf.getInputDeserializer
```
StephanEwen: I guess you use a lazy val here to make sure it is not instantiated before being shipped and in its proper context.

We should replace this through some other mechanism at some point, because these vals are very frequently accessed. When mapping to generic records instead of pact record.

aljoscha: First I did not use lazy vals and did the initialization in the open method. But it can happen that a user overrides open() and forgets to call super.open(). In that case the code will fail with null pointer exceptions. Even I forget this at first and was wondering whether I had made a mistake.

StephanEwen: Yes, it is a valid workaround for the time. Is it possible to make them
private nonetheless?

aljoscha: I will make them protected, because I use them for generated MapStubs.

aljoscha:
This can now be merged, all operators have rich stubs.

StephanEwen:
Merged in c7febc1ff1c7d54313da29a07d2eaaf469aab252

PR #320, opened 2013-12-04
Accumulators

andrehacker:
DON'T MERGE !!!

This pull request has the purpose to discuss the API for Accumulators (this is how I call it right now). I want to talk with Stephan tomorrow about implementation so good to have your feedback.

Please note: I mean Hadoop-style accumulators here: Write in the UDFs and ONLY read at the end of the job in the driver. Goal is also to make results available in next iteration.

I implemented the test case [`AccumulatorITCase`](https://github.com/andrehacker/ozone/blob/counters/pact/pact-tests/src/test/java/eu/stratosphere/pact/test/counters/AccumulatorITCase.java) which shows how you can use accumulators from the UDFs (see the Map UDF). I implemented 4 accumulators which I think should be shipped (see in the [Accumulators package](https://github.com/andrehacker/ozone/tree/counters/pact/pact-common/src/main/java/eu/stratosphere/pact/common/stubs/accumulators)). I think we should support convenience methods for these because it should be as simple as possible from UDF. The user can implement it's own accumulator type.

About the status: The counters are working locally in the UDFs, but are not merged and sent to the JobManager and Client (for clarity I don't write about the plans how to do this here). And we will probably have to move many classes around (e.g. Accumulator needs to be available in Nephele). But we should discuss about the API first.

@StephanEwen: I could not/didn't want to use the existing Aggregator classes for some reasons. Anyway, I think it's better to think of a good interface first and then see how we can merge these two.
#### Some design decisions
- Introduced ExecutionResult for local and remote executor to get counter results (not working yet, only partially implemented)
- Accumulators are identified by name (initially not by Enum as in Hadoop)
- All accumulators share a single namespace (i.e. stored in a single HashTable)
- No need to register accumulators, they are created on the fly in the UDFs. This means that type conflicts are possible, e.g. different UDFs use accumulators with same name but different type. I already check for this.

So far, please let me know what you think, e.g. about naming, functionality. We must be able to integrate this in Scala. Our interfaces Accumulator and SimpleAccumulator are comparable to Sparks [Accumulable](https://spark.incubator.apache.org/docs/0.6.2/api/core/spark/Accumulable.html) and [Accumulator](https://spark.incubator.apache.org/docs/0.6.2/api/core/spark/Accumulator.html) classes. They have a [test suite for accumulators](https://github.com/apache/incubator-spark/blob/2ce200bf7f7a38afbcacf3303ca2418e49bdbe2a/core/src/test/scala/org/apache/spark/AccumulatorSuite.scala) showing their functionality and API.


rmetzger:
Cool. I like the quality and speed of your work!

I think you can move everything from the `common` packages to `generic` since they are not related to `PactRecord` (which is nice!)
I also like the accumulators you ship by default
Am I right that your example does not contain a "usercode" accumulator?
I'm not sure if it is a good idea to hard-wire the standard accumulators into the `RuntimeContext`

```
 IntCounter getIntCounter(String name);
 LongCounter getLongCounter(String name);
 DoubleCounter getDoubleCounter(String name);
```

maybe at most the longCounter() (because that's probably the most frequent use-case)

andrehacker:
Thanks robert

> I think you can move everything from the common packages to generic...

I moved it, but we have to move it again to some Nephele layer anyways (e.g. to do the merging in JobManager). Otherwise we would need a dependency from nephele to pact-common. We will see this when discussing about the implementation...

> Am I right that your example does not contain a "usercode" accumulator?

Very good point. Now I added a custom accumulator `SetAccumulator<T>` to the test case and realized two problems:
- The code of the custom accumulator will not be available in the JobManager since the jar-classes are not available there (afaik). So custom accumulators would need to be added as jar to the lib-dir of the server. But this would add security, because we would not run arbitrary user code on the JobManager (only admins can add jar-files to the JobManager).
- I couldn't use the generic `getAccumulator(String name, Class<? extends Accumulator<V, A>> accumulatorClass)` method to create a parametrized accumulator object (e.g. `SetAccumulator<String>`). So I added a new function `void addAccumulator(String name, Accumulator<V, A> accumulator)` where the user creates an instance of it's custom accumulator previously. I commented-out the previous interface (to have only a single interface). Disadvantage is that the user can forget to add the counter.
- There is another API option: Accumulators themselves could care for being added to the RuntimeContext: `IntCounter numLines = new IntCounter(this)` (Constructor argument has type `Stub` and thus the accumulator can access the runtime context). This way we force the user to add the counter, but it's less transparent. This involves some abstract class and additional logic, because the runtime is not initialized when the UDF members are being initialized.

> I'm not sure if it is a good idea to hard-wire the standard accumulators into the RuntimeContext
> maybe at most the longCounter() (because that's probably the most frequent use-case)

Ok, we can leave this open for now, easy to change...

andrehacker:
Update: Accumulators are working now, including custom accumulators.
Will talk with Stephan this week how to finalize. If anyone is interested to look at the code - don't hesitate.
The test case `AccumulatorITCase` and the new example `WordCountAccumulators` are good entry points (we can also remove the example later) .

@markus-h You could already have a look at ´´´AccumulatorManager``` to see how you can visualize the accumulators in web interface. There is a accumulator-history of the same size as the MemoryArchivist.

So far as an update...

fhueske:
I had a look at the WordCountAccumulators example. 
Looks really good!

andrehacker:
I will open a new pull request where everything is in a single commit

Issue #310, opened 2013-12-02
Merge "Spagel" - Stratosphere Pregel/BSP into main code line


rmetzger:
There is code to have an BSP-style interface for Graph processing with Stratosphere's iterations feature.

I suggest to call it "Spagel", for the following reasons:
- it sounds like s-pregel (stratosphere pregel)
- Spark named their Pregel-implementation bagel
- its a german word.

I would like to have this code merged into the project soon.
A master student is going do compare Stratosphere with Giraph, so we need that interface anyways.

aljoscha:
What is "Spagel"?

asteriosk:
I secretly hope it's this:
http://www.urbandictionary.com/define.php?term=spagel

No idea if this was the intention of Robert though!

rmetzger:
Oh, now I'm getting it. I misspelled Spargel (http://de.wikipedia.org/wiki/Spargel).

We should still keep the name.

aljoscha:
Which name? Spagel, the dirty sex thing. I'm in. :+1:

uce:
I like both. Initially, I was for SpaRgel, but maybe Spagel is better -- easier to spell and the dirty & sexy pun. On the other hand, it might be too similar to Bagel.

One way or the another: I also think that it is important to have the API soon! :+1:

ktzoumas:
Alternative name: Pregelsphere (but I do prefer Spargel)

uce:
@StephanEwen I think you wrote the initial API for this. Can you say whether we can use it in the near future? I would be happy to look at it, implement some tests and extend it if necessary. :stuck_out_tongue_closed_eyes:

StephanEwen:
Merged in f2657cca2b981c1ab1f685d45186896c51fb7676

It is called "Pregelsphere" now, but there are discussions to rename it to "Spargel" (Asparagus)

uce:
I don't like "Pregelsphere". Could we discuss the name? I think most people were for Spargel or Spagel.

:+1: Spagel, because SpaRgel is quite hard to pronounce in English.

ktzoumas:
I also like Spargel (for asparagus)

sscdotopen:
+1 for Spargel

rmetzger:
+1 for Spargel (I have to be for this name since I suggested it)

uce:
OK, let's go with Spargel then. :+1:  I can do the renaming:
- Rename the Maven module from pregelsphere to spargel
[...]

PR #321, opened 2013-12-05
Rename Pregelsphere to Spargel (issue #310)

uce:
This is issue #310.

Renamed the BSP API of @StephanEwen (maven Module `pregelsphere` and packages `vertexcentric`) to Spargel (both Maven module and packages). Spargel is German for asparagus.


uce:
I also wanted to add a test case, but had problems when adding `pact-tests` as a test-scoped dependency to the pom.

The test case is just an adoption of the existing ConnectedComponentsITCase to use `SpargelConnectedComponents` instead of `WorksetConnectedComponents`. If anyone has an idea what I am missing, I could add the extra commit here.

StephanEwen:
Do we always want to rename the packages to "spargel"? We can keep the name Spargel for the project/feature, but call the packages "vertexcentric".

rmetzger:
Why? Will there also be a "edgecentric" implementation?
I vote for renaming the packages as well.

uce:
I think spargel as a package name is better from a user's perspective (aka I agree with @rmetzger).

StephanEwen:
There is actually an "edgecentric" view, Vasia is implementing something
similar to that. There are also other possible views of graph processing,
such as matrix views, ...

uce:
The last commit is a look into our bright feature without the name pact :sun_with_face:  @aljoscha will add the Scala API in `*.spargel.scala`.

StephanEwen:
Merged in 696c8fe49dd644476014e9f1eea7c8832104efbb

Issue #325, opened 2013-12-08
JobManager webinterface does not work with YARN Tracking UI Proxy


rmetzger:
YARN's web interface allows to track the application master using a special tracking url.
It is a proxy that routes the requests to the right container (that of the application master).
Since the request urls of the web interface are somewhat hard coded. Currently the ajax requests look like `/jobsInfo` but the proxy urls look like `http://localhost:8088/proxy/application_1386532236961_0001/`

markus-h:
What do you mean by application master? So if I understand it right, the webinterface should be accessible when using a proxy, for example by using http://localhost:8088/proxy/application_1386532236961_0001/jobsInfo? So we'd need a new configuration value on the client side to specify a proxy url.

rmetzger:
This picture explains what the application master is in the yarn context:
https://f.cloud.github.com/assets/5525371/1298189/1ac85602-30f0-11e3-9fd5-cd3993b38ce3.png

So its basically a Jvm that runs the JobManager with some yarn management stuff.

> So if I understand it right, the webinterface should be accessible when using a proxy, for example by using http://localhost:8088/proxy/application_1386532236961_0001/jobsInfo? 

Exactly!

Do you think it is possible to do this without an additional configuration value? I mean I'm sure that it is possible to find out the current url?
There should be a standard solution to that problem? http://stackoverflow.com/questions/4765740/relative-urls-in-ajax-requests ?

Probably changing `/jobsInfo` to `jobsInfo` should do the job?

fhueske:
@rmetzger did you resolve this issue in 02f6328c09723afa0ce9a3e9bec6b72080aaa74a?

rmetzger:
This issue is not resolved.
The JobHistory link to the analzye page is still hardcoded and not compatible to yarn's proxy!

rmetzger:
I fixed the issue with commit: 5300f4b4b4cfdb00edbd5b4e87721640b0c7dc0c

PR #327, opened 2013-12-09
YARN + pact-client enhancements, fix stalling travis builds

rmetzger:
- YARN now also starts the web interface
- extended pact-client to nicely support remote execution using new option "remote"

```
Action "run" compiles and submits a PACT program.
  "run" action arguments:
     -a,--arguments <programArgs>   Pact program arguments
     -c,--class <classname>         Pact program assembler class
     -j,--jarfile <jarfile>         Pact program JAR file
     -w,--wait                      Wait until program finishes

Action "remote" is similar to "run" but allows to specify the JobManager connection
  "remote" action arguments:
     -a,--arguments <programArgs>   Pact program arguments
     -c,--class <classname>         Pact program assembler class
     -j,--jarfile <jarfile>         Pact program JAR file
     -r,--address <arg>             Hostname:port of JobManager
     -w,--wait                      Wait until program finishes
```

The webinterface starts but is not really usable from the YARN webinterfaces due to https://github.com/stratosphere/stratosphere/issues/325.

### DIFF
--- a/pact/pact-clients/src/main/java/eu/stratosphere/pact/client/CliFrontend.java
+++ b/pact/pact-clients/src/main/java/eu/stratosphere/pact/client/CliFrontend.java
@@ -50,9 +50,7 @@ import eu.stratosphere.nephele.util.StringUtils;
 import eu.stratosphere.pact.client.nephele.api.Client;
 import eu.stratosphere.pact.client.nephele.api.ErrorInPlanAssemblerException;
 import eu.stratosphere.pact.client.nephele.api.PactProgram;
-import eu.stratosphere.pact.client.nephele.api.PlanWithJars;
 import eu.stratosphere.pact.client.nephele.api.ProgramInvocationException;
-import eu.stratosphere.pact.common.plan.Plan;
 import eu.stratosphere.pact.compiler.CompilerException;
 
 /**
@@ -73,11 +71,12 @@ public class CliFrontend {
 	private static final Option HELP_OPTION = new Option("h", "help", false, "Show the help for the CLI Frontend.");
 	private static final Option VERBOSE_OPTION = new Option("v", "verbose", false, "Print more detailed error messages.");
 	
-	// run options
+	// run/remote options
 	private static final Option JAR_OPTION = new Option("j", "jarfile", true, "Pact program JAR file");
 	private static final Option CLASS_OPTION = new Option("c", "class", true, "Pact program assembler class");
 	private static final Option ARGS_OPTION = new Option("a", "arguments", true, "Pact program arguments");
 	private static final Option WAIT_OPTION = new Option("w", "wait", false, "Wait until program finishes");
+	private static final Option ADDRESS_OPTION = new Option("r", "address", true, "Hostname:port of JobManager");
 	
 	// info options
 	private static final Option DESCR_OPTION = new Option("d", "description", false, "Show argument description of pact program");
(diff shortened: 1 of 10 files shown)


fhueske:
How about merging both actions (run and remote) into one (run) with an optional paramter -r (I would vote to call it -h for host or -m for master, though).
If -r is specified, the job is submitted to the remote JM. If not, the client tries to submit the job to the JM specified in the local configuration.

What do you think?

rmetzger:
Yeah... you are right. Sometimes, I don't see the obvious. (But I have to say the implementation is "historically grown" ;)
I'll fix it.

rmetzger:
Okay. Changed according to @fhueske's suggestion.
Updated the documentation on the website in 5cf6e23bb31b7d1c2f4cd302a4bd5bf8c393345f

pact/pact-clients/src/main/java/eu/stratosphere/pact/client/CliFrontend.java
```diff
@@ -78,6 +75,7 @@
 	private static final Option CLASS_OPTION = new Option("c", "class", true, "Pact program assembler class");
 	private static final Option ARGS_OPTION = new Option("a", "arguments", true, "Pact program arguments");
 	private static final Option WAIT_OPTION = new Option("w", "wait", false, "Wait until program finishes");
+	private static final Option ADDRESS_OPTION = new Option("m", "hostname", true, "Hostname:port of JobManager [optional]");
```
fhueske: "m", "master"?

rmetzger: I prefer `hostname` because we decided against `master` `worker` naming.
I can't use `-h` because it is already taken by `help`. 
That's why I decided for this inconsistency (which are not uncommon)

fhueske: I don't like inconsistencies...
How about --jobManager?

rmetzger: so "-m" and "--jobManager" ? ("-j" is also taken!)

StephanEwen:
While you are at changing the cli, can we remove the -w option and make it always wait? I think that is intuitive, I don't see cases where you want that asynchronously running and could not use the OS way of moving a process to the background.

One less option is one step easier to start...

pact/pact-clients/src/main/java/eu/stratosphere/pact/client/CliFrontend.java
fhueske: Given the restrictions due to used parameter codes, I think "-m", "--jobmanager" (with or without capital m) is the best option so far. 
At least JobManager contains an "m" for manager. "hostname" can refer to any host and does in common terminology not include the port...

rmetzger:
I'm against this Stephan.
We should write the address to the JobManager webinterface there .. but a blocking submission client probably not the best user experience.

rmetzger:
Okay, I changed it as discussed with @fhueske. (also on the website).
I now print this information after job submission:

```
Job successfully submitted. Use -w (or --wait) option to track the progress here.
JobManager web interface: http://127.0.0.1:8081
```

StephanEwen:
@rmetzger Allright, if it is more prominent how to track status, that is fine with me. I was just comparing to the way it works in Hadoop. It always blocks there.

StephanEwen:
Merged in 06168008ad8d2523d9a951ef499eb1c22bcf2939

Issue #326, opened 2013-12-09
Create pact-client.bat and start-local.bat for Windows users


rmetzger:
Hi,

we need these two simple scripts (without log rotation, just write everything into the console) for windows users without cygwin.
So there should be a (documented) way to download, extract and run Stratosphere on Windows.

The scripts should be really simple. I guess it should be doable in less than 10 lines each.

markus-h:
I have a first working version of the two files. Actually they are really around 10 lines each.
In the current setting, when every output is just printed to the console, you need two command line windows to run a job (one for running stratosphere and one for submit jobs with pact-client). To stop stratosphere you have to use STRG+C. Is that good enough for now?

StephanEwen:
I guess that is a start. I assume that the config parsing, memory
extraction, etc are not really implemented there.

 We can have a look at Microsofts HD-Insights distribution to see how one
writes such scripts.

rmetzger:
Path to config directory should work.
But parsing heapsizes not.

Sent from my iPhone

On 09.12.2013, at 15:11, Stephan Ewen notifications@github.com wrote:

I guess that is a start. I assume that the config parsing, memory
extraction, etc are not really implemented there.

We can have a look at Microsofts HD-Insights distribution to see how one
writes such scripts.

On Mon, Dec 9, 2013 at 3:07 PM, markus-h notifications@github.com wrote:

> I have a first working version of the two files. Actually they are really
> around 10 lines each.
> In the current setting, when every output is just printed to the console,
> you need two command line windows to run a job (one for running
> stratosphere and one for submit jobs with pact-client). To stop
> stratosphere you have to use STRG+C. Is that good enough for now?
> 
> —
> Reply to this email directly or view it on GitHub<
> https://github.com/stratosphere/stratosphere/issues/326#issuecomment-30134129>
> 
> .

—
Reply to this email directly or view it on
GitHubhttps://github.com/stratosphere/stratosphere/issues/326#issuecomment-30134385
.

rmetzger:
Hi,

I tried this today with a windows xp virtual machine. I first opened a a console (`cmd.exe`) and started the `start-local.bat` file.

```
C:\Dokumente und Einstellungen\Robert\Desktop\stratosphere-0.4\stratosphere\bin>
java -Xms768m -Xmx768m -cp C:\Dokumente und Einstellungen\Robert\Desktop\stratos
[...]

PR #329, opened 2013-12-09
Simple batch scripts to run stratosphere under windows

markus-h:
See Issue #326. Two simple batch scripts for running stratosphere in local mode and to submit jobs.


rmetzger:
Could you also add this to the documentation in http://stratosphere.eu/docs/setup/local.html
(You need to make a separate pull request for that)

StephanEwen:
A question: Should we create a separate directory for the windows files? (bin/windows) or anything? Or leave all in the same dir? The later screws my shell command completion, but that's okay...

StephanEwen:
Merged in 5efd24167a5d249cf4942464818e7a029aba38a8

Issue #300, opened 2013-11-26
Combiner has same task name as Reducer


rmetzger:
I have job with two consecutive mappers. 

```
              Map2
             /
Input -- Map1
             \
              Reducer
```

It seems that Map1 has the name of Reducer. So the name of Reducer occurs twice.
The webinterface does not show that Map1 and Map2 are connected to each other. It looks like Input is connected to map2.
I'm not 100% sure about this bug but it seems weird.

fhueske:
Which webinterface are you talking about (job submission / optimization or execution)?
Can you provide a screenshot?

Some notes about chaining: 
- Output that is going to more than one task cannot be chained (so neither Reducer nor Map2 cannot be chained behind Map1).
- Chaining is an aspect of execution and chained tasks do not have names. They run under the name of their "parent" task.

rmetzger:
Okay, than it's probably not related to chaining.

Its the "Stratosphere Management Console" ;) (the job manager web interface).
I can not really provide a screenshot, because telekom (NDA).

fhueske:
I think, a combiner gets the same name as the reducer. 
Usually, combine tasks are invisible during execution, since they are chained behind the preceding task.
Since, the output is split, the Combiner cannot be chained behind Map1 and runs in an additional task.
Therefore, Map2 would be connected to Input (since, Map1 is probably the Combiner).

So the real mapping might be:
Input -> Map1
Map1 -> Combiner
Map2 -> Map2
Reducer -> Reducer

Would that explain the effect?

rmetzger:
Yes, this is very likely, since I just made my reducer combinable.

Is it possible to automatically add something like "(Combiner)" or so to the name?

fhueske:
Shouldn't be a problem. Will prepare a pull request.

fhueske:
Well... It's not so super easy as I thought.
But we need a more flexible task naming mechanism soon anyways...

fhueske:
Has been fixed in #331

PR #331, opened 2013-12-09
Update Nephele task names

fhueske:
- Names have prefix of Pact contract (Map, Red, Cmb, Mat, Crs, CGp, ...)
- All chained tasks are added to name of chain parent

Task names are only proposals. Let me know, if you prefer a different naming.

This pull request addresses #300

### DIFF
--- a/nephele/nephele-common/src/main/java/eu/stratosphere/nephele/jobgraph/AbstractJobVertex.java
+++ b/nephele/nephele-common/src/main/java/eu/stratosphere/nephele/jobgraph/AbstractJobVertex.java
@@ -53,7 +53,7 @@ public abstract class AbstractJobVertex implements IOReadableWritable {
 	/**
 	 * The name of the vertex or task, respectively.
 	 */
-	private final String name;
+	private String name;
 
 	/**
 	 * The ID of the vertex.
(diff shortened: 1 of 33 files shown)


rmetzger:
I'm not sure but I guess the full names are probably nicer: Map, Reduce, Compare, Cross.
Horizontal space is not a problem in the interface.

fhueske:
You're probably right. Might be a bit unintuitive... 
Cmb was meant to be an abbreviation for Combine and not Compare ;-)

Other opinions?

rmetzger:
Seems like I'm a bit tired. there is no compare operator ;)

fhueske:
Alright, I'll extend the names.

StephanEwen:
I am merging this for now, because of the pending code reorganization.

Asdding the names inside the operators that define properties (required/produced) is actually not the right place, as they have not a 1:1 correspondance to UDF operators. We should move that in the near future to a different place.

fhueske:
I know, but this is the place where combiners are generated. Before that Reduce is just Reduce. 

We can extend the naming code with a check for the Contract type, but I thought right now there is a 1:1 relationship between strategies and contracts and omitted the check. Which contracts share strategies at the moment?

PR #332, opened 2013-12-09
Allow version-specific documentation starting with v0.4

uce:
With the upcoming release we will need to have documentation for different versions of Stratosphere. @rmetzger already suggested to have this, when I did the initial overhaul of the documentation.

With this PR, the documentation is under `/docs/0.4` instead of just `/docs`. The docs root and `/docs/latest` redirect to the latest version, i.e. the upcoming 0.4.

Preview under http://uce.github.io.

I suggest to start with 0.4 as the initial open source version for which we provide documentation, as porting the old documentation will be too much work right and we will probably suggest to anybody who uses something older to move to 0.4 anyways.


fhueske:
:thumbsup:

rmetzger:
Nice work! Thank you!

The pull request contains `docs/0.4/.DS_Store`.
I like the redirects. We should always link to the `latest` version.
Is it possible to add a "version switcher" somewhere? Then, we could make two copies of this documentation.
one `pre-0.4 (unstable)`, one `0.4` and (after the release) `0.5-SNAPSHOT`. 
I think a prominent switcher to the versions would be good.

uce:
Oh, shit. I thought .DS_Store was     in gitignore. Will fix.

Yes, we need a version switch. Will add.

uce:
OK, I changed it.

Preview at http://uce.github.io.

This will break #334. @fhueske would have to change the page layout from `inner_docs` to `inner_docs_v04` and put the file into the `docs/0.4` folder.

PS: Startpage hasn't updated yet, but if you go to a subpage, its up-to-date.

Issue #330, opened 2013-12-09
Update JavaDocs for pact-common module


fhueske:
see title

PR #337, opened 2013-12-11
Fixed all JavaDoc warnings in pact-common.

fhueske:
Fixed all JavaDoc warnings in pact-common. 
I focused on fixing the build warnings but also added and updated some JavaDoc on the way.

Adresses issue #330.


StephanEwen:
Merged in cfd0ac3d20ef78cab8d0b191a6a383650bad625d

PR #340, opened 2013-12-12
Implemented accumulators

andrehacker:
Now everything is in a single commit and merged with the current master.

Old pull request with discussion: https://github.com/stratosphere/stratosphere/pull/320

### DIFF
--- a/nephele/nephele-common/src/main/java/eu/stratosphere/nephele/client/JobClient.java
+++ b/nephele/nephele-common/src/main/java/eu/stratosphere/nephele/client/JobClient.java
@@ -19,6 +19,7 @@ import java.io.IOException;
 import java.io.PrintStream;
 import java.net.InetSocketAddress;
 import java.util.Iterator;
+import java.util.Map;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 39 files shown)


rmetzger:
I tested the code on the Telekom cluster (2 nodes) and I got the same result as the keyless reducer (and Hive).
I needed to add a open() method and one line inside my reducer() .. so the whole thing is very usable.

I will open a pull request to andre that shows the accumulator results after the pact-client has finished (only if started with `--wait`)
It will look like this

```
rmetzger@hadoop01:~$ ./runPersons.sh 
Job Runtime: 85300
Accumulator Results: 
- numGroups (java.lang.Integer): 39486796
```

rmetzger:
Do you think it would be a cool addition to have a way to explicitly disable the accumulators?
I mainly use them for debugging, people probably want to disable them for performance reasons in some point?
(On the other hand,  it is nice to see them in operation to get an idea of your data)

Another nice thing (for debugging) would be something like Standard accumulators that count how many records go in and out each Operator. I currently add them manually, but ideally I would like to add something like
`implement ProfileOperator` to my operators.
`implement StandardAccumulator`  or as a property to the `MapOperator.enableAccumulatorProfiling(true);`
What do you think?

I don't know if it is a good idea that you can call getAccumulator() on non-existing accumulators.
I just had the case that I copy pasted my accumulators from some mappers to a match (for counting incoming and outgoing tuples) and I accidentally forgot to change the name. Having a "createAccumulator" method could throw a exception "accu already exists"
I now use this construct to avoid the mentioned problem:

```
cnt = getRuntimeContext().getIntCounter(getRuntimeContext().getTaskName()+": counter");
```

StephanEwen:
Should we have the WordCount with Accumulators packed into an example JAR? Or is the code mainly for illustrative purposes?

andrehacker:
First my reply to Stephan, since he is merging:
Currently I have a separate jar, but I think we don't need it. You can remove the section from the pom if you like.

fhueske:
I think we don't need this as a packaged example job. 
We should definitely keep the code and link it from the documentation.

PR #342, opened 2013-12-13
Code Projects Reorganization

StephanEwen:
Current Status:
- Added stubs for reorganized projects.
- Packages reorganized


StephanEwen:
Current project structure with dependencies:

picture1

StephanEwen:
Still a bug with the dependencies when using hadoop 2.x. I think it is a conflicting version of the ASM library.

StephanEwen:
Fix: Added ASM exclusion to hadoop dependency in hadoop v2 profile.

rmetzger:
The webinterface does not start properly

```
22:40:56,590 ERROR eu.stratosphere.nephele.jobmanager.JobManager                 - Cannot start jobmanager web info server. The directory containing the web documents does not exist: /home/robert/Projekte/ozone/ozone/stratosphere-dist/target/stratosphere-dist-0.4-SNAPSHOT-bin/stratosphere-0.4-SNAPSHOT/bin/../conf/.././resources/web-docs-infoserver
```

In the `bin.xml`:

```
        <fileSet>
            <!-- copy files for Jobmanager web frontend -->
            <directory>../nephele/nephele-server/resources</directory>
            <outputDirectory>resources</outputDirectory>
            <fileMode>0644</fileMode>
            <excludes>
                <exclude>*etc/users</exclude>
            </excludes>
        </fileSet>
```

This does not look so nice

```
eu.stratosphere:stratosphere-java-examples
```

What do you think about removing the `stratosphere-` prefix from all projects?

pact-array-datamodel still depends on pact-client, this causes maven to download those dependencies from sonatype ;)

```
<       dependency>
            <artifactId>pact-clients</artifactId>
        </dependency>
```

(use `mvn dependency:tree` to find those problems)

StephanEwen:
Can you base a pull request on my one and submit a fix?
Am 13.12.2013 22:42 schrieb "Robert Metzger" notifications@github.com:

rmetzger:
I'll open a PR tomorrow, ok. I'll update the previous comment.

StephanEwen:
I updated the assembly. It was sill pointing to pact-clients and nephele-server.

StephanEwen:
I am okay with removing the `stratosphere-` prefix from all artifact names.

uce:
Cool stuff! You really accomplished a lot! :)

The only name I am unsure about is the current Java API. Maybe let's add `java` to it? Like `java-record-api`?

StephanEwen:
For renaming, see #343

rmetzger:
I'm against adding empty projects into a version that is going to be released soon! (java api)

StephanEwen:
This is only to give a look at what it will be like. We can remove the stub project before we merge it into the main line.

StephanEwen:
This PR is subsumed by #343

StephanEwen:
Very well, the suggestions look good to me. I would actually remove the
java api for now.

One question I have been playing around in my head with its whether we
should use 'udf' instead of 'function'.
Am 17.12.2013 10:54 schrieb "Fabian Hueske" notifications@github.com:

fhueske:
I had copied and extended my comments to the updated PR #343 (this PR is the subsumed one).
I suggest you do the same to have everything in one place. Please check also my extended comments.

I don't have a clear preference whether to use function or udf. UDF is more specific but also DB-terminology and not every user might be able to make sense of that. If I have to choose, I'd pick function because its easier to understand. But as I said, UDF would also be fine...

I am closing this PR to avoid further confusion between the PRs, OK?

PR #343, opened 2013-12-14
Draft for renaming members of stratosphere-core

StephanEwen:
Until now, only `stratosphere-core` was renamed.

Comments welcome.


StephanEwen:
Preliminary refactoring protocol:

```
package names:

 eu.stratosphere.nephele.configuration      -->  eu.stratosphere.configuration
 eu.stratosphere.nephele.util               -->  eu.stratosphere.util
 eu.stratosphere.nephele.fs                 -->  eu.stratosphere.core.fs
 eu.stratosphere.nephele.fs.file            -->  eu.stratosphere.core.fs.local
 eu.stratosphere.nephele.protocols          -->  eu.stratosphere.core.protocols
 eu.stratosphere.nephele.io                 --> eu.stratosphere.core.io
 eu.stratosphere.nephele.services.memorymanager  -->  eu.stratosphere.core.memory


 eu.stratosphere.pact.common.util           -->  eu.stratosphere.util
 eu.stratosphere.pact.generic.stub          -->  eu.stratosphere.api.functions
 eu.stratosphere.pact.common.plan           -->  eu.stratosphere.api.plan
 eu.stratosphere.pact.common.io             -->  eu.stratosphere.api.io 
 eu.stratosphere.pact.common.io.statistics  --> eu.stratosphere.api.io.statistics 
 eu.stratosphere.pact.generic.io            -->  eu.stratosphere.api.io
 eu.stratosphere.pact.generic.types         -->  eu.stratosphere.api.typeutils
 eu.stratosphere.pact.common.distributions  --> eu.stratosphere.api.distributions
 eu.stratosphere.pact.common.type.base.parser  -->  eu.stratosphere.types.parser
 eu.stratosphere.pact.common.type              -->  eu.stratosphere.types
 eu.stratosphere.pact.common.type.base         -->  eu.stratosphere.types
 eu.stratosphere.pact.common.stubs             --> eu.stratosphere.api.record.functions
 eu.stratosphere.pact.common.io                --> eu.stratosphere.api.record.io
 eu.stratosphere.pact.common.contract          --> eu.stratosphere.api.record.operators 

 eu.stratosphere.pact.client.nephele.api       --> eu.stratosphere.client.program

 eu.stratosphere.pact.compiler.*      -->  eu.stratosphere.compiler.*
             compiler.plan            -->  compiler.dag
             ompiler.plan.candidate   -->  compiler.plan

 class names:

 eu.stratosphere.pact.common.io.RecordOutputFormat -->   eu.stratosphere.pact.common.io.CsvOutputFormat
 eu.stratosphere.pact.common.io.RecordInputFormat  -->   eu.stratosphere.pact.common.io.CsvInputFormat


 PactProgram  --> PackagedProgram
 *Contract    --> *Operator
 Match*       --> Join*

 eu.stratosphere.nephele.io.Record       -->   Removed and Replaced by eu.stratosphere.core.io.IOReadableWritable  (later to be replaced by Hadoop Writable, if we want easy dropin of hadoop types)


Moved classes:

 FieldSet
 FieldList
 PactConfigConstants           --> eu.stratosphere.configuration
 AsciiUtils                    --> eu.stratosphere.util
 IllegalConfigurationException --> eu.stratosphere.configuration
 StringRecord                  --> eu.stratosphere.core.io
```

rmetzger:
Cool!

I vote for renaming `stratosphere-scala-api` to `stratosphere-scala`. (same for `java-api`)
only `stratosphere-record-api` should end with `-api`.
`stratosphere-runtime` still contains a `nephele` in its package names?

StephanEwen:
The runtime packages should also be renamed, but I thought that less
critical (and postponed it). They are not part of a public api after all.
Am 14.12.2013 20:05 schrieb "Robert Metzger" notifications@github.com:

AHeise:
Sound good to me.

But why 
compiler.plan            -->  compiler.dag? Don't you have more than DAGs now.

I'm also not completely convinced by
 PactProgram  --> PackagedProgram
 _Contract    --> *Operator
 Match_       --> Join*

Why PackagedProgram? Sounds a bit like a jar ;)
Why Operator? It's not bad at all, but I still like to know the reasoning.
Match should be EquiJoin.

fhueske:
Commented in wrong PR, sorry...

I had a look at the PR and here are some comments:
- The overall structure looks very good and much more organized than before! :thumbsup:
- We should put a README file into the stratosphere-java-api module which tells the user that this will be a new Java API which is not done yet.
- stratosphere-runtime has a lot of "legacy"-terminology but that's not exposed to the user and can be changed later.
- The package structure of stratosphere-client looks a bit messy (same as before). Do we want to clean that up a bit as well? Shouldn't be too much effort since its only a few classes.
- The project is compiling with and without tests for me, but I did not check if the build artifact is working.
- I had a closer look at stratosphere-core module and propose the following changes:
  - eu.stratosphere.accumulators -> eu.stratosphere.api.accumulators
  - eu.str.api.function.GenericMatcher -> eu.str.func.GenericJoiner
  - eu.str.api.function.Stub -> eu.str.func.Function (or FunctionStub if Function is a key word in Scala?)
  - eu.str.api.function.StubAnnotation -> eu.str.func.FunctionAnnotation
  - eu.str.api.function.GenericStub -> eu.str.func.GenericFunction(Stub)
  - eu.str.api.operators.AbstractPact -> eu.str.op.AbstractOperator
  - eu.str.api.operator.Contract -> eu.str.api.op.Operator (??)
  - eu.str.api.operator.(SingleInputContract, DualInputContract, IterationContract) -> eu.str.api.op.(SingleInputOperator, ...)
  - eu.str.api.operator.base.(XyzContract) -> eu.str.api.operator.base.(XyzOperator)
  - eu.str.api.operator.util.StubAnnotationConfigurable -> eu.str.api.op.util.FunctionAnnotationConfigurable
  - eu.str.configuration.PactConfigConstants -> merge with ConfigConstants (??)
  - eu.str.types.(PactXyz) -> some other prefix. I prefer (SXyz: SInteger, SDouble, ...) but maybe someone has a better idea
- Proposed changes for stratosphere-record-api:
  - eu.stratosphere.api.record.function.MatchStub -> eu.str.api.rec.func.JoinFunction
  - eu.stratosphere.api.record.function.(XyzStub) -> eu.str.api.rec.func.(XyzFunction)
  - Shouldn't PactRecord be part of stratosphere-record-api instead of stratosphere-core? Or do you want to avoid the dependency of str-runtime on str-record-api?

StephanEwen:
@AHeise 

Here are the rationals behind the renamings. Please try to evaluate the names from the perspective of someone how does not know the code and concepts, yet, and needs to find an intuitive api.

compiler.dag is the optimizer's representation of the DAG. The iterations are a meta operator, they do not make the data flow cyclic. So I think dag should be accurate. optimizer.plan is the actually created optimizer plan.

In general, we tried to make names more intuitive , it at the cost of being less technically precise at times. The decision to move away from Contract came from the idea to move away from the terminology of pact, which were always a bit confusing. A new user has no idea what a contract is supposed to be, while operator is an approachable term.

The same is actually the rational for Match --> Join. Because match is an inner equi-join with a udf, just calling it join makes it more graspable, I think. It is true that EquiJoin is technically more correct, but since equi-join is the default, we kept it simple here.

PackagedProgram is really a self contained program jar, with optional parameters. In contrast to that is the PlanWithJars which describes a Plan (program) with optional jar attachments for the UDFs. We use them in different circumstances: The PackagedProgram is the jar that you can upload and simply invoke from the web client or command line client. The PlanWithJars is intended to be used with the Local/Remote Executors (and later with the shell-like interface), where you do not need to package the program to send it to the cluster for execution. You only need to attach libraries, if you used them.

rmetzger:
FullAck Stephan!
I had a long meeting today at the Telekom to discuss the Stratosphere Code I've written for them (a non trivial amount of code) and they totally agreed that "Contract" is very contra-intuitive, the same applies to all the Pact-prefixed stuff.
Someone of them also told me that he quickly looses interest in reading texts about Stratosphere because we introduce so many new names for our stuff (nephele, pact, Sky?, contract, ...).

aljoscha:
Could we please rename SequentialInputFormat (and the output format). I know what it does, but I would never have guessed that from the name.

For the java record api RecordInputFormat would be a good name since it does exactly that, read records from the binary stream using their read() method. I do realize, however, that the name is a bit tainted because CsvInputFormat used to be called that (is still). It't still a good name for what it does, though.

For scala I propose ObjectInputFormat, because that is exactly how it behaves. If you had a sink with `ObjectOutputFormat[(Int, String)]` you could read the output files with a source of `ObjectInputFormat[(Int, String)]`.

AHeise:
eu.stratosphere.api.common.operators.util.ContractUtil should be renamed to OperatorUtil
what about renaming JoinFunction#match to join?

Issue #286, opened 2013-11-25
Writing to local file system in distributed operation does not work properly


rmetzger:
I have a task that runs on two cluster nodes.
The output is set to "file:///home/rmetzger/cmpDiscount/", the `FileDataSink` has a DOP of 64.

After the job has finished, on host1, there is a directory in `/home/rmetzger/` called `cmpDiscount/` with 32 files inside.
On host2 on the other hand, there is only one file in my home directory, called `cmpDiscount`. So only one of the 32 threads on host2 writes into the file system?

StephanEwen:
Did the job produce any errors? Did the log mention failed operations?
Am 25.11.2013 11:21 schrieb "Robert Metzger" notifications@github.com:

rmetzger:
With INFO logging level, there are no warnings, nothing fails.

rmetzger:
Even though the file exists, I got this exception (I guess its the same bug)
 (Just for the one who is going to fix this)

```
21:05:43,844 ERROR eu.stratosphere.nephele.taskmanager.runtime.RuntimeTask       - java.lang.RuntimeException: Stream to output file could not be opened: Opening the file output stream failed: File file:/home/rmetzger/cmpAge does not exis
t.
        at eu.stratosphere.pact.generic.io.FileOutputFormat.open(FileOutputFormat.java:115)
        at eu.stratosphere.pact.common.io.RecordOutputFormat.open(RecordOutputFormat.java:167)
        at eu.stratosphere.pact.runtime.task.DataSinkTask.invoke(DataSinkTask.java:163)
        at eu.stratosphere.nephele.execution.RuntimeEnvironment.run(RuntimeEnvironment.java:344)
        at java.lang.Thread.run(Thread.java:679)
Caused by: java.io.IOException: Opening the file output stream failed: File file:/home/rmetzger/cmpAge does not exist.
        at eu.stratosphere.pact.generic.io.FileOutputFormat$OutputPathOpenThread.waitForCompletion(FileOutputFormat.java:202)
        at eu.stratosphere.pact.generic.io.FileOutputFormat.open(FileOutputFormat.java:112)
        ... 4 more
Caused by: java.io.FileNotFoundException: File file:/home/rmetzger/cmpAge does not exist.
        at eu.stratosphere.nephele.fs.file.LocalFileSystem.getFileStatus(LocalFileSystem.java:106)
[...]

PR #363, opened 2013-12-20
#286: Fixed distributed writing to local filesystem

fhueske:
This fix enables local filesystem writes in a cluster setup.
Output directories for local FS writes are created on each TM but not on the JM (except if it hosts a TM as well).

The currently implemented override policy for local filesystem writes is:
- existing files is not deleted to create output directories (to avoid deleting OS files)
- existing files in existing output directories are overwritten

Let me know, if you want to enable deleting local FS files to create output directories.


stratosphere-runtime/src/main/java/eu/stratosphere/pact/runtime/task/DataSinkTask.java
```diff
@@ -385,43 +385,52 @@ public int getMaximumNumberOfSubtasks()
...
-			try {
-				final FileStatus f = fs.getFileStatus(path);
-				if (f == null) {
-					return 1;
-				}
-				// If the path points to a directory we allow an infinity number of subtasks
-				if (f.isDir()) {
-					return -1;
+			
+			if(fs.getUri().getScheme().toLowerCase().equals("hdfs") || fs.getUri().getScheme().toLowerCase().equals("s3")) {
```
rmetzger: I'm not sure if this is the correct place for this.
I thought that file systems are pluggable. This here looks like hardcoded rules for specific file systems.

is there a way to move this code into the DistributedFileSystem (and S3) implementation?

The `FileOutputFormat` also contains specific `file:///` scheme code. I'm think In/Outputformats are independent of the filesystems (e.g. the file systems provide a data stream and the IF/OF converts it into an internal representation?)

fhueske: You are certainly right, that this is not the optimal place. In fact, most of that was already present and I only added a guard for distributed FS at this place. I was looking for a quick fix that doesn't break anything shortly before the release and didn't want to change too many things.

Now with the release out of the way, I can look for a proper solution based on file systems. However, the distinction between distributed and local file system will most likely remain in the DataSinkTask and the FileInputFormat as the distributed FS should be initialized at planning time (within the JM) and local FS need to be initialized in parallel at runtime...

rmetzger: Okay. I also think that it is a better idea to implement it correctly (e.g. no quick fix).

For the initialization at planning time: Can't you just add a new method to the FileSystems class for that?

fhueske:
New PR for this: #421

PR #371, opened 2013-12-24
Rework download page & versions as variables

rmetzger:
We download page now points to the most recent stable release (0.4-rc1)

I also introduced variables for the current stable and snapshot versions.
(I guess we will have some 0.4.1, ... releases soon).

Btw: I promised to work on Christmas!


uce:
Merged in 4b21ac50f638da80d9fd6728f75bdc9ca5274be5.

All relevant version-specific  download links etc. are in `_config.yml` now.

PR #374, opened 2014-01-06
Add scala programming guide

aljoscha:
This is not yet complete, but almost. The sections about rich stubs and execution are missing.

You can have a look at it here: http://aljoscha.github.io/stratosphere/docs/0.4/programming_guides/scala.html

Sorry for the delay, it turned out to be quite a bit of writing. :sweat_smile:

### DIFF
--- a/docs/0.4/programming_guides/scala.markdown
+++ b/docs/0.4/programming_guides/scala.markdown
@@ -1,20 +1,801 @@
 --- 
 layout: inner_docs_v04
 title: Scala Programming Guide
+sublinks:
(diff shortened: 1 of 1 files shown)


StephanEwen:
Looks good for a start.

A few quick comments at a first glance:

1) Please rename "Usable Data Types" to "Data Types". Sounds simpler.
2) In that same Section, Key and Value are confused.
3) Please reorder the input formats such that Text, Csv, and Delimited come
first (I expect those to be the most interesting ones).

uce:
I will merge this to get an overview of the current state of things on the site.

PR #381, opened 2014-01-08
Polling for jobmanager webinterface

markus-h:
I added polling functionality to the webinterface of the jobmanager, so no more pressing of F5 is neccessary and bandwidth is saved. The script is polling for changes every 2secs.
I tested it on FF26, Chrome 31 and IE 11 (on windows).

### DIFF
--- a/stratosphere-runtime/resources/web-docs-infoserver/js/jobmanagerFrontend.js
+++ b/stratosphere-runtime/resources/web-docs-infoserver/js/jobmanagerFrontend.js
@@ -4,30 +4,68 @@ var widthProgressbar = 120;
 // For coloring the dependency graph
 var colors = [ "#37485D", "#D9AADC", "#4F7C61", "#8F9C6A", "#BC8E88" ];
 
-(function poll() {
-	$.ajax({ url : "jobsInfo", type : "GET", success : function(json) {
+var timestamp = 0;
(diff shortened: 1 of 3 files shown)


stratosphere-runtime/resources/web-docs-infoserver/js/jobmanagerFrontend.js
```diff
@@ -4,30 +4,68 @@ var widthProgressbar = 120;
...
+	}, dataType : "json",
+	});
+};
 
+/*
+ * Polls the job history on page load
+ */
 (function pollArchive() {
-	$.ajax({ url : "jobsInfo?get=archive", type : "GET",
+	$.ajax({ url : "/jobsInfo?get=archive", cache: false, type : "GET",
```
rmetzger: I intentionally removed the leading `/` at the URL because otherwise the YARN proxy would not work anymore.

markus-h:
I fixed the YARN proxy compartibility and a small bug concerning duplicates in the job history. Since there were some problems testing the new polling functionality a small reminder: Very often the browser cache has to be cleared to test new javascript files.

StephanEwen:
Merged in e8194b0d350dc9344c7ac07d4ee218a3dedf4364

Issue #66, opened 2013-08-21
Attach output of a contract as additional input for another contract (Broadcast variables / Closures)


andrehacker:
Another PACT API feedback from user perspective.
### Requirement
- **Make the output (all records) of one or more contracts available in every udf call of another contract.**
- This should work in an efficient way. Using the Cross-Workaround, we receive the same pactrecord for every udf call and the first line is usually a deserialization (we cache the value, but then need an additional check if we already cached).
- It should be as explicit as possible to use this functionality (broadcast/attach). In my opinion it is a very common pattern. This gives the system more knowledge about our intention and more options for optimization.
### Problem
- **The often proposed workaround, to use Cross, only works for the special case where 1) the input to broadcast consists of a single record and 2) we want to broadcast the output of a single contract only and 3) it only works for Map, not for Reduce, CoGroup or Match**
- Using Cross for broadcasting of a single output record feels more like a workaround than a natural fit (especially because it works only for a very special cases) Broadcasting, distributing or attaching feels much more natural.
### Solutions
1. Implement a Distributed cache in the hadoop way. It has functionality to read or write files.
2. Provide generic workaround contracts (like "flatten" multiple records to a single one).
3. Add explicit support to broadcast and attach the output of a previous contract to a contract as an additional input. Before the udf is called, another method should be called with all the inputs so that the udf can deserialize the records, optionally customize them and store in-memory.
   I didn't think too long about this, but a api (here for Map) could look like this:

```
MapContract myMap = MapContract.builder(MyMap.class)
  .input(myDataSource)
  .attachAsAdditionalInput(contract1ToAdd)
  .attachAsAdditionalInput(contract2ToAdd
  .build()
```

And a new method in MapStub that is called before  the first `map()` call

```
public void additionalInputs(Iterator<PactRecord>[] iterators) {
  // read records from additional inputs, process and cache in-memory 
}
```
### Discussion
- In Hadoop the usual way to solve this is the distributed cache. This is very powerfull and flexible, but is probably not a good fit for this requirement. The output is not stored as a file (as it is for hadoop) and we want to pipeline it.
- I currently use workaround contracts, but this is just a (slow) workaround and not a good long term solution in my opinion
- I would vote for adding explicit support - probably there are many more options than the one I described.
### Use Cases / Examples
[...]

PR #383, opened 2014-01-08
Added API for broadcast variables.

aalexandrov:
This is a first step towards realizing the functionality requested in issue #66. 

The broadcast inputs API consists of two aspects:
- Setters: ~~setParameterInput()~~ `setBroadcastVariable()` / `getBroadcastVariable()` functions in `AbstractUdfOperator` and all Operator Builders. Used to broadcast (bind) subplan results to UDF-local variables.
- Getters: ~~getParameterInput()~~ `setBroadcastVariable()` / `getBroadcastVariable()` in `RuntimeContext`. Used within an UDF open() method for parametrization of the UDF with values provided by the plans rooted at the configured parameter inputs.

A sample intended usage of the API can be seen in the modified `KMeansIterativeWithBroadcastVariables` Java example.

### DIFF
--- a/stratosphere-java/src/main/java/eu/stratosphere/api/java/record/operators/CoGroupOperator.java
+++ b/stratosphere-java/src/main/java/eu/stratosphere/api/java/record/operators/CoGroupOperator.java
@@ -18,7 +18,9 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 import eu.stratosphere.api.common.operators.Operator;
 import eu.stratosphere.api.common.operators.Ordering;
@@ -27,7 +29,6 @@ import eu.stratosphere.api.common.operators.util.UserCodeClassWrapper;
 import eu.stratosphere.api.common.operators.util.UserCodeObjectWrapper;
 import eu.stratosphere.api.common.operators.util.UserCodeWrapper;
 import eu.stratosphere.api.java.record.functions.CoGroupFunction;
-import eu.stratosphere.api.java.record.functions.JoinFunction;
 import eu.stratosphere.types.Key;
 
 /**
@@ -92,6 +93,7 @@ public class CoGroupOperator extends CoGroupOperatorBase<CoGroupFunction> implem
 		this.keyTypes = builder.getKeyClassesArray();
 		setFirstInputs(builder.inputs1);
 		setSecondInputs(builder.inputs2);
+		setBroadcastVariables(builder.broadcastInputs);
 		setGroupOrderForInputOne(builder.secondaryOrder1);
 		setGroupOrderForInputTwo(builder.secondaryOrder2);
 	}
@@ -209,6 +211,7 @@ public class CoGroupOperator extends CoGroupOperatorBase<CoGroupFunction> implem
 		/* The optional parameters */
 		private List<Operator> inputs1;
 		private List<Operator> inputs2;
+		private Map<String, Operator> broadcastInputs;
 		private Ordering secondaryOrder1 = null;
 		private Ordering secondaryOrder2 = null;
 		private String name = DEFAULT_NAME;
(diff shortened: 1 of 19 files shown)


rmetzger:
Do you have any ideas for alternative Names (for `setParameterInput()`)
What do you say to `setBroadcastVariable()` ? and `getBroadcastVariables()` ?

uce:
Thanks for sharing early. I like @rmetzger's proposal. I think _broadcast_ should definitely be in the method names.

I would even propose a simpler variation like (`broadcast(String, Value)` and (`getBroadcastValue(String)` xor just `broadcastValue(String)`)).

aalexandrov:
Ok I renamed the methods as follows
- `broadcast(String name, Operator root)` and 
- `Collection<Record> getBroadcastVariable(String name)`.

stratosphere-java/src/main/java/eu/stratosphere/api/java/record/operators/CoGroupOperator.java
```diff
@@ -357,6 +361,24 @@ public Builder inputs2(List<Operator> inputs) {
 		}
 		
 		/**
+		 * Binds the result produced by a plan rooted at {@code root} to a 
+		 * variable used by the UDF wrapped in this operator.
+		 */
+		public Builder broadcast(String name, Operator input) {
```
StephanEwen: "broadcast(...) or "broadcastVariable(...)" ?

fhueske:
I think naming the function in `AbstractUDFOperator` `broadcast()` could be misleading.
I would expect that the operator on which `broadcast()`is called, actually broadcasts its result. Instead it would read the broadcasted variable...

How about `addBroadcastVariable()` or `addBroadcastInput()`?

fhueske:
Isn't the K-Means example a bit too complex to start with?
In this case, the broadcast variable becomes the data that is fed back into the iteration. I am not very familiar with the iterations compiler and runtime code, but I would guess that this will cause some non-trivial changes there.

Wouldn't a simpler example job without iterations be easier to handle for a first draft?

aalexandrov:
I was just using for an example job that lends itself for rewriting. I think K-Means should work with a naive implementation because the broadcasted intermediate result and the consumer are within the same iterative path.

A more general case which will cause problems will be a program with the following structure:

``` scala
{ 
  var A = ... // broadcast variable

  iterate {
    var B = ... // broadcast variable

    iterate {
      var C = // broadcast variable
      // depends on A, B, and C
      var D = map( t => { ... } )
    } // inner iteration
  } // outer iteration
} // program
```

Here, the parameterized mapper function depends on the broadcast variables A, B, and C. And while C is re-computed and has to be re-drained from the RegularPactTask in each iteration, B and A have to be drained only once per outer iteration and program execution respectively (in K-Means the "centers" variable is computed within the same iteration just like C). 

In oder to implement this logic we need to annotate tasks with some sort of scope information and keep track of changes in the scope paths on task re-deployment:

``` scala
{ 
  var A = ... // broadcast variable

  iterate {
    var B = ... // broadcast variable

    iterate {
      var C = // broadcast variable
      // depends on:
      // * A (drained on change in s1),
      // * B (drained on change in s2)
      // * C (drained on change in s3)
      var D = map( t => { ... } )
    } // inner iteration (scope s1.s2.s3)
  } // outer iteration (scope s1.s2)
} // program (scope s1)
```

I'm not quite sure whether and how the cost estimation logic for plan enumeration and pruning is affected by this addition. @StephanEwen: your thoughts on this?

fhueske:
I thought of an example without iteration.
Something like:

```
X -> Reduce(avg) --broadcast-----\
                             Y --> Map(y - avg(X)) --> out
```

which can currently be emulated in Stratosphere with a Cross (instead of broadcast Map) or the distributed cache in Hadoop.

sscdotopen:
I really like to comments from your example: boradcast variable, I think BoradCast is a much cooler name than broadcast

StephanEwen:
I agree with Fabian. It is a good idea to use K-Means to model the API and
validate its ease of use. However, for a first test of the runtime code, a
non-iterative program would be easier.

I will try and comment on the code soon.

sscdotopen:
You could try to rewrite this program: https://github.com/stratosphere/gagarin/blob/master/src/main/scala/de/tuberlin/dima/gagarin/recommendation/CooccurrenceAnalysis.scala

If numInteractionsPerItem and numInteractionsPerUser are made broadcast variables, the code should be much easier to read.

fhueske:
TPC-H Query 11 gives another good example, how a broadcast variable could replace a cross: 
[TPC-H 11](https://github.com/stratosphere/stratosphere-tpch/blob/master/src/main/scala/eu/stratosphere/tpch/query/TPCHQuery11.scala)

aalexandrov:
Ooops, just noted @sscdotopen comment. Some BoradCasted variables should be flying around in the integration test for PR #460.

PR #387, opened 2014-01-09
Set LocalExecutor logging to INFO; print runtime. reduce logging verbosity of IPC server

rmetzger:


uce:
I think this is good to merge.

uce:
Merged in 3b682a98e949a052a907e501be067190d4f0a1dd.

Issue #376, opened 2014-01-07
Create parallel collections for stratosphere


rmetzger:
The goal is that you can simply make a Java or Scala List of elements the input to a Stratosphere program. Initially, we can assume small lists that can be sent as part of the program parameters. 
(Copied from https://github.com/stratosphere/stratosphere/wiki/Starter-Jobs)

API Examples

``` scala
source = DataSource(Seq("some", "strings", "that", "get", distributed"));
source = DataSource(1 to 10000);
```

``` java
source = new CollectionSource("some", "strings", "that", "get", "distributed");
source = new CollectionSource(new Object[][] { { 1, "a"}, { 2, "b"}, {3, "c"} } );
```

**Update**:
The next step after this to also allow iterators as a CollectionDataSource.

``` java
CollectionDataSource cds = new CollectionDataSource(new Iterator<Object>() {
    private Random rand = new Random(1337);
    private int cnt = 0;
    @Override
    public void remove() {}
    @Override
    public Object next() {
        return rand.nextLong();
    }
    @Override
    public boolean hasNext() {
        return cnt++ < 42;
    }
});
```

One of the new students (@qmlmoon) is assigned to this task.
I created this issue to track the progress (and give everybody a chance to comment on this)

PR #388, opened 2014-01-10
create parallel collections and add java value to stratosphere value interface

qmlmoon:
This is the implementation for issue https://github.com/stratosphere/stratosphere/issues/376.


stratosphere-core/src/main/java/eu/stratosphere/api/common/io/CollectionInputFormat.java
```diff
@@ -0,0 +1,74 @@
...
+		if (readObject()) {
+			return readRecord(record, this.currObject);
+		} else {
+			this.end = true;
+			return false;
+		}
+	}
+	
+	public void setData(List<Object> data) {
+		this.steam = new ArrayList<Object>(data);
```
rmetzger: Why are you creating a copy of the List here?

stratosphere-core/src/main/java/eu/stratosphere/api/common/operators/CollectionDataSource.java
```diff
@@ -0,0 +1,104 @@
...
+
+/*
+ * Operator for input nodes which read data from collection.
+ * 
+ */
+public class CollectionDataSource extends GenericDataSource<GenericInputFormat<?>> {
+
+	private static String DEFAULT_NAME = "<Unnamed Collection Data Source>";
+	
+	protected final List<Object> data;
```
rmetzger: Can you use a `Collection<Object>` here? (then also in the constructor and the CollectionInputFormat)
I don't see a reason not to use a collection (you would hereby allow to use sets and stuff as well)

rmetzger:
@qmlmoon: I updated the issue description (https://github.com/stratosphere/stratosphere/issues/376) with an iterator example.

qmlmoon:
Yeah. It should be a collection. I was not familiar with this collection data format. I’ll change it.

stratosphere-core/src/main/java/eu/stratosphere/api/common/io/CollectionInputFormat.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.Iterator;
+
+import eu.stratosphere.api.common.operators.util.SerializableIterator;
+import eu.stratosphere.core.io.GenericInputSplit;
+import eu.stratosphere.types.Record;
+import eu.stratosphere.types.ValueUtil;
+
+/**
+ * input format for java collection input. It can accept collection data or serializable iterator
+ * @author qml_moon
```
rmetzger: Hey, we don't add `@author` tags to the Stratosphere code because we want to have something like "shared code ownership".

stratosphere-core/src/main/java/eu/stratosphere/api/common/operators/CollectionDataSource.java
```diff
@@ -0,0 +1,149 @@
...
+	 * Creates a new instance for the given input using the given input format. The contract has the default name.
+	 * The input types will be checked. If the input types don't agree, an exception will occur.
+	 * 
+	 * @param f The {@link CollectionInputFormat} implementation used to read the data.
+	 * @param data The input data. It should be a collection, an array or a serializable iterator of object.
+	 */
+	@SuppressWarnings("unchecked")
+	public CollectionDataSource(CollectionInputFormat f, Object[] data) {
+		super(f, DEFAULT_NAME);
+		Collection<Object> data_tmp = new ArrayList<Object>();
```
rmetzger: `data_tmp` does not conform to the standard Java way to name variables. The common way is to use "CamelCase" for variables (however with the first letter being lower-case). To in your case, the variable should be named `data_tmp`.
Sorry for this but we want the code to look consistent.

stratosphere-core/src/main/java/eu/stratosphere/types/ValueUtil.java
```diff
@@ -0,0 +1,33 @@
...
+		if (java instanceof java.lang.Float)
+			return new FloatValue(((java.lang.Float)java).floatValue());
+		if (java instanceof java.lang.Long)
+			return new LongValue(((java.lang.Long)java).longValue());
+		if (java instanceof java.lang.Short)
+			return new ShortValue(((java.lang.Short)java).shortValue());
+		if (java instanceof java.lang.String)
+			return new StringValue(((java.lang.String)java).toString());
+		
+		return NullValue.getInstance();
```
rmetzger: could you throw an exception instead of returning the `NullValue` ? (except for `java == null`)
The problem with returning the `NullValue` by default is that it will be really hard for developers to debug errors in their jobs if they run into this condition.

stratosphere-scala/src/main/scala/eu/stratosphere/api/scala/DataSource.scala
```diff
@@ -60,6 +63,37 @@ object DataSource {
   }
 }
 
+object CollectionDataSource {
```
rmetzger, line 65: You should not follow @aljoscha's example not to comment the Scala code ;)
It would be great if you could quickly describe what the code does.

stratosphere-tests/src/test/java/eu/stratosphere/test/exampleRecordPrograms/WordCountCollectionCase.java
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * test the collection and iterator data input using Wordcount example
+ * @author qml_moon
+ *
+ */
+public class WordCountCollectionCase extends WordCountITCase {
+	
+	@SuppressWarnings("hiding")
+	public static class SerializableIteratorTest<Object> extends SerializableIterator<Object> {
```
rmetzger: Why did you set the generic type to `Object` here? Can't you set it so `String` (because that's what you return with `next()`.

stratosphere-tests/src/test/java/eu/stratosphere/test/exampleRecordPrograms/WordCountCollectionCase.java
```diff
@@ -0,0 +1,97 @@
...
+		
+		public Plan getPlan(String arg1, String arg2) {
+			// parse job parameters
+			int numSubTasks   = Integer.parseInt(arg1);
+			String output    = arg2;
+			
+			/*
+			 * uncomment this to test List input.
+			 */
+//			List<Object> tmp= new ArrayList<Object>();
```
rmetzger: Your test case should test both variants (our tests are being executed automatically) .. so all variants should be included in the code.
(You should be able to use multiple `CollectionDataSource`s in one job).
You also have to verify the correctness of the result somewhere (because otherwise testing would not make any sense)

stratosphere-core/src/main/java/eu/stratosphere/api/common/operators/CollectionDataSource.java
```diff
@@ -0,0 +1,149 @@
...
+	}
+	
+	
+
+	
+	// --------------------------------------------------------------------------------------------
+	/*
+	 * check whether the input field has the same type
+	 */
+	private void checkFormat(Collection<Object> c) {
```
rmetzger: Could you write a Unit-Test that checks and validates the correctness of this method (by trying all valid variants and some invalid variants) ?

I'm not 100% sure but are you certain that you have to use Strings to store the type of the elements?
I guess you could probably also do it by making the `type` variable a `Class` reference.

stratosphere-core/src/main/java/eu/stratosphere/api/common/operators/CollectionDataSource.java
```diff
@@ -0,0 +1,149 @@
...
+	}
+	
+	
+	/**
+	 * Creates a new instance for the given input using the given input format. The contract has the default name.
+	 * The input types will be checked. If the input types don't agree, an exception will occur.
+	 * 
+	 * @param data The input data. It should be a collection, an array or a serializable iterator of object.
+	 */
+	public CollectionDataSource(Object... args) {
```
rmetzger: Can you describe more explicitly (with stating the name of the `SerializableIterator`)  that user can also pass the iterator here?

rmetzger:
Hey @qmlmoon,

you're code is really good! I tested it today and everything worked as expected.
Once you resolved the issues I commented on, it is good to merge.

qmlmoon:
Thank you:D. I’ll work on your comments at Monday.

.gitignore
```diff
@@ -1,3 +1,4 @@
+.DS_Store
```
StephanEwen: Please remove this from the pull request to not conflict with the cleaned up gitignore from @aalexandrov

stratosphere-core/src/main/java/eu/stratosphere/types/ValueUtil.java
```diff
@@ -0,0 +1,33 @@
+package eu.stratosphere.types;
+
+/**
+ * convert the java.lang type into stratosphere type
+ */
+public class ValueUtil {
+
+	public static Value toStratosphere(Object java)  {
+		
+		if (java instanceof java.lang.Boolean)
```
StephanEwen: I think the code (java.getClass() == java.lang.Boolean.class) is a bit faster, because it really only checks for a comparison. The instanceof operator traverses the hierarchy metadata in the class object.

Since all these classes are final, this check should be absolutely equivalent.

stratosphere-core/src/main/java/eu/stratosphere/api/common/io/CollectionInputFormat.java
```diff
@@ -0,0 +1,103 @@
...
+
+	private transient boolean end;
+	
+	private Collection<Object> steam;		//input data as collection
+	
+	private SerializableIterator<Object> serializableIter;	//input data as serializable iterator
+	
+	private transient Iterator<Object> it;
+	
+	private transient Object currObject;
```
StephanEwen: Can you rework the logic to work without a currObject?

In `reachedEnd()` you can simply check whether the iterator has no next element ( `boolean readedEnd() { return !it.hasNext(); }`

And in `boolean nextRecord(Record record)`you can directly take the next object from the iterator.

stratosphere-core/src/main/java/eu/stratosphere/api/common/io/CollectionInputFormat.java
```diff
@@ -0,0 +1,103 @@
...
+		}
+		else {
+			target.setField(0, ValueUtil.toStratosphere(b));
+		}
+		return true;	
+	}
+	
+	@Override
+	public void open(GenericInputSplit split) throws IOException {
+		this.partitionNumber = split.getSplitNumber();
```
StephanEwen: This seems like reproducing the super class behavior. It is cleaner to call `super.open(split)`.

stratosphere-core/src/main/java/eu/stratosphere/api/common/io/CollectionInputFormat.java
```diff
@@ -0,0 +1,103 @@
...
+ * 
+ */
+public class CollectionInputFormat extends GenericInputFormat<Record> implements UnsplittableInput {
+
+
+	private static final long serialVersionUID = 1L;
+
+	private transient boolean end;
+	
+	private Collection<Object> steam;		//input data as collection
```
StephanEwen: `steam` seems like an odd name. Is it a type for `stream`? I would use a name like `dataSet`, which is probably more intuitive to someone who reads the code.

stratosphere-core/src/main/java/eu/stratosphere/api/common/io/CollectionInputFormat.java
```diff
@@ -0,0 +1,103 @@
...
+			return false;
+		}
+	}
+	
+	public void setData(Collection<Object> data) {
+		this.steam = data;
+		this.serializableIter = null;
+	}
+	
+	public void setIter(SerializableIterator<Object> iter) {
```
StephanEwen: The SerializableIterator is a clumsy concept in my opinion. If someone has an iterator that is also serializable, he would still have to wrap it in an extra iterator.

Can you change this to something like

```
public <T extends Iterator<?>, Serializable> void setIterator(T iter) {
  this.iter = iter;
}
```

That way, all classes that are already serializable iterators work out of the box.

stratosphere-core/src/main/java/eu/stratosphere/api/common/operators/util/SerializableIterator.java
```diff
@@ -0,0 +1,20 @@
+package eu.stratosphere.api.common.operators.util;
```
StephanEwen: I would remove this class. See the comment above about iterators that are already serializable.

stratosphere-core/src/main/java/eu/stratosphere/api/common/operators/CollectionDataSource.java
```diff
@@ -0,0 +1,151 @@
...
+	// --------------------------------------------------------------------------------------------
+	
+	/**
+	 * Creates a new instance for the given input using the given input format.
+	 * 
+	 * @param f The {@link CollectionInputFormat} implementation used to read the data.
+	 * @param data The input data. It should be a collection, an array or a {@link SerializableIterator}.
+	 * @param name The given name for the Pact, used in plans, logs and progress messages.
+	 */
+	public CollectionDataSource(CollectionInputFormat f, Collection<Object> data, String name) {
```
StephanEwen: The choice of constructors is a bit wild. What system do they follow?

I would suggest to have the following set of constructors:
- `CollectionDataSource(Object ... elements)`  // this also matches the Object[]
- `CollectionDataSource(String name, Object ... elements)`  // this also matches the Object[]
- `CollectionDataSource(Object[][] ... objs)`
- `CollectionDataSource(String name, Object[][] ... objs)`
- `CollectionDataSource(Collection<?> objs)`
- `CollectionDataSource(String name, Collection<?> ... objs)`
- `<T extends Iterator<?>, Serializable> CollectionDataSource(T iterator)`
- `<T extends Iterator<?>, Serializable> CollectionDataSource(String name, T iterator)`

stratosphere-tests/src/test/java/eu/stratosphere/test/exampleRecordPrograms/WordCountCollectionCase.java
```diff
@@ -0,0 +1,213 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+
+/**
+ * test the collection and iterator data input using Wordcount example
+ *
+ */
+public class WordCountCollectionCase extends WordCountITCase {
```
StephanEwen: I think the name WordCount is a bit confusing here. You use the word count result data, but the test is actually a join between two ways of representing that. Can you think of a better name? Even `CollectionSourceTest` would be fine.

stratosphere-tests/src/test/java/eu/stratosphere/test/exampleRecordPrograms/WordCountCollectionCase.java
```diff
@@ -0,0 +1,213 @@
+package eu.stratosphere.test.exampleRecordPrograms;
```
StephanEwen: Can you split this into two test?

There are two things happening here:
- The unit test style testing of the validation methods
- The end-to-end (ITCase) style testing of an entire program that uses the collection source.

StephanEwen:
Mingliang, I wrote some comments to different parts of the code. Please have a look at them and let me know if you have questions, or if you do not agree with the comment.

stratosphere-core/src/main/java/eu/stratosphere/api/common/io/CollectionInputFormat.java
```diff
@@ -0,0 +1,79 @@
...
+            if (b.getClass().isArray()) {
+                for (Object s : (Object[])b){
+                    record.addField(ValueUtil.toStratosphere(s));
+                }
+            }
+            else if (b instanceof Collection) {
+                @SuppressWarnings("unchecked")
+                Iterator<Object> tmp_it = ((Collection<Object>) b).iterator();
+                while (tmp_it.hasNext())
+                {
```
rmetzger: Please use the same style for all loops (opening parentheses in the same line)
while (..) {
...
}

stratosphere-core/src/main/java/eu/stratosphere/api/common/io/CollectionInputFormat.java
```diff
@@ -0,0 +1,79 @@
...
+	@Override
+	public boolean reachedEnd() throws IOException {
+		return !it.hasNext();
+	}
+
+
+	@Override
+	public void open(GenericInputSplit split) throws IOException {
+        super.open(split);
+		if (serializableIter != null)
```
rmetzger: we always use parentheses, also for one-line statements.

stratosphere-core/src/main/java/eu/stratosphere/api/common/io/CollectionInputFormat.java
```diff
@@ -0,0 +1,79 @@
...
+            Object b = it.next();
+            //check whether the record field is one-dimensional or multi-dimensional
+            if (b.getClass().isArray()) {
+                for (Object s : (Object[])b){
+                    record.addField(ValueUtil.toStratosphere(s));
+                }
+            }
+            else if (b instanceof Collection) {
+                @SuppressWarnings("unchecked")
+                Iterator<Object> tmp_it = ((Collection<Object>) b).iterator();
```
rmetzger: `tmp_it` does not conform our (and the java) naming guidelines (because it contains an underscore). I would suggest `tmpIter` or so here. (camelCase)

stratosphere-core/src/main/java/eu/stratosphere/api/common/operators/CollectionDataSource.java
```diff
@@ -0,0 +1,183 @@
...
+                    if (!s.getClass().equals(typeList.get(index++))) {
+                        throw new RuntimeException("elements of input list should have the same type");
+                    }
+            }
+
+            //check the input types for 2-dimension collection
+            if (typeList.size() == 0 && o instanceof Collection) {
+                @SuppressWarnings("unchecked")
+                Iterator<Object> tmpIt = ((Collection<Object>) o).iterator();
+                while (tmpIt.hasNext())
```
rmetzger: style

stratosphere-tests/src/test/java/eu/stratosphere/test/exampleRecordPrograms/CollectionValidationTest.java
```diff
@@ -0,0 +1,113 @@
...
+
+    @Test
+    public void TestCollectionInputValidation() throws Exception {
+        /*
+        valid collection input
+         */
+        try {
+            List<Object> tmp= new ArrayList<Object>();
+            for (int i = 0; i < 100; i++)
+                tmp.add(i);
```
rmetzger, line 62: style. Please check your own code yourself for the correct style. Thank you!

rmetzger:
Merged in 8dd2470162167a6bb051ca3383aa0c364a680985

qmlmoon:
The newest update of tuple up to 25 seems not update the MAX_ARITY in abstract class Tuple, which leads to the IllegalArgumentException in TupleTypeInfo

PR #398, opened 2014-01-13
Meteor/Sopremo documentation added

AHeise:


rmetzger:
Hi
thanks for the pull request. looks good!

One minor comment (but I think it is okay to to use the text even with that) is that you're mentioning the "Nephele JobManager". We don't describe nephele and pact in the documentation anymore, so these are unknown concepts to the average user.

AHeise:
Fixed in    e02260a

Issue #404, opened 2014-01-16
Failing CustomDataTypeTest


uce:
Can anybody confirm that this test is not working:

`mvn clean compile test -DfailIfNoTests=false -Dtest=CustomDataTypeTest`

The weird thing is that the test doesn't fail when the all tests are run, e.g. `mvn clean verify`.

I have an idea why this might happen. It would be nice though if someone could make sure that it is nothing local to my setup. After that I will try the fix. ;-)

aljoscha:
Yes, I talked to @rmetzger about this today. It fails when I run it as a single test but not with mvn verify and so on.

rmetzger:
`mvn clean compile test -DfailIfNoTests=false -Dtest=CustomDataTypeTest` gives me: 

```
Failed tests: 
  testJob[0](eu.stratosphere.test.operators.io.CustomDataTypeTest)
```

So I can confirm this.
`CustomDataTypeTest` also fails from Eclipse.

`mvn clean verify` does not fail. BUT it also shows the exception: https://s3.amazonaws.com/archive.travis-ci.org/jobs/17051218/log.txt

The maven output looks like this

```
    at com.sun.proxy.$Proxy0.invoke(Unknown Source)
    at org.apache.maven.surefire.booter.SurefireStarter.invokeProvider(SurefireStarter.java:139)
    at org.apache.maven.surefire.booter.SurefireStarter.runSuitesInProcess(SurefireStarter.java:82)
    at org.apache.maven.surefire.booter.ForkedBooter.main(ForkedBooter.java:81)
Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 4.299 sec <<< FAILURE!
```

We have to find out why `maven-surefire-plugin` is accepting failed tests.
So I would argue this particular issue is caused by the "build system"

uce:
It's not a problem with the build system and the test fails with my upcoming changes. Anyways, the test is flawed. I vote to remove it.

PS: It throws a `ClassNotFoundException`, because a jar has been deleted by accident before. @rmetzger is on it. ;-)

rmetzger:
The jar file has been deleted by commit 33cb2ca9898809d2fc90765996ea56bbea458e59

StephanEwen:
We should have some test that tests the class loading, though. Will this be
checked now by the new big regression job that Robert is creating?

uce:
Yes, @aljoscha's test case will involve class loading.

[...]

PR #407, opened 2014-01-17
Add shutdown method to LocalDistributedExecutor

uce:
This PR adds a shutdown method to the `LocalDistributedExecutor`.

While reworking the corresponding test case I noticed a problem with `CustomDataTypeTest` (see #404). As @rmetzger noted, a JAR file has been accidently deleted in 33cb2ca9898809d2fc90765996ea56bbea458e59. Because @aljoscha is going to implement a similar test case, we voted to remove `CustomDataTypeTest`.


uce:
BTW: I will add a switch to `TestBase2` to allow using `LocalDistributedExecutor` for test cases (will be off by default).

aljoscha:
I like :thumbsup:

rmetzger:
Merged in bdd378c8b754d05a32ad260ada6843c104ffbf2a. Thanks.

PR #408, opened 2014-01-17
Reenable ClosureCleaner and add new Test case for everything

aljoscha:
Closure cleaner did not work before because the user code class loader
was not used when it was trying to load inner anonymous classes.

The new test checks whether the ClosureCleaner works together with the
user code class loader.

It also used the local distributed executor and remote executor.


rmetzger:
Do we really want a test called `UserCodeClassLoaderRemoteExecutorLocalDistributedExecutorClosureCleanerTest` in our code ?

stratosphere-tests/src/test/java/eu/stratosphere/test/localDistributed/UserCodeClassLoaderRemoteExecutorLocalDistributedExecutorClosureCleanerTest.java
```diff
@@ -0,0 +1,85 @@
...
+import org.junit.Test;
+
+import eu.stratosphere.client.localDistributed.LocalDistributedExecutor;
+import eu.stratosphere.example.java.record.wordcount.WordCount;
+import eu.stratosphere.test.exampleRecordPrograms.WordCountITCase;
+
+// When the API changes WordCountForTest needs to be rebuilt and the WordCountForTest.jar in resources needs
+// to be replaced with the new one.
+
+public class UserCodeClassLoaderRemoteExecutorLocalDistributedExecutorClosureCleanerTest {
```
StephanEwen: Can we call this something like "PackagedProgramEndToEndTest" ?

StephanEwen:
Looks good to me, besides the name ;-)

rmetzger:
Travis indicates that there is something wrong

Sent from my iPhone

On 19.01.2014, at 21:09, Stephan Ewen notifications@github.com wrote:

Looks good to me, besides the name ;-)

—
GitHubhttps://github.com/stratosphere/stratosphere/pull/408#issuecomment-32718391
.

uce:
We also have to make sure that it is not a problem with the `LocalDistributedExecutor`. Actually, I've waited to integrate `TestBase2` with the LDE in order to make sure that everything is fine with this test case.

@aljoscha Do you have any ideas why the test case fails? Is it because of LDE?

aljoscha:
Stupid me, I forgot to add the jar file to the repo...
I also changed to name per @StephanEwen's suggestion.

PR #409, opened 2014-01-17
Fix bug in scala cogroup.flatMap

aljoscha:
Did fail with null pointer exception when one of the inputs is empty.


rmetzger:
Thanks for the fix. Commited in 418fb21ff886cd4be782c2499e7548c0303baad6

Issue #411, opened 2014-01-18
Change title of website start page


fhueske:
Google shows the title of the start page in is search results.
For stratosphere.eu that is currently: 
"Stratosphere >> Overview"
which is not telling anything.

I propose to change the page title to:
"Stratosphere - Big Data looks tiny from here"
or
"Stratosphere - Next Generation Big Data Analytics"

Google's more detailed explaining text is also not well chosen and should be updated. Currently it is:
"Stratosphere extends the well-known MapReduce model with new operators. These operators represent many common data analysis tasks more naturally and ..."

uce:
I agree and I've pushed the following changes:

Title from _Overview_ to _Next Generation Big Data Analytics Platform_.

The _description_ metatag is now the same as the first paragraph of the startpage:

> Stratosphere is the next generation Big Data Analytics Platform. It combines the strengths of MapReduce/Hadoop with powerful programming abstractions in Java and Scala and a high performance runtime. Stratosphere has native support for iterations, incremental iterations, and programs consisting of large DAGs of operations."

fhueske:
Well, actually Google ignores the description meta tag if it is identical on most pages of the website.
That's why I extracted it. Will rebase and update my PR.

uce:
OK, thanks! :-)

PR #412, opened 2014-01-19
Some search engine optimization for the website.

fhueske:
Updated title of start page from "Stratosphere > Overview" to "Stratosphere > Next-Generation Big Data Analytics".

Extracted static meta description and keywords from layout. 
Set individual descriptions and keywords for the most important pages.

Search engines (at least Google) penalize identical description and keyword meta data on multiple pages.

This should give us a better search result entries (title + description) and hopefully also better keyword coverage.

This PR resolves issue #411


uce:
See #411.

PR #415, opened 2014-01-20
Added reflective Avro type

StephanEwen:


StephanEwen:
Yes, I agree, SAvroValue is not too nice. How about
- AvroValueBase
- AvroValueHolder

uce:
AvroBaseValue to keep the Value suffix?

Issue #414, opened 2014-01-20
Adjust CsvOutputFormat to take data types through constructor


StephanEwen:
Refactor the CsvOutputFormat the same way as the CsvInputFormat to accept the types through the constructor or a setter. The config builder should remain for compatibility and detailed confuguration (sparse records)

PR #417, opened 2014-01-20
Changed CsvOutputFormat constructor to support additional arguments

skunert:
I changed the CsvOutputFormat constructor to support non-generic instantiation.

### DIFF
--- a/stratosphere-java/src/main/java/eu/stratosphere/api/java/record/io/CsvOutputFormat.java
+++ b/stratosphere-java/src/main/java/eu/stratosphere/api/java/record/io/CsvOutputFormat.java
@@ -160,8 +160,16 @@ public class CsvOutputFormat extends FileOutputFormat {
 		this.recordPositions = new int[types.length];
 		for (int i = 0; i < types.length; i++) {
 			if (types[i] == null)
-				throw new IllegalArgumentException("Invalid Constructor Parameter: No type class for parameter " + (2 + i));
+				throw new IllegalArgumentException("Invalid Constructor Parameter: No type class for parameter " + (2 * i));
 			this.recordPositions[i] = i;
+		}
+		
+		if (this.fieldDelimiter == null) {
+			this.fieldDelimiter = ",";
+		}
+		
+		if (this.recordDelimiter == null) {
+			this.recordDelimiter = "\n";
 		}
 	}
 	
--- a/stratosphere-java/src/main/java/eu/stratosphere/api/java/record/io/CsvOutputFormat.java
+++ b/stratosphere-java/src/main/java/eu/stratosphere/api/java/record/io/CsvOutputFormat.java
@@ -85,14 +85,11 @@ public class CsvOutputFormat extends FileOutputFormat {
 
 	private boolean lenient;
 
-	private boolean genericInstantiation;
-
 	// --------------------------------------------------------------------------------------------
 	// Constructors and getters/setters for the configurable parameters
 	// --------------------------------------------------------------------------------------------
 
 	public CsvOutputFormat() {
-		genericInstantiation = true;
 	}
 
 	/**
@@ -155,8 +152,6 @@ public class CsvOutputFormat extends FileOutputFormat {
 		this.lenient = false;
 
 		setTypes(types);
-		
-		genericInstantiation = false;
 	}
 	
 	public void setTypes(Class<? extends Value>... types) {
@@ -168,7 +163,6 @@ public class CsvOutputFormat extends FileOutputFormat {
 				throw new IllegalArgumentException("Invalid Constructor Parameter: No type class for parameter " + (2 + i));
 			this.recordPositions[i] = i;
 		}
-		genericInstantiation = false;
 	}
 	
 	public void setLenient(boolean lenient) {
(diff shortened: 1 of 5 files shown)


rmetzger:
This pull-request implements https://github.com/stratosphere/stratosphere/issues/414

stratosphere-java/src/main/java/eu/stratosphere/api/java/record/io/CsvOutputFormat.java
```diff
@@ -85,14 +85,78 @@
...
+	}
+	
+	/**
+	 * Creates an instance of CsvOutputFormat. The position of the fields in the record is determined by the order in which the classes are given to this constructor.
+	 * As the default value for separating records '\n' is used. The default field delimiter is '|'.
+	 * @param fieldDelimiter The delimiter that is used to separate the different fields in the record.
+	 * @param types The types of the fields that are in the record.
+	 */
+	public CsvOutputFormat(Class<? extends Value> ... types) {
+		this("\n", "|", types);
```
rmetzger: Can you use "," as the default separator? 
The thing is called "CSV" ... ;)

stratosphere-java/src/main/java/eu/stratosphere/api/java/record/io/CsvOutputFormat.java
```diff
@@ -85,14 +85,78 @@
...
 	
+	/**
+	 * Creates an instance of CsvOutputFormat. The position of the fields in the record is determined by the order in which the classes are given to this constructor.
+	 * 
+	 * @param recordDelimiter The delimiter that is used to separate the different records.
+	 * @param fieldDelimiter The delimiter that is used to separate the different fields in the record.
+	 * @param types The types of the fields that are in the record.
+	 */
+	public CsvOutputFormat(String recordDelimiter, String fieldDelimiter, Class<? extends Value> ... types) {
+		if (recordDelimiter == null)
```
rmetzger: we use always braces { .. } for if etc.

stratosphere-java/src/main/java/eu/stratosphere/api/java/record/io/CsvOutputFormat.java
```diff
@@ -85,14 +85,78 @@
...
+		}	
+		genericInstantiation = false;
+	}
 
 	@Override
 	public void configure(Configuration parameters)
 	{
 		super.configure(parameters);
 		
+		if (!genericInstantiation)
```
rmetzger: Lets see if somebody else has an objection here.
It should all work with this check here. 
This method is also called on all the machines where the CsvFile is written to configure it. (In that case, the framework creates and instance using the default constructor and calls configure()).

StephanEwen:
Looks good in most parts, I'll give some comments inline.

Also, please reformat the code for a wider line-width ( >= 120 ). We are in the era of larger screens and higher resolutions ;-) Most classes assume 120 chard line width.

stratosphere-java/src/main/java/eu/stratosphere/api/java/record/io/CsvOutputFormat.java
```diff
@@ -49,231 +49,343 @@
  */
 public class CsvOutputFormat extends FileOutputFormat {
 	private static final long serialVersionUID = 1L;
-	
+
```
StephanEwen, line 52: Can you please remove the `pact` prefix from all the config key strings.

stratosphere-java/src/main/java/eu/stratosphere/api/java/record/io/CsvOutputFormat.java
```diff
@@ -49,231 +49,343 @@
...
+		if (types.length == 0) {
+			throw new IllegalArgumentException("No field types given.");
+		}
+
+		this.fieldDelimiter = fieldDelimiter;
+		this.recordDelimiter = recordDelimiter;
+		this.numFields = types.length;
+
+		@SuppressWarnings("unchecked")
+		Class<Value>[] classes = new Class[types.length];
```
StephanEwen: Can't you just assign `this.classes = types`?

stratosphere-java/src/main/java/eu/stratosphere/api/java/record/io/CsvOutputFormat.java
```diff
@@ -49,231 +49,343 @@
...
-	public void configure(Configuration parameters)
-	{
+	public void configure(Configuration parameters) {
 		super.configure(parameters);
-		
+
+		if (!genericInstantiation)
+			return;
+
 		this.numFields = parameters.getInteger(NUM_FIELDS_PARAMETER, -1);
```
StephanEwen: How about checking here for the following conditions:
- `numFields == -1 && this.classes != null`  --> Types via parameters. Good!
- `numFields > 0 && this.classes == null` --> Types via config object. Good!
- `numFields == -1 && this.classes == null` --> Neither of both, error!
- `numFields >0 && this.classes != null` --> Both at the same time, error!

This would also not silently ignore the config, is the parameters are set.

StephanEwen:
Please avoid reformatting source. It makes the diffs impossible to read. Right now, I cannot really figure out where you really changed something and where the formatter changed something.

StephanEwen:
It would also be nice to add a setter for the types, as another way to set the types in addition to the constructor.

stratosphere-java/src/main/java/eu/stratosphere/api/java/record/io/CsvOutputFormat.java
skunert: Maybe I am missing something here, when you instantiate a CsvOutputFormat via the new constructor, the numFields is set as well as the classes. Thus, I can only differentiate between generic instantiation and parameters. Could you explain what you mean by silently ignoring the config?

skunert:
I think the issues you mentioned are fixed.

rmetzger:
I merged together the 7 commits into one and removed some empty lines.
Merged in 8344bbcce3bb1b524e75e22def90a4be30083a25

rmetzger:
I merged this pull request. But I did not see that Travis did not test the most recent commits.
Sadly, the CSVOutputFormat tests are failing!

The problem is that the OF does not support multiple `configure()` calls.
But I don't really get why we want to throw an exception if users call configure after a ctor-based initialization.

rmetzger:
Fixed in db40aa164112b2c4de5867dcb4adb8baca9561a8 @skunert and @StephanEwen please review.

rmetzger:
It seems that this does not work in a cluster

```
ERROR: The job was not successfully submitted to the nephele job manager: eu.stratosphere.nephele.executiongraph.GraphConversionException: java.lang.RuntimeException: The user defined 'configure()' method in the Output Format caused an error: Invalid configuration for CsvOutputFormat: Need to specify number of fields > 0.
    at eu.stratosphere.pact.runtime.task.DataSinkTask.initOutputFormat(DataSinkTask.java:300)
```

rmetzger:
Okay. Its is working. I have some versions-mixup in my cluster.

PR #418, opened 2014-01-20
Added broadcast variables support to the runtime.

aalexandrov:
This is a first step towards realizing the functionality requested in issue #66 in the runtime, building upon the API proposed in PR #383.

Broadcast variables are realized as additional outputs attached to a `RegularPactTask`. Convenience methods for accessing the configuration information for broadcast input setup was added to the `TaskConfig`.

An integrated test case for with a simple hardwired JobGraph can be found in `BoradcastVarsNepheleITCase`.

### DIFF
--- a/stratosphere-core/src/main/java/eu/stratosphere/api/common/functions/RuntimeContext.java
+++ b/stratosphere-core/src/main/java/eu/stratosphere/api/common/functions/RuntimeContext.java
@@ -12,6 +12,7 @@
  **********************************************************************************************************************/
 package eu.stratosphere.api.common.functions;
 
+import java.util.Collection;
 import java.util.HashMap;
 
 import eu.stratosphere.api.common.accumulators.Accumulator;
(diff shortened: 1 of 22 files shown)


aalexandrov:
I would like to get feedback on the proposed changes to the runtime from @StephanEwen. 

In particular, I wasn't able to decouple the modifications in the `RegularPactTask` from concrete implementation of the record model (see the comment on line 412 in the modified `RegularPactTask`).

However, since `Record` is used as a concrete type parameter in other places of the `RegularPactTask`. I guess that the issue runs deeper and the decoupling could be done as a refactoring at once later.

aalexandrov:
OK ignore the last comment I was able to resolve the issue with the hard-coded `Record` after a chat with @StephanEwen. The fix is up in an ammended commit.

rmetzger:
Redundant with https://github.com/stratosphere/stratosphere/pull/460

PR #421, opened 2014-01-21
Introduced CREATE and OVERWRITE file output write modes.

fhueske:
Introduced CREATE and OVERWRITE file output write modes. OVERWRITE is default (as before).
Introduced optional OUTPUT_DIRECTORY parameter to specify write location of output tasks with DOP = 1. 
Rewrote code for file output path preparation.
Added FileOutputFormatTest.
Fixed distributed writing to local filesystem #286

### DIFF
--- a/stratosphere-core/src/main/java/eu/stratosphere/api/common/io/FileOutputFormat.java
+++ b/stratosphere-core/src/main/java/eu/stratosphere/api/common/io/FileOutputFormat.java
@@ -56,6 +56,17 @@ public abstract class FileOutputFormat<IT> implements OutputFormat<IT> {
 	public static final String WRITEMODE_OVERWRITE = "stratosphere.output.writemode.overwrite";
 	
 	/**
+     * The key under which the output directory mode parameter is stored in the configuration
+     */
+    public static final String OUT_DIRECTORY_PARAMETER_KEY = "stratosphere.output.directory";
+    
+    /**
+     * Value keys for the output directory modes
+     */
+    public static final String OUT_DIRECTORY_ALWAYS = "stratosphere.output.directory.always";
+    public static final String OUT_DIRECTORY_PARONLY = "stratosphere.output.directory.paronly";
+    
+	/**
 	 * The config parameter for the opening timeout in milliseconds.
 	 */
 	public static final String OUTPUT_STREAM_OPEN_TIMEOUT_KEY = "stratosphere.output.file.timeout";
@@ -71,6 +82,11 @@ public abstract class FileOutputFormat<IT> implements OutputFormat<IT> {
 	protected WriteMode writeMode;
 	
 	/**
+	 * The output directory mode
+	 */
+	protected OutputDirectoryMode outDirMode;
+	
+	/**
 	 * The stream to which the data is written;
 	 */
 	protected FSDataOutputStream stream;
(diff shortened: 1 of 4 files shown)


stratosphere-core/src/main/java/eu/stratosphere/api/common/io/FileOutputFormat.java
```diff
@@ -44,6 +45,17 @@
 	public static final String FILE_PARAMETER_KEY = "pact.output.file";
 	
 	/**
+	 * The key under which the write mode is stored in the configuration
+	 */
+	public static final String WRITEMODE_PARAMETER_KEY = "pact.output.writemode";
```
rmetzger: Seems like we missed some `pact.` stuff here?
Do you think it makes sense to rename them with your PR?
I would suggest not to introduce new `pact.` prefixed keys.

stratosphere-core/src/main/java/eu/stratosphere/core/fs/FileSystem.java
```diff
@@ -390,13 +399,191 @@ public abstract FSDataOutputStream create(Path f, boolean overwrite, int bufferS
...
+	 * Since concurrent DataSinkTasks are not aware of each other, coordination of delete and create
+	 * operations would be difficult.
+	 * 
+	 * @param outPath Output path that should be prepared.
+	 * @param writeMode Write mode to consider. 
+	 * @param parallelOutput True, if it will be written to the output path in parallel, false otherwise.
+	 * @return True, if the path was successfully prepared, false otherwise.
+	 * @throws IOException
+	 */
+	public boolean initOutPathLocalFS(Path outPath, WriteMode writeMode, boolean parallelOutput) throws IOException {
```
rmetzger: I did not read all changes you suggest yet but: wouldn't it make sense to have a abstract `initOutPath(..)` in `FileSystems` and implement the cases for local and distributed in their respective classes.
With the current approach, we would have to add a method for each new file system we add to the system.

stratosphere-core/src/main/java/eu/stratosphere/api/common/io/FileOutputFormat.java
fhueske: Damn, I wanted to rename the keys but forgot.
Will push an updated PR.

rmetzger:
@fhueske: we once had the discussion whether it is better to create one output file in the dop=1 case or always use a directory.
How hard would it be to make this configurable (in the FileOutputFormat) ?

stratosphere-core/src/main/java/eu/stratosphere/core/fs/FileSystem.java
fhueske: There is no need to write an initialization function for a new filesystem.
Both functions in `FileSystem` should cover all use cases. The distinction is, that `initOutPathLocalFS()` needs to be called by every `FileOutputFormat` because the target file system is local (currently only LocalFS) and cannot be remotely accessed. `initOutPathDistFS()` works on distributed, remotely accessible filesystems (HDFS, S3, ...) and is called at planning time in the JobManager. So initialization happens only once.

JobManager and OutputFormat decide based on the abstract function `FileSystem.isDistributedFS()` whether to call the initialization function or not. So every file system must only implement `isDistributedFS()` and not an initialization function.

fhueske:
Shouldn't be a big deal.

fhueske:
I've added an optional parameter to specify that always a new directory is created (even for DOP = 1).

StephanEwen:
Looks good. Configurable for overwrite and DOP 1 create file or directory. Nice!

The defaults are overwrite and DOP 1 = directory?

fhueske:
The defaults are as before, i.e., OVERWRITE and directories only for DOP > 1.

I think, CREATE write mode is more conservative and a better default. However, it changes the behavior of all existing jobs and we would need to adapt many tests. Therefore, I kept the defaults as they were before and left this as a future issue / discussion.

 So, are we good to merge?

StephanEwen:
Can you add config parameters for the defaults? That would be good. After
that, I think we a re good to merge.

fhueske:
Merged PR and opened an issue to add system config parameters (#449).

PR #423, opened 2014-01-22
removeField method in Record class

jcamachor:
Is there a reason to not have a method that deletes (projects) a field from a Record?

The code of the method has been taken from an older version of Stratosphere, where it was commented out. I just needed to update it, and now it works with the current version.

### DIFF
--- a/stratosphere-core/src/main/java/eu/stratosphere/types/Record.java
+++ b/stratosphere-core/src/main/java/eu/stratosphere/types/Record.java
@@ -426,28 +426,31 @@ public final class Record implements Value {
 	}
 	
 	/**
-	 * Removes the field at the given position.
+	 * Removes the field at the given position. 
+	 * <p>
+	 * This method should be used carefully. Be aware that as the field is actually removed from the record, the
+	 * total number of fields is modified, and all fields to the right of the field removed shift one position to
+	 * the left.
 	 * 
-	 * @param field The index number of the field to be removed.
+	 * @param fieldNum The position of the field to be removed, starting at zero.
 	 * @throws IndexOutOfBoundsException Thrown, when the position is not between 0 (inclusive) and the
 	 *                                   number of fields (exclusive).
 	 */
-	public void removeField(int field)
+	public void removeField(int fieldNum)
 	{
 		// range check
-		if (field < 0 || field >= this.numFields) {
+		if (fieldNum < 0 || fieldNum >= this.numFields) {
 			throw new IndexOutOfBoundsException();
 		}
 		int lastIndex = this.numFields - 1;		
 
-		if (field < lastIndex) {
-			int len = lastIndex - field;
-			System.arraycopy(this.offsets, field + 1, this.offsets, field, len);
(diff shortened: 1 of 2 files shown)


fhueske:
_If I remember correctly_, the reason for removing the `removeField()` method was that is is hard to reason about.
When this method is called, all fields with a higher index are shifted by one which makes it hard to keep track of the field indexes. We mostly set fields that were no longer needed to null (via `setNull(i)` or `setField(i, null)`) which preserves the field indexes.

However, this is just an assumption. There might also be some issues with the function itself.
The Record object is quite complex and I cannot tell if the function is correct right now.
@StephanEwen is the expert for this class ;-)

In the long run, we are moving away from the record-based Java API. It will still be offered and supported, but we are in the progress of drafting a new one which makes the whole thing easier to use.

StephanEwen:
I think it is fine to re-add this functionality.

We are moving away from the Record, but until then, I see no reason not to
have this functionality.

fhueske:
I think this is a question of whether the Record API will be supported in the future or whether we will drop it completely. 
If we keep it (some people are using it and would need to do a second port after the refactoring), I would vote to not add the method. I think, we had a good reason to remove it. The weird semantics of shifting indexes is very confusing to users. However, if there is a good reason to include it, I would be fine with adding it.

@jcamachor: Would setting the fields to null work for you as well? The overhead of having sparse records should be negligible, right @StephanEwen ?

jcamachor:
> We are moving away from the Record

What do you mean? Will you kill the Record API in future releases? Will you stick to the key-value pairs?

Independently, I think the method could be added to the Record class, so the user is given the **freedom** to use it. You are not enforcing him to use it, and if he does, it is up to him to work properly with it. The name `removeField` or `projectField` would be already self explanatory. Further, a mention _'Use carefully as this method actually removes a field and shift indexes in a record'_ could be added to the method.

> @jcamachor: Would setting the fields to null work for you as well?

Sure, I could do that. But at the PAXQuery algebra level, that's weird semantics, as 'projecting a field' would not change indexes in a record (which is the expected behaviour).

rmetzger:
I think we will not kill the record API immediately.
We will develop a second, much improved Java API for Stratosphere. Our internals are flexible enough to provide different APIs at the same time.

> Will you stick to the key-value pairs?
> We do not have key-value pairs currently and I think the new API won't have them either.

@jcamachor: Can you write a little comment to the method that warns the users? Can you also uncomment the test-case that belongs to the remove method? (in `eu.stratosphere.types.RecordTest`).

You can change code in a pending pull request by adding comments to the branch the PR is based on.

StephanEwen:
The Record API will stay for all the things that were built on top of it.

The rational for the new Java API is to add compile time safety for Java
Programs, less confusing play with positions in the records, and strong
schema for runtime performance. It will be added, it will not be a
replacement.

As Robert mentioned, the architecture has a generic API at the optimizer
level, and then specialized on top of it (Record, Scala, new Java)

rmetzger:
Thank you for your contribution. I merged it in b7217d35ce01a4c4fec4ace8f3155e1a389214c0.

Issue #425, opened 2014-01-23
Use Iterable instead of Iterator for the messages in Spargel (cosmetic)


sscdotopen:
Spargel provides an Iterator over the messages. Iterables are more handy as they allow for loops.

StephanEwen:
Can we do an IterableIterator? One that is both, returning 'this' on the
'getIterator()' call?

Spargel provides an Iterator over the messages. Iterables are more handy as
they allow for loops.

—
Reply to this email directly or view it on
GitHubhttps://github.com/stratosphere/stratosphere/issues/425
.

sscdotopen:
I think in our case this should be fine, people would not expect to be able to create two different iterators from the messages.

rmetzger:
Fixed with PR https://github.com/stratosphere/stratosphere/pull/433

PR #433, opened 2014-01-23
Provide Iterable instead of Iterator to updateVertex method of Spargel API 

uce:
This is issue #425.

I did this, before @StephanEwen and @sscdotopen discussed the `IterableIterator` in #425. Why do we want to expose the Iterator?

### DIFF
--- a/stratosphere-addons/spargel/src/main/java/eu/stratosphere/spargel/java/examples/connectedcomponents/SpargelConnectedComponents.java
+++ b/stratosphere-addons/spargel/src/main/java/eu/stratosphere/spargel/java/examples/connectedcomponents/SpargelConnectedComponents.java
@@ -72,10 +72,10 @@ public class SpargelConnectedComponents implements Program, ProgramDescription {
 		private static final long serialVersionUID = 1L;
 
 		@Override
-		public void updateVertex(LongValue vertexKey, LongValue vertexValue, Iterator<LongValue> inMessages) {
+		public void updateVertex(LongValue vertexKey, LongValue vertexValue, Iterable<LongValue> inMessages) {
 			long min = Long.MAX_VALUE;
-			while (inMessages.hasNext()) {
-				long next = inMessages.next().getValue();
+			for (LongValue msg : inMessages) {
+				long next = msg.getValue();
 				min = Math.min(min, next);
 			}
 			if (min < vertexValue.getValue()) {
--- a/stratosphere-addons/spargel/src/main/java/eu/stratosphere/spargel/java/examples/connectedcomponents/SpargelConnectedComponents.java
+++ b/stratosphere-addons/spargel/src/main/java/eu/stratosphere/spargel/java/examples/connectedcomponents/SpargelConnectedComponents.java
@@ -46,7 +46,6 @@ public class SpargelConnectedComponents implements Program, ProgramDescription {
 		FileDataSource initialVertices = new FileDataSource(DuplicateLongInputFormat.class, verticesPath, "Vertices");
 		FileDataSource edges = new FileDataSource(new CsvInputFormat(' ', LongValue.class, LongValue.class), edgesPath, "Edges");
 		
-		// create DataSinkContract for writing the new cluster positions
 		FileDataSink result = new FileDataSink(CsvOutputFormat.class, resultPath, "Result");
 		CsvOutputFormat.configureRecordFormat(result)
 			.recordDelimiter('\n')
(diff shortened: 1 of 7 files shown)


sscdotopen:
Exposing the iterator is part of the Iterable interface :)

uce:
Yes. I meant why not just provide Iterable to the user?

sscdotopen:
We discuss the proposal to provide an Iterable to the user. This is more 
convenient and in sync with what other systems like Hadoop or Giraph do.

There is only one problem: You can call iterator() on an Iterable to get 
an Iterator for the underlying data.

Now, in general you could create two Iterators and advance them 
individually in different orders. This doesn't work in our case, as we 
don't want to materialize all the records of the Iterable in memory, 
instead Stratosphere deserializes them one by one according to next() 
calls on the Iterator (afaik).

I think always returning the same iterator instance should be fine as it 
is reasonable. We should simply document it.

uce:
OK, if the others agree, I would add the missing documentation.

@rmetzger I wanted to add a test case based on `SpargelConnectedComponents` example, but had a problem with the dependencies. When I add `stratosphere-tests` as a (test-scoped) dependency to Spargel, I only have access to `TestBase2`, which is in the main branch of `stratosphere-tests`. But I also want to build on existing tests, which provide utility methods for connected components.

It worked in IntelliJ, when I provided the following two dependencies:

``` XML
<!-- TestBase2 -->
<dependency>
    <artifactId>stratosphere-tests</artifactId>
    <scope>test</scope>
</dependency>
<dependency>
    <artifactId>stratosphere-tests</artifactId>
    <type>test-jar</type>
    <scope>test</scope>
</dependency>
```

This works in the IDE, but not in Maven on the CLI (which makes sense, because we are not packaging the test-jars, right?). Is there a way to do it ([a](http://stackoverflow.com/questions/7000812/how-can-i-make-the-test-jar-include-dependencies-in-maven), [b](http://nerdgerl.wordpress.com/2009/07/24/creating-a-test-jar-in-maven/))?

rmetzger:
You forgot to export the code of `stratosphere-tests` as a jar.
I pushed the test to a branch in my repository, see the changes here: https://github.com/rmetzger/stratosphere/compare/rtfm

uce:
OK, I didn't test the advice from [the link I provided](http://nerdgerl.wordpress.com/2009/07/24/creating-a-test-jar-in-maven/), because I thought that we had to somehow rely solely on the Maven assembly plugin for packaging.

I'll add it to this PR.

uce:
The `stratosphere-tests` test jars have to be deployed for the new Spargel test case to work.

I hope it will work with the addition of the `maven-jar-plugin` to `stratosphere-tests/pom.xml`.

StephanEwen:
Do we need a test jar? Does it not work to add to spargel a dependency on
stratosphere-tests in scope test?
Am 24.01.2014 16:35 schrieb "Ufuk Celebi" notifications@github.com:

rmetzger:
No, thats not enough.
The problem is that maven does not package the contents of `src/test/java` into the `jar` file.
My change builds now two jar-files for the `stratosphere-tests` project (one with the tests).
I dont see any disadvantages for my approach.

StephanEwen:
What classes from stratosphere-tests do you need? The test base is under
src/main/java, so it should be packaged and available by default. I am not
against a test jar, I am only wondering what I am missing...
Am 26.01.2014 01:33 schrieb "Robert Metzger" notifications@github.com:

uce:
The helper methods from test. In this case, NepheleConnectedComponentsITCase for the odd/even result checking and stuff.

StephanEwen:
Looks good to me.

stratosphere-addons/spargel/src/main/java/eu/stratosphere/spargel/java/examples/connectedcomponents/SpargelConnectedComponents.java
```diff
@@ -72,10 +71,10 @@ public Plan getPlan(String... args) {
 		private static final long serialVersionUID = 1L;
 
 		@Override
-		public void updateVertex(LongValue vertexKey, LongValue vertexValue, Iterator<LongValue> inMessages) {
+		public void updateVertex(LongValue vertexKey, LongValue vertexValue, Iterable<LongValue> inMessages) {
```
StephanEwen, line 74: Make this an IterableIterator? Or is that confusing to people?

rmetzger:
merged in 71a721cabcb7df38054668aca97255e7b545b6ef and subsequent commits.

PR #435, opened 2014-01-24
First prototype of language-binding for python

filiphaase:
The first prototype consists of a small python- and java-framework, which can currently work with an Reduce and Map-Operator and sent over sockets and stdio. Google Protocol Buffers are used to send records between the java and python processes. 

This commit also includes an running-wordcount example. The UDFs can be defined in python, while the configuration and Plan-setup still has to be made in java.


stratosphere-addons/stratosphere-language-binding/src/main/java/eu/stratosphere/language/binding/wordcountexample/PythonWordCount.java
```diff
@@ -0,0 +1,180 @@
...
+		plan.setDefaultParallelism(numSubTasks);
+		return plan;
+	}
+
+	@Override
+	public String getDescription() {
+		return "Parameters: [numSubStasks] [input] [output]";
+	}
+	
+	private static final Log LOG = LogFactory.getLog(JobManager.class);
```
rmetzger, line 163: always pass the name of the own class to `getLog`, in this case `PythonWordCount.class`

fhueske:
Nice work!

I guess you have this already on your roadmap, but in any case: You can extend `MapFunction` and `ReduceFunction` to `PyMapFunction` and `PyReduceFunction` to hide much of the stream setup and make the example WordCount program more concise and attractive. Also, 8080 is a port commonly used for HTTP. Maybe using a less popular port would be better.

StephanEwen:
8080 is also used for our own web client by default ;-)

fhueske:
I thought about how to represent the data flow in external languages.
Couldn't we build a generic compiler in Java, that translates data flows defined in JSON into Stratosphere programs. Then, each language binding could offer native language primitives to build a data flow, compile it to JSON, and generate a Stratosphere program from that.

Any ideas?

rmetzger:
We thought about something similar, but using Protobuf for that. So our approach would represent the Graph using a set of protobuf objects. At the end of the day, it does not matter which data serialisation we're using.

Here is a list of pro and cons:

JSON:
- Human Readable
- String-based, so easier to use from all languages.
- Verbose than a binary representation
- One more dependency since we need a JSON parser.

Protobuf:
- We already use it

@filiphaase is going to work on this this week to figure out whats the best approach for this. We will discuss his results on Friday.

filiphaase:
@fhueske Thanks :) Your definitively right about the PyMapFunction,... classes and they will absolutely be part of the next version of the prototype. But currently I still have hardcoded values in my classes(Path of python script, the classes of each value of the records, connectionType and if necessary the port), so I would still have to overwrite the open() function to include them in this code. Therefore I would like to include the extended classes in the next version. 

And for the JSON/Protobuf discussion: Like robert stated, I'm currently working on the protobuf-version to transfer the plan to java, and therefore enabling the user to use stratosphere by just writing python. Since I'm already using protobuf for the tuple streaming and know how to use it, it's not very hard to use protobuf for transfering the plan and hopefully it will work until friday :) 

Since I'm a stratosphere-newbie I also have two general questions:
- Can I update this pull-request by just pushing into filiphaase:langbinding-pr ? If not, how can I update my code?
- I need to transfer some values like the python-script-path and the classes for each value of the records to my Map/ReduceFunctions. As I see it using setParameter() and the Configuration-Object is the way to do this. But I can not call it with a list of classes. Is there any way I could give something else than string,int or boolean to my Functions? Otherwise my best idea is to give a String-Representation of the classes to my functions(f.e."StringValue,IntValue")

fhueske:
@filiphaase Thanks for sharing your plans. Sounds good! 
Regarding you questions:
- You can update your PR by simply pushing into the same branch. If you changed some previous commits, you need to force the push (`git push origin abc --force`).
- We're moving a bit away from the configuration. You can also pass objects to the operators (instead of classes). The objects need to be serializable (implement the interface). Then Java serialization will take care of moving the object state which can be set by a custom constructor or settters.

filiphaase:
@fhueske Thanks for the quick answer, I will try it that way :)

rmetzger:
We will not merge this pull request into the `master`-branch since the advanced python support is making good progress. (we are also planning to release 0.5 soon). 
Full python support (with python-based plan creation) goes into 0.6.

PR #424, opened 2014-01-22
Hadoop compat

faisalmoeen:
Implemented issue #382.

There is an org.apache package which was needed to extend Task class of hadoop which has package visibility. Any suggestions to solve this?

PR #437, opened 2014-01-25
hadoop-compat addons project with HadoopDataSource

rmetzger:
I picked up the code of https://github.com/stratosphere/stratosphere/pull/424 and enhanced it with the following:
- make hadoop-compat compatible to hadoop yarn and remove code in package `org.apache.hadoop`.
- Extended usercode object wrapper (@aljoscha, please validate my changes), fixed InputFormat serialization
- introduce pluggable type converter
- Add a generic wrapper for Hadoop's `Writable` and `WritableComparable`.

The generic wrapper allows to do the following: (The example is from Mongodb's Hadoop InputFormat)

``` java
public void map(Record record, Collector<Record> out) throws Exception {
    WritableWrapper wrap = record.getField(0, WritableWrapper.class);
    Writable wr = wrap.value();
    Writable valWr = record.getField(1, WritableWrapper.class).value();
    BSONWritable bson = (BSONWritable) wr;
    BSONWritable value = (BSONWritable) valWr;
    System.err.println("bson value has "+value.toString());
}
```

Lets discuss if we want the project being called "hadoop-compat" or if you prefer "hadoop-compatability" ?

Open issues:
- If the Hadoop IF is a FileInputFormat, we should make the split assignment file locality aware
- Add junit tests for various formats such as ORC, Parquet and Avro.
- Write website documentation for this code.
- Add HadoopDataSink
- See if it is possible to map hadoop counters to our accumulators


uce:
Nice. I vote for `hadoop-compatability` as `compat` is imho not a standard acronym.

What do you mean with "Add HadoopDataSource"? Typo for **sink**?

rmetzger:
I saw `compat` being used by some Linux developers ;)
And the source was a typo, I meant sink.

I added an additional converter that allows to use any Hadoop `Writable`!

fhueske:
Not sure, if I would call it either. 
This feature gives "only" support for Hadoop InputFormats, but there is lots of other stuff in Hadoop which we do not support right now such as OutputFormats, Map, and Reduce functions.
If we plan to extend the support for these interfaces as well, I would go with `hadoop-compatibility`. 
Otherwise, `hadoop-input` would be a good name IMHO.

rmetzger:
The plan is to put everything related to hadoop compatability into this package. 
I think the next step would be a `OutputFormat`.

I will change the name!

StephanEwen:
We are thinking about adding proper interface compatibility with Hadoop
later. That could all be in one project, called  hadoop-compatibility.

stratosphere-core/src/main/java/eu/stratosphere/api/common/operators/util/UserCodeObjectWrapper.java
```diff
@@ -33,30 +34,47 @@
...
+				/**
+				 * Check if the usercode class has custom serialization methods.
+				 * (See http://docs.oracle.com/javase/7/docs/api/java/io/Serializable.html for details).
+				 * We can not guarantee that the user handles serialization correctly in this case.
+				 */
+				boolean hasCustomSerialization = false;
+				Method customSerializer = null;
+				Method customDeserializer = null;
+				try {
+					customSerializer = current.getClass().getDeclaredMethod("writeObject", java.io.ObjectOutputStream.class);
```
rmetzger, line 51: @aljoscha: Could you review this change? (I committed it already to master) ... but it would be good if you can confirm my approach.

Issue #147, opened 2013-10-10
Make Stub and Input/Output Formats Serializable


StephanEwen:
Currently, all stubs need to manually declare serializability, to be able to be shipped as code. This causes non-runnable stubs, if they forget this declaration (easy trap) and ugly user code.

We should either
- declare serializability in the Stub, so it does not need to be added to all implementations. It still required to either ignore the serializability warning, or declare a serial version uid. - OR -
- use a different serializer than the java default serializer, fir example kryo. I am not sure how that interplays with the scala frontend, since scala serializes closures through java's default serialization.

We may be able to use both, meaning we use java serializability if a stub declared the java.io.Serializable interface, and use kryo otherwise. That would lead to the best support for user code.

aljoscha:
At the beginning I declared serializability in the stub. I decided against this though because serializability does not just mean implementing Serializable, you also have to make sure that your class is actually serializable. It has to be a deliberate choice by the user.

rmetzger:
I don't know if you know that @aljoscha's implementation is able to handle both, classes and objects as stubs.

I think Hadoop users are used to define classes in their stubs and configure them using the Configuration. If (advanced) Stratosphere users want to pass objects, they have to take care of serializability.

uce:
I would be curious to see how much we would gain by using Kryo instead of Java serialization (see https://github.com/eishay/jvm-serializers/wiki for some numbers).

Scala classes need some extra setup though (https://code.google.com/p/kryo/#Scala).

Spark for example uses Java serialization for closures (see `spark.closure.serializer` at http://spark.incubator.apache.org/docs/latest/configuration.html) and recommends Kryo for general object serialization (see `spark.serializer`) although Java serialization is the default (because users have to manually register their classes with Kryo).

aljoscha:
But we only use java serialization in non-performance-critical stuff, right? So I wouldn't worry about this to much.

uce:
Yeah, sure. I think I wanted to say that we might also think about the performance-critical stuff in the long term.

aljoscha:
So, will we close this issue?

uce:
Yes, I think so.

A different issue would be to think about using classes instead of objects in the examples if we think that @StephanEwen points (easy trap + ugly code) are valid.

StephanEwen:
Yes, I think it is resolved as follows:

[...]

PR #438, opened 2014-01-25
Discussion: Allow Anonymous Classes for Functions in Plan

uce:
**This PR is just for discussion.**

I've been thinking about our Java API and @StephanEwen is also prototyping a new API, which will bring the Java and Scala APIs closer together (among other things) by moving the Java API to the distributed DataSet abstraction instead of operator wiring (which is imho super important and will be a big plus).

With the upcoming support for Lambda expressions in Java 8, the upcoming API will be very concise. Since this will only be supported in Java 8, I assume that we will provide separate Java 6/7 and Java 8 APIs. For the Java 6/7 API, I thought that it might be worthwhile to support anonymous classes as function implementations.

`AnonymousWordCount` of this PR shows the default `WordCount` implementation with anonymous classes instead of static nested classes, which makes it imho more readable.

But there is a problem with serializability. With anonymous classes you can only implement a single interface (e.g. `MapFunction`), but we also need to implement `Serializable` (and make sure that everything is indeed serializable). For this discussion I've declared `AbstractFunction` as `Serializable` as a work around, but it has problems (see #147). One way around this would be to provide something like `SerializableMapFunction`. But I find it somehow ugly and inconvenient in terms of usability for newcomers.

So the questions for discussions are:
- What do you think about supporting this?
- What is the best solution if we want it?

### DIFF
--- a/stratosphere-core/src/main/java/eu/stratosphere/api/common/Program.java
+++ b/stratosphere-core/src/main/java/eu/stratosphere/api/common/Program.java
@@ -13,11 +13,13 @@
 
 package eu.stratosphere.api.common;
 
+import java.io.Serializable;
+
 /**
  * Interface for classes that represent Stratosphere programs. The program creates the {@link Plan}, which is the
  * instance of the program that will be executed.l
  */
-public interface Program {
+public interface Program extends Serializable {
 	
 	/**
 	 * The method which is invoked by the compiler to get the job that is then compiled into an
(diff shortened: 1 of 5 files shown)


StephanEwen:
I have stumbled over that issue as well. I think serializability on the
abstractstub level would be fine. I think we should also promote the object
passing instead of class passing as the default case.
Am 25.01.2014 05:10 schrieb "Ufuk Celebi" notifications@github.com:

fhueske:
I like to have anonymous classes as an option. This is much nicer to use for simple UDFs. 
Giving The user the choice between regular and anonymous classes is the right way to go, I think.

StephanEwen:
I vote for making the AbstractFunction serializable.

rmetzger:
This PR is mergable, right? AbstractFunction is serializable.

rmetzger:
Merged in d74d4bf7e360de4c230314285725a54472e095d5

PR #439, opened 2014-01-27
new blog post on accessing mongodb from stratosphere

rmetzger:
Preview: http://robertmetzger.de/stratosphere/news/blog/tutorial/2014/01/28/querying_mongodb.html


fhueske:
Sweet! Nice work!

You might want to add, what the example program is doing with the data in MongoDB.
How about putting the example code somewhere in the stratosphere repository instead of linking against your repo? This post might exists for quite some time.

There might also be some follow up issues:
- Is the format also integrated with the Scala API. If not, we should open another issue for that. 
- We should also add it to the documentation.

StephanEwen:
Looks good!

How did you implement the writable wrapper? Does it serialize the class of
the type every time, or is there some fancy different mechanism?

Am 27.01.2014 09:20 schrieb "Robert Metzger" notifications@github.com:

rmetzger:
> You might want to add, what the example program is doing with the data in MongoDB.

Thats already in the text: "The example program reads data from the enron dataset that contains about 500k internal e-mails. The data is stored in MongoDB and the Stratosphere program counts the number of e-mails per day."

> How about putting the example code somewhere in the stratosphere repository instead of linking against your repo? This post might exists for quite some time.
> Agree, will do.

You are right, documentation and Scala support is missing. I think we need to find a generic way to support Stratosphere Java InputFormats in Scala.

@StephanEwen: The wrapper is not fancy at all ;) It serializes the type as a string all the time. Feel free to suggest improvements (if the usability does not suffer).

StephanEwen:
The only way to avoid that is with static type info, similar to the way I
did it with the Avro value. Have a look at that.

I am a bit torn between the the PROs of writing such a blog post and the
CONs of making it depend on such a workaround that is really not feasible
for anything real where you expect only a little bit performance.

For the new type safe record API, we need to directly support:
1. Out Value types
2. Java's primitive types
3. Hadoop's writable types
4. Avro types

There are nice ways of doing this efficiently, without needing to serialize
types.

aalexandrov:
While we're at MongoDB, there is an upcoming [Meetup in the MongoDB user group](http://www.meetup.com/MUGBerlin/events/161353802/) in Berlin. 

I know the guy who is presenting and I'm going this Thurday from 6PM, and I also have a +2 for the event if you are interested.

Issue #33, opened 2013-06-12
Rework instance configuration.


StephanEwen:
Right now, Nephele still uses the EC2-inspired instance configuration model. The Pact compiler connects to obtain information about these instances, such as how many are available, and how much memory they have. This is error prone to configure and also a bit buggy, it frequently leads to wrong memory bookkeeping if different instance types are configured.

Do we need support for heterogeneous setups where different nodes have different capabilities and should be assigned a different amount of work? If we defer this to later, we can greatly simplify the logic and configuration:

1) No configuration for the instance type. The internal instance manager has a default profile which is okay for all cluster instances.

2) An explicit value of how many slots for parallel operators we have on each node (such as 8 on an eight core machine). There should be a default value in the config which could be overridden via query-specific parameters.

3) An explicit config entry that defines how much memory should be used for networking and how much for query processing. The query processing memory amount is used to initialize the MemoryManager and is also used by the pact-compiler to parameterize the memory available to the operators. That way we can also get rid of the communication between the compiler and the job manager on plan compilation. Eventually it would be good to run the compiler as a child process of the job-manager anyways.

In the long run we want to make query processing memory and network memory one value (overall system memory, the rest is the UDF Java heap memory) which is shared for materialization in the network stack and the runtime operators.

PR #443, opened 2014-01-28
Speeded up startup time for Local Executor and Tests

StephanEwen:
Involves the following:
- Faster registration of TM at JM
- Optional lazy memory allocation in TM (used in local and test mode)
- Some cleanup in TM startup


stratosphere-runtime/src/main/java/eu/stratosphere/nephele/taskmanager/TaskManager.java
```diff
@@ -92,6 +90,11 @@
 public class TaskManager implements TaskOperationProtocol {
 
 	private static final Log LOG = LogFactory.getLog(TaskManager.class);
+	
+	private final static int FAILURERETURNCODE = -1;
```
uce, line 94: Can you rename this to `FAILURE_RETURN_CODE`?

uce:
Nice! :-)

stratosphere-runtime/src/main/java/eu/stratosphere/nephele/instance/local/LocalInstanceManager.java
```diff
@@ -115,54 +100,39 @@
 	 *        the path to the configuration directory
 	 */
 	public LocalInstanceManager() {
-
-		final Configuration config = GlobalConfiguration.getConfiguration();
```
uce, line 119: Removing lines 119--134 causes `LocalInstanceManagerTest` to fail, because `createDefaultInstanceType()` (line 103 in your version) returns the characteristics of the caller instead of reading them from the config file.

I think we should just remove `LocalInstanceManagerTest` and someone should finally do #33.

**Note**: When I tried to run the test inside of IntelliJ, I noticed that the path extraction for the config also doesn't work, whereas it works on the command line via `mvn`. In IntelliJ, instead of .../stratosphere **/stratosphere-runtime/** src/test/resources it returns _.../stratosphere/src/test/resources_ w/o the subdirectory.

StephanEwen:
That is fixed by now. Some of the legacy instance manager code is failing non-deterministically at this point. Seems that it somewhat relied on a delayed heartbeat.

rmetzger:
This pull request has not been merged, right?

PR #445, opened 2014-01-28
Adapted KMeans example

fhueske:
Adapted KMeans example. Point-cluster assignments are also written as result in addition to the cluster centers.
Adapted KMeans TestCases.


StephanEwen:
Looks good. Feel free to merge.

PR #446, opened 2014-01-29
Added K-Means Demo to Quickstart (Website)

rmetzger:
Preview: http://robertmetzger.de/stratosphere/quickstart/

Thanks to @fhueske for fixing the KMeans example and writing the python tool!


uce:
Great idea!

Some suggestions:
- What about the following ordering on the quickstart page: Setup, **Example**, Java, Scala
- For the lead paragraph, I would remove the linebreak before "on the way"
- What about adding K-Means to the title (and other places) instead of just clustering
- The markdown markup (pun intended) is not working for `*relative standard deviation*`

StephanEwen:
Beautiful. Like Ufuk said, some markup stuff is not working, bot otherwise,
I like it...

rmetzger:
Thanks. I fixed the issues (I did not add k-means into the box on the quickstart page, but its in the text below).

fhueske:
Some remarks:
- Since I changed the KMeans Job (added a data source, two operators, and a sink) you should update the screenshots of the web frontend that show the plan and the progress monitoring.
- The plots of the raw data have not the same size.

rmetzger:
I updated the figures and merged it!

PR #453, opened 2014-02-02
hadoop-compat addons project: package renaming, compile-time type safety, usercode-classloader.

rmetzger:
This pull request fixes the hadoop-compatibility project to use the usercode-classloader.

I also introduced the Hadoop MR `<K, V>` types for key and value.

The intention of this pull request was to allow the following syntax:

```
BSONWritable value = record.getField(1, WritableWrapper.class).value();
```

But I was not able to achieve this, even though the method signature of `value()` looks like this:
`public <X extends Writable> X value()`.
The error message is `Type mismatch: cannot convert from Writable to BSONWritable`.

I guess its because of the generics usage in the `Record.getField()` method because the whole thing works if used like this:

```
WritableWrapper<Writable> test = new WritableWrapper<Writable>();
BSONWritable aa = test.value();
```

Has anybody an idea how to solve this? (or a good explanation)


rmetzger:
TODO: Test on cluster

rmetzger:
merged in 169e456d3bea1252952ed1d706318901d004c540

Issue #395, opened 2014-01-13
Release 0.4.1 Planning


rmetzger:
With this issue, I want to plan the first bugfix release in the 0.4 line.

List of pull requests and commits for 0.4.1.
- `start-local.bat` fix for paths with spaces https://github.com/stratosphere/stratosphere/pull/394
- NPE in Scala interface https://github.com/stratosphere/stratosphere/pull/409
- improved error message in spargel/iterations https://github.com/stratosphere/stratosphere/pull/432
- fix in execution graph test https://github.com/stratosphere/stratosphere/pull/420
- NPE in Scala interface (commit: eedfe1e0e5598630dd446e89db4c68d122f3d32a) reporting issue: https://github.com/stratosphere/stratosphere/issues/434
- Tests failing with non-us locale https://github.com/stratosphere/stratosphere/issues/416 f17ccfb52eac786497dd2a7d150f5e361cfca5b1
- YARN URL fix: 5300f4b4b4cfdb00edbd5b4e87721640b0c7dc0c
- dist-binary fix: 00676c56a875a987ce23371c1665e7d28c1f46ee, travis fix: 82542d68edd5999cd7cd8330bb7a3b19852b8eb6

fhueske:
- Configurable write modes for FileDataSink: #421
- fix for distributed writing to local FS: #421
- improved YARN package #485

StephanEwen:
Subsumed by 0.5 release planning.

PR #459, opened 2014-02-05
deploy spargel with dist

rmetzger:
We currently do not ship `spargel` with the stratosphere downloadable binary.
The spargel out-of-the-box experience is bad since the quickstart does not use `jar-with-dependencies` as the default packaging method (for some good reasons). But, users can not simply add `spargel` as a dependency to their project and run it on a cluster.

I suggest to include this PR to the 0.4.1 release https://github.com/stratosphere/stratosphere/issues/395

### DIFF
--- a/stratosphere-dist/pom.xml
+++ b/stratosphere-dist/pom.xml
@@ -55,6 +55,12 @@
 
 		<dependency>
 			<groupId>eu.stratosphere</groupId>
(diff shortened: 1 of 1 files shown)


rmetzger:
Merged in 9b717e4971f73633f8cd52f111f8af494507ae50

PR #468, opened 2014-02-11
Stratosphere Job to generate input data for wordcount

rmetzger:
I hacked together a little utility Stratosphere job that generates input data for wordcount. I wanted to have this to demonstrate Stratosphere with a few TB of input data on an Amazon AWS cloud.

Its not sophisticated at all, uses the same seed for all data sources and the generated strings are equally random distributed, not zipf or so. (So its not very good for a Wordcount-benchmark).

I also changed the current word count to generate some (really small) input data. It generates data automatically if the user did not specify any arguments. BUT: Our Wordcount now runs out of the box!

### DIFF
--- a/stratosphere-examples/stratosphere-java-examples/pom.xml
+++ b/stratosphere-examples/stratosphere-java-examples/pom.xml
@@ -298,6 +298,7 @@
 
 							<includes>
 								<include>**/wordcount/*.class</include>
+								<include>**/util/*.class</include>
 							</includes>
 						</configuration>
 					</execution>
--- a/stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/record/util/WordCountDataGenerator.java
+++ b/stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/record/util/WordCountDataGenerator.java
@@ -1,0 +1,85 @@
+package eu.stratosphere.example.java.record.util;
+
+import java.io.IOException;
+import java.io.Serializable;
+import java.util.Iterator;
+import java.util.Random;
+
+import org.apache.commons.lang.RandomStringUtils;
+
+import eu.stratosphere.api.common.Plan;
+import eu.stratosphere.api.common.Program;
+import eu.stratosphere.api.common.operators.CollectionDataSource;
+import eu.stratosphere.api.common.operators.FileDataSink;
+import eu.stratosphere.api.common.operators.GenericDataSource;
+import eu.stratosphere.api.java.record.io.CsvOutputFormat;
+import eu.stratosphere.api.java.record.io.FileOutputFormat;
+import eu.stratosphere.api.java.record.io.GenericInputFormat;
+import eu.stratosphere.client.LocalExecutor;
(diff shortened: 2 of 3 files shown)


aalexandrov:
We have a [wordcount generator](https://github.com/TU-Berlin-DIMA/wordcount-gen) on top of the that [Myriad Toolkit](https://github.com/TU-Berlin-DIMA/myriad-toolkit) which follows the distribution of the top 100000 words in the English Gutenberg corpus.

Take a look at the actual [distribution specification here](https://github.com/TU-Berlin-DIMA/wordcount-gen/tree/master/src/config/distributions/lexicon) and at the [CombinedPrFunction](https://github.com/TU-Berlin-DIMA/myriad-toolkit/blob/master/src/cpp/math/probability/CombinedPrFunction.h) for an implementation of skewed sampling based on a uniform double sequence.

BTW all Myriad Generators now ship with a [map-only hadoop-job](https://github.com/TU-Berlin-DIMA/myriad-driver), so you can run them on the cluster without too much effort. I would welcome a PR with a similar implementation for Stratosphere.

PS. Data skew actually doesn't influence performance for wordcount.

rmetzger:
Cool! my suggested generator had a really poor performance on Amazon EC2. 
But their machines only have a disk write throughput of 20 MB/s.
I thought generating 3 TB of wordcount data would be a matter of an hour or so.

Does the Map only job take care of shipping the binary within the cluster? 
Do you think I'm able to get the myriad+MR tool to run within 10 minutes? If yes, I'll try it. Otherwise I'll stick with hamlet.

aalexandrov:
In my experience data generation is mostly bounded by the CPU. On my machine the Myriad version spills out ~ 11MB/s, so 20 seems pretty fast.

If you want to try it out on a Ubuntu image on EC2 I might be able to send you a binary which works out of the box. Otherwise you'll have to follow the instructions from the Readme and build it on site (takes ~ 10 mins time).

To start the Hadoop job you have to type the following command:

``` bash
${HDP_MAPR_BIN}/hadoop jar \
    ${WC_DGEN_HOME}/bin/wordcount-gen-driver-jobs.jar \
    ${WC_DGEN_HOME} \
    -s${SCALING_FACTOR} \
    -N${NODE_COUNT} \
    -m${DATASET_ID} \
    -o${HDFS_PATH} -xtoken
```

This will start a Haoop job with DOP of `${NODE_COUNT}` and generate `${SCALING_FACTOR}` GB of data in the `${HDFS_PATH}/${DATASET_ID}` folder.

rmetzger:
20 MB/s is the performance writing zeros to disk ;)

The build instructions are quite intimidating. It goes beyond a simple "git clone and make", right?
The EMR instances are Amazon AMI (which seems to be redhat-based). Not sure if it will work with your binary. Can you send it to me anyway?

aalexandrov:
They are intimidating cause I didn't now better at the time of writing the generic README.md which is shipped with all Myriad-based project.

Actually it boils down to the 4 commands [described here](https://github.com/TU-Berlin-DIMA/wordcount-gen#building-on-unix--linux-distributions).

rmetzger:
Lets discard this code for now.

PR #470, opened 2014-02-12
Website: Blogpost: Use Stratosphere with Amazon Elastic MapReduce

rmetzger:
Please review the blog post.

Preview: http://robertmetzger.de/stratosphere/tutorial/blog/2014/02/12/amazon-elastic-mapreduce-cloud-yarn.html

I will now go through the steps one more time to validate that everything is working. Then, I'll update everything once again, adding your feedback.


fhueske:
Nice post!

A few comments:
- 3GB heap is not really necessary for the JM.
- I think, it is confusing to use the uber Jar a second time as the job Jar file. It would be better to use the dedicated Wordcount example jar file, even if it means to do another download, IMHO. 

On a side note:
Can't we create a "proper" YARN bundle, with the uber Jar, example job jars, (start) scripts, etc?

rmetzger:
Okay. I addressed all your comments and updated the post accordingly.
Once the build for the yarn changes passed, I'll publish the post.

uce:
Nice. Thank you for the post.

rmetzger:
I published the post on the website

PR #460, opened 2014-02-05
Added broadcast variables support to the optimizer.

aalexandrov:
This is the third an final PR for issue #66 after PR #383 and PR #418. With this PR broadcast variables should be supported in trivial dataflows where the broadcast input is within the same iteration scope.

Please check [my comment](https://github.com/stratosphere/stratosphere/pull/383#issuecomment-33047450) in PR #383 for a proposed solution in order to add proper support for broadcast variables with nested iterations.

PR #471, opened 2014-02-13
Broadcast Variables

StephanEwen:
The pull request merges @aalexandrov 's pull request #460 with the current master and adds support for iterations. Also includes a variety of cleanups.

Missing is the support for branching plans. I would suggest to merge this part now, such that newer patches can build on it.

Since the K-Means example is affected by this change, we need to add an additional variant that assigns the  center id to teh data point, to support the clustering tutorial.


StephanEwen:
There is something goofy about this pull request. It includes commits d1e7d06c1c9f30e9e0220d2ce6cafb1fac68a5ea and 2ee0255c8d9e070aa1f77f9ad485573aec695d0e, which are subsumed in the master already. However, when I merge the master into this PR, it says that everything is up to date. Rebasing this PR on the master fails.

Do not merge until this git problem is resolved.

rmetzger:
Sadly, it is not possible to apply the pr as a single patch to master.

```
$ wget https://github.com/stratosphere/stratosphere/pull/471.patch
$ git apply --check 471.patch
error: patch failed: stratosphere-compiler/src/main/java/eu/stratosphere/compiler/PactCompiler.java:952
error: stratosphere-compiler/src/main/java/eu/stratosphere/compiler/PactCompiler.java: patch does not apply
error: patch failed: stratosphere-addons/avro/src/test/java/eu/stratosphere/api/avro/EncoderDecoderTest.java:19
error: stratosphere-addons/avro/src/test/java/eu/stratosphere/api/avro/EncoderDecoderTest.java: patch does not apply
error: patch failed: stratosphere-addons/avro/pom.xml:41
error: stratosphere-addons/avro/pom.xml: patch does not apply
error: patch failed: stratosphere-tests/src/main/java/eu/stratosphere/test/util/TestBase2.java:66
error: stratosphere-tests/src/main/java/eu/stratosphere/test/util/TestBase2.java: patch does not apply
```

If you want to try the merge, use `git am --signoff 471.patch`

aalexandrov:
> [...] we need to add an additional variant that assigns the center id to the data point [...].

What do you mean with that? How does this differ from the test jobs added in a6f0603 and ba11e19?

StephanEwen:
We need that not for testing, but to support this quickstart example here:

http://stratosphere.eu/quickstart/example.html

aalexandrov:
Can't we just modify the adapted KMeansTest instead of adding a variant? It would still work as a test case, right?

StephanEwen:
Yes, we could. I was a bit hesitant, because this would not show core
K-Means any more, rather a modified variant tailored to that demo case.

I vote for keeping the core examples simple and to think first in terms of
presentation of examples (as these are somewhat acting like documentation
as well) and then in terms of tests (these are internal and somewhat
invisible).

So from the view of the test, I guess it is fine. Another question is
whether we want two variants of KMeans in the examples.

aalexandrov:
Alright, agreed. We should close this PR before we get another conflict.

StephanEwen:
Commit history is fixed.

I will add the additional KMeans example (so we do not break the current demo) and then merge.

StephanEwen:
Merged in 0c7fe86c398a5fff33f92d9970a7ebe8e517639c

PR #472, opened 2014-02-13
Termination Criterion for Bulk Iterations

markus-h:
I implemented the setTerminationCriterion method for BulkIterations. 

I did my best to rebase my changes to the current version. Unfortunately the current test base is again not Windows compatible, so I was not able to run the tests.

Since the changes are quite complex I would highly appreciate if some of you would test the new termination criterion with some of their currently working iterations. The adjustment to the new criterion should be not too hard.


markus-h:
I had to adjust the ConnectedComponentsNepheleITCase testcase because in my opinion flags for the IterationTailPactTask were not set correctly,

StephanEwen:
You refer to the flags that mark a tail as updating the workset or the
solution set?

markus-h:
Yes. For the solution set only isSolutionSetUpdate was set but not isWorksetIteration. For the current implementation of the termination criterion to work it hast to be speciefied as workset iteration. Before my changes it didn't matter.

markus-h:
I do not understand why travis fails to compile. I compiled and run all testcases on a cluster and it worked.

StephanEwen:
Merged in 2f1050f7e0505dc9bd23e417be9d6d208065eb2a

Thanks for the feature!

PR #477, opened 2014-02-14
Adjusted file URIs in some tests for Windows compatibility

markus-h:
In some of the newer tests filesystem URIs were again hard coded. I did a quick adjustment for Windows to get the test running in Windows environments again.
All tests that are running external shell commands were excluded when running under windows for now.

### DIFF
--- a/stratosphere-java/src/test/java/eu/stratosphere/api/java/record/io/TextInputFormatTest.java
+++ b/stratosphere-java/src/test/java/eu/stratosphere/api/java/record/io/TextInputFormatTest.java
@@ -73,10 +73,12 @@ public class TextInputFormatTest {
 			
 			Record r = new Record();
 			assertTrue("Expecting first record here", inputFormat.nextRecord(r));
-			assertEquals(FIRST, r.getField(0, StringValue.class).getValue());
+			String field = r.getField(0, StringValue.class).getValue().replace("\r", ""); // Windows bug?
+			assertEquals(FIRST, field);
 			
 			assertTrue("Expecting second record here",inputFormat.nextRecord(r ));
-			assertEquals(SECOND, r.getField(0, StringValue.class).getValue());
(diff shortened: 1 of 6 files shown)


StephanEwen:
In other places, we used file.toURI().toString(), rather than hardcoding the "file:/" or "file:///" prefix. Can we do the same thing here?

markus-h:
You are right, that is the nicer solution. I adjusted my changes.

stratosphere-java/src/test/java/eu/stratosphere/api/java/record/io/TextInputFormatTest.java
```diff
@@ -73,10 +73,12 @@ public void testPositionBug() {
 			
 			Record r = new Record();
 			assertTrue("Expecting first record here", inputFormat.nextRecord(r));
-			assertEquals(FIRST, r.getField(0, StringValue.class).getValue());
+			String field = r.getField(0, StringValue.class).getValue().replace("\r", ""); // Windows bug?
+			assertEquals(FIRST, field);
 			
 			assertTrue("Expecting second record here",inputFormat.nextRecord(r ));
-			assertEquals(SECOND, r.getField(0, StringValue.class).getValue());
+			field = r.getField(0, StringValue.class).getValue().replace("\r", ""); // Windows bug?
```
StephanEwen, line 80: This is actually not a windows bug. The test data is created with a PrintWriter/PrintStream. The println() function uses the OS dependent line delimiter. On Unix that is `\n`, on Windows it is `\r\n`.

This is a good indication that our text input format does not work with Windows properly.

I would suggest to change the test such that the println() function is avoided and new lines are added manually.

StephanEwen:
I opened #505 to address the text input format on windows issue.

StephanEwen:
Fixed and merged in d0ef7c48bba8c55a6d932ba7e2a19d2278b65a75

Issue #487, opened 2014-02-19
Move gh-pages branch to seperate repository


uce:
The gh-pages branch blows up our repo and initial checkout takes quite long. Also both things are seperate.

We might want to move gh-pages to a seperate repo.

Note: if we do this, we have to fix the "Edit this page" link of the documentation.

rmetzger:
Hi,
we should certainly do this! I will create a repository in the stratosphere organization.
I would like to do the migration on the next Sunday, thats the day where we have the least traffic.
I'm not sure how the migration will work. We can not just push the gh-pages into the master of the web-repo. GitHub will complain because the domain is already registered with another repository. 
Can you ask GitHub how to do the migration?

rmetzger:
Okay, I'm doing it now.

I pushed the branch to the new repository. Now I'll check if it is working from the new one.

rmetzger:
Okay, the move was successful. I'll delete the `gh-pages` branch here.

PR #489, opened 2014-02-20
Initial BCVars documentation

twalthr:


docs/0.5/index.markdown
```diff
@@ -19,4 +20,47 @@ The JDBC input and output formats allow to read from and write to any JDBC-acces
...
+    @Override
+    public void map(Record record, Collector<Record> out) {       
+       for (Record r : myBcRecords) {
+       	   // do something with the records
+       }
+
+    }
+}
+{% endhighlight %}
+*Note*: As the content of broadcast variables is kept in-memory on each node, it should not become too large. Currently broadcast variable do not support branching, therefore it is not possible to use <code>someBcInput</code> (in the above example) as input for another operator.
```
fhueske: We use the term "branching" (plan or operator) only internally and I think it might confuse users who are not familiar with the concept.
I would therefore avoid using this term in the documentation and simply state, that at the moment a bc input can not be used as another input (either regular input or as another bc input). Also mention, that we plan to drop this limitation in a future release.

StephanEwen: Good point, I agree with Fabian.

StephanEwen:
We can update that one and remove the part that branching is not supported

twalthr:
I will do it tomorrow.

rmetzger:
Please reopen the pull request against the new website repository (https://github.com/stratosphere/stratosphere.github.io) See also https://github.com/stratosphere/stratosphere/issues/487

Issue #464, opened 2014-02-09
Accessing a job in the job history right after it finished leads to NPE


rmetzger:
If you want to inspect a finished job right after it finished, the web gui will not load the JSON data.

In the jobmanager log file, you find the following NPE:

```
20:58:13,694 WARN  eu.stratosphere.nephele.jobmanager.web.JobmanagerInfoServlet  - java.lang.NullPointerException
    at eu.stratosphere.nephele.jobmanager.web.JobmanagerInfoServlet.writeJsonForArchivedJob(JobmanagerInfoServlet.java:229)
    at eu.stratosphere.nephele.jobmanager.web.JobmanagerInfoServlet.doGet(JobmanagerInfoServlet.java:81)
    at javax.servlet.http.HttpServlet.service(HttpServlet.java:707)
    at javax.servlet.http.HttpServlet.service(HttpServlet.java:820)
    at org.eclipse.jetty.servlet.ServletHolder.handle(ServletHolder.java:532)
    at org.eclipse.jetty.servlet.ServletHandler.doHandle(ServletHandler.java:453)
    at org.eclipse.jetty.server.session.SessionHandler.doHandle(SessionHandler.java:227)
    at org.eclipse.jetty.server.handler.ContextHandler.doHandle(ContextHandler.java:965)
    at org.eclipse.jetty.servlet.ServletHandler.doScope(ServletHandler.java:388)
    at org.eclipse.jetty.server.session.SessionHandler.doScope(SessionHandler.java:187)
    at org.eclipse.jetty.server.handler.ContextHandler.doScope(ContextHandler.java:901)
    at org.eclipse.jetty.server.handler.ScopedHandler.handle(ScopedHandler.java:117)
    at org.eclipse.jetty.server.handler.HandlerList.handle(HandlerList.java:47)
    at org.eclipse.jetty.server.handler.HandlerWrapper.handle(HandlerWrapper.java:113)
    at org.eclipse.jetty.server.Server.handle(Server.java:352)
    at org.eclipse.jetty.server.HttpConnection.handleRequest(HttpConnection.java:596)
    at org.eclipse.jetty.server.HttpConnection$RequestHandler.headerComplete(HttpConnection.java:1048)
    at org.eclipse.jetty.http.HttpParser.parseNext(HttpParser.java:549)
    at org.eclipse.jetty.http.HttpParser.parseAvailable(HttpParser.java:211)
[...]

PR #495, opened 2014-02-24
fix NullPointException in web interface when inspecting a job immediately after it finished

qmlmoon:
issue #464

### DIFF
--- a/stratosphere-runtime/resources/web-docs-infoserver/js/jobmanagerFrontend.js
+++ b/stratosphere-runtime/resources/web-docs-infoserver/js/jobmanagerFrontend.js
@@ -324,20 +324,13 @@ function updateTable(json) {
 		if(event.newstate == "FINISHED" || event.newstate == "FAILED" || event.newstate == "CANCELED") {
 			// stop polling
 			pollfinished = true;
-			
-			// add to history
-			var jobjson = {};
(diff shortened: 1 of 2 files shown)


markus-h:
Where exactly did this NullPointerException occur? Is events null? Because we're iterating over it at two places.
If I understand your fix correctly you are iterating now fully over jobmanager.getOldJobs() and thereby writing out all state changes of all archived jobs. That produces a lot of network overhead because we are only intersted in the state changes of one specific job.

rmetzger:
@markus-h Are your concerns addressed by the last commit? (So is the PR ready to merge?)

markus-h:
Yes it is addressed and can be merged.
For the future I'm still not sure if the new solution is a good one. 8 secs wait time is really long. Perhaps we should think about a better solution at some time.

rmetzger:
So the refresh-rate of the interface (also for running jobs?) has been set to 8 seconds with this pull request?

qmlmoon:
No. Only for the job that is finished, it will be shown in the job history
8 secs later after it finishes

2014-02-26 19:21 GMT+01:00 Robert Metzger notifications@github.com:

StephanEwen:
Merged in 7aacbf3994cca859d4097ebea117934899a7b684

Issue #25, opened 2013-05-21
Offer a buffer-oriented abstraction from Nephele to Pact


StephanEwen:
Currently, Nephele offers a record-oriented API to Pact, which frequently results in unnecessary date (de)serialization.

StephanEwen:
## Current State
1.  Each Output (Writer/Gate) has a local pool of buffers, which is shared among all channels of the output. Different channels go to different endpoints, e.g. in a hash-redistribute. The Writer evaluates the ChannelSelector (for example the PactRecordOutputEmitter) and send the record to all relevant channels. Each channels serializes the record into its current buffer. If the buffer overflows, it grabs a new one from the pool. Full buffers are dispatched to the network for transfer. The serializers have basically long byte arrays into which the first materialize the record. Then they copy it to the buffer, prepended  by a 4 byte length field.
2. Each input channel has a local buffer pool. When the network thread reads an envelope header (with target channel id), it grabs a buffer from that channel's pool to read the remainder into. It then queues the envelope in the channel, and notifies the gate of the available envelope. When a record is to be read from an input gate, it takes any input channel with available data. Each channel has an individual deserializer that gets the buffer, reads the length fields and and then deserializes the record.
3. Nephele maintains a global buffer pool with buffers for the network stack When input or output gates are instantiated, the local buffer pools grab buffers from the global buffer pool. Nephele occasionally redistributes the buffers when new gates are instantiated to make sure that all buffers are used, but non is 

Note: Since #25 that the buffers in the local pool are the same MemorySegment types as used in the Pact layer.

## Changes needed
1. Buffer pool for the writers is allocated by the pact program.
2. The writer offers an API `sendBuffer(MemorySegment buffer, int offset, int len, int targetChannel)` and `broadcasthBuffer(MemorySegment buffer, int offset, int len)`. The writer constructs the envelope and send the buffer to the network stack. Once the buffer has been sent by the network stack, it is returned to the buffer pool. The mechanism of buffer-recyclers can be kept. To support sending the same buffer to many receivers (broadcast), the envelope constructed by the writer may have an atomic reference counter that is increased for every channel the buffer is added to and decreased for every time it is send. Reaching zero, the buffer is added back to the pool.
3. There is a nephele record-oriented writer, that does the serialization of records to buffers in a generic manner, much like the current channel code does it. This one is for nephele uses and prototyping, it is not used by the pact runtime code.
[...]

PR #511, opened 2014-02-26
Offer buffer-oriented API for I/O (#25)

uce:
The runtime offered a record-oriented API (#25) for data transfers, which
- resulted in unnecessary data (de)serialization,
- complicated the upcoming fault tolerance implementation, and
- blocked more efficient implementations of higher-level operators.

With this commit, the runtime offers a buffer-oriented API for the output side (sending), which is oblivious to records. The buffer oriented input side (receiving) is still to be implemented.

---
- Previously, the I/O classes allowed different implementations of channels, gates etc. with the `ByteBuffered*` implementation being the default one. I've merged those classes to only have the default implementation (e.g. InputChannel, OutputChannel), which greatly simplifies the code.
- Every I/O class had an associated `*Context`, which was initiated at runtime for buffer allocation. All these classes could be removed, since the I/O stack does not allocate buffers anymore, but instead is fed buffers to send/receive.
- Record serialization has been moved from the channels (for the output side, input side is still pending) to the writers/readers. For example `RecordWriter` still offers the record-oriented API, but serializes records into buffers and dispatches full buffers via the new buffer-oriented API (`BufferWriter`). 
- The (previosuly singleton) `GlobalBufferPool` (for network buffers) is now instantiated by the `ChannelManager` for each TaskManager. The network buffers are used by `RecordWriter` for record serialization and by the input channels (will be changed) for deserialization. In the future, the buffers can directly come from the upper layers (formerly PACT).
- I've reorganized the touched classes into the package `eu.stratosphere.runtime.*` (previously `eu.stratosphere.nephele.*`) to start with the removal of the deprecated Nephele/PACT separation.


StephanEwen:
Merged into the development branch for network stack reorganization and fault tolerance at https://github.com/StephanEwen/stratosphere/tree/netstack_rework as of 4506d1f973121b6833aec0a9c52f4d7cfe69f6f9

PR #512, opened 2014-02-26
[tachyon] implemented

zentol:
Implementation of the Tachyon file system.

The code was tested using junit and the wordcount example on a tachyon file system running in local mode.

### DIFF
--- a/stratosphere-addons/tachyon/pom.xml
+++ b/stratosphere-addons/tachyon/pom.xml
@@ -1,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
+	
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+	
+    <modelVersion>4.0.0</modelVersion>
+	
+    <parent>
+        <artifactId>stratosphere-addons</artifactId>
+        <groupId>eu.stratosphere</groupId>
+        <version>0.5-SNAPSHOT</version>
+        <relativePath>..</relativePath>
+    </parent>
+
+    <artifactId>tachyon</artifactId>
+    <name>tachyon</name>
+
+    <packaging>jar</packaging>
+
+    <dependencies>
+        <dependency>
+            <groupId>eu.stratosphere</groupId>
+            <artifactId>stratosphere-core</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.tachyonproject</groupId>
+            <artifactId>tachyon</artifactId>
+            <version>0.4.0</version>
+        </dependency>
+    </dependencies>
+</project>
--- a/stratosphere-addons/tachyon/pom.xml
+++ b/stratosphere-addons/tachyon/pom.xml
@@ -26,7 +26,7 @@
         <dependency>
             <groupId>org.tachyonproject</groupId>
             <artifactId>tachyon</artifactId>
-            <version>0.4.0</version>
+            <version>0.4.1</version>
         </dependency>
         <dependency>
             <groupId>eu.stratosphere</groupId>
--- a/stratosphere-addons/tachyon/pom.xml
+++ b/stratosphere-addons/tachyon/pom.xml
@@ -29,6 +29,12 @@
             <version>0.4.0</version>
         </dependency>
         <dependency>
+            <groupId>eu.stratosphere</groupId>
+            <artifactId>stratosphere-java-examples</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
 		<groupId>org.slf4j</groupId>
 		<artifactId>slf4j-api</artifactId>
 		<version>1.7.2</version>
--- a/stratosphere-addons/tachyon/pom.xml
+++ b/stratosphere-addons/tachyon/pom.xml
@@ -28,5 +28,17 @@
             <artifactId>tachyon</artifactId>
             <version>0.4.0</version>
         </dependency>
+        <dependency>
+		<groupId>org.slf4j</groupId>
+		<artifactId>slf4j-api</artifactId>
+		<version>1.7.2</version>
+                <scope>test</scope>
+	  </dependency>
+	  <dependency>
+		<groupId>org.slf4j</groupId>
+		<artifactId>slf4j-log4j12</artifactId>
+		<version>1.7.2</version>
+                <scope>test</scope>
+	  </dependency>
     </dependencies>
 </project>
--- a/stratosphere-addons/tachyon/src/main/java/eu/stratosphere/fs/tachyon/TachyonBlockLocation.java
+++ b/stratosphere-addons/tachyon/src/main/java/eu/stratosphere/fs/tachyon/TachyonBlockLocation.java
@@ -1,3 +1,16 @@
+/***********************************************************************************************************************
+ * Copyright (C) 2010-2013 by the Stratosphere project (http://stratosphere.eu)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ **********************************************************************************************************************/
+
 package eu.stratosphere.fs.tachyon;
 
 import eu.stratosphere.core.fs.BlockLocation;
@@ -37,6 +50,6 @@ public class TachyonBlockLocation implements BlockLocation {
      */
     @Override
     public int compareTo(BlockLocation o) {
-        return ((Long)(offset)).compareTo(o.getOffset());
+        return ((Long) (offset)).compareTo(o.getOffset());
     }
 }
--- a/stratosphere-addons/tachyon/src/main/java/eu/stratosphere/fs/tachyon/TachyonBlockLocation.java
+++ b/stratosphere-addons/tachyon/src/main/java/eu/stratosphere/fs/tachyon/TachyonBlockLocation.java
@@ -37,6 +37,6 @@ public class TachyonBlockLocation implements BlockLocation {
      */
     @Override
     public int compareTo(BlockLocation o) {
-        return (new Long(offset)).compareTo(o.getOffset());
+        return ((Long)(offset)).compareTo(o.getOffset());
     }
 }
--- a/stratosphere-addons/tachyon/src/main/java/eu/stratosphere/fs/tachyon/TachyonBlockLocation.java
+++ b/stratosphere-addons/tachyon/src/main/java/eu/stratosphere/fs/tachyon/TachyonBlockLocation.java
@@ -1,0 +1,42 @@
+package eu.stratosphere.fs.tachyon;
+
+import eu.stratosphere.core.fs.BlockLocation;
+import java.io.IOException;
+
+/**
+ * Implementation of the {@link BlockLocation} interface for the Tachyon File System.
+ */
+public class TachyonBlockLocation implements BlockLocation {
+    private final String[] hosts;
+    private final long offset;
+    private final long length;
+
+    public TachyonBlockLocation(String[] hosts, long offset, long length) {
+        this.hosts = hosts;
+        this.offset = offset;
+        this.length = length;
+    }
+
+    @Override
+    public String[] getHosts() throws IOException {
+        return hosts;
+    }
+
+    @Override
+    public long getOffset() {
+        return offset;
+    }
+
+    @Override
+    public long getLength() {
+        return length;
+    }
+
+    /**
+     * Compares two BlockLocations based on their offset.
+     */
+    @Override
+    public int compareTo(BlockLocation o) {
+        return (new Long(offset)).compareTo(o.getOffset());
+    }
+}
--- a/stratosphere-addons/tachyon/src/main/java/eu/stratosphere/fs/tachyon/TachyonFileSystem.java
+++ b/stratosphere-addons/tachyon/src/main/java/eu/stratosphere/fs/tachyon/TachyonFileSystem.java
@@ -1,3 +1,16 @@
+/***********************************************************************************************************************
+ * Copyright (C) 2010-2013 by the Stratosphere project (http://stratosphere.eu)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ **********************************************************************************************************************/
+
 package eu.stratosphere.fs.tachyon;
 
 import eu.stratosphere.core.fs.BlockLocation;
--- a/stratosphere-addons/tachyon/src/main/java/eu/stratosphere/fs/tachyon/TachyonFileSystem.java
+++ b/stratosphere-addons/tachyon/src/main/java/eu/stratosphere/fs/tachyon/TachyonFileSystem.java
@@ -82,7 +82,11 @@ public class TachyonFileSystem extends FileSystem {
             }
             return null;
         } catch (IOException ioe) {
-            throw new FileNotFoundException();
+            if (ioe.getCause() instanceof FileDoesNotExistException) {
+                throw new FileNotFoundException(ioe.getCause().getMessage());
+            } else {
+                throw ioe;
+            }
         }
     }
 
--- a/stratosphere-addons/tachyon/src/main/java/eu/stratosphere/fs/tachyon/TachyonFileSystem.java
+++ b/stratosphere-addons/tachyon/src/main/java/eu/stratosphere/fs/tachyon/TachyonFileSystem.java
@@ -38,7 +38,7 @@ public class TachyonFileSystem extends FileSystem {
      */
     @Override
     public Path getWorkingDirectory() {
-        return new Path(pathToString(workingDirectory));
+        return workingDirectory;
     }
 
     /**
@@ -59,7 +59,7 @@ public class TachyonFileSystem extends FileSystem {
     @Override
     public void initialize(URI name) throws IOException {
         this.name = URI.create(name.getScheme() + "://" + name.getAuthority());
-        fileSystem = TachyonFS.get(name.getAuthority());
+        fileSystem = TachyonFS.get(name.toString());
     }
 
     /**
@@ -71,13 +71,13 @@ public class TachyonFileSystem extends FileSystem {
     @Override
     public FileStatus getFileStatus(Path path) throws IOException {
         List<ClientFileInfo> tachyInfo = fileSystem.listStatus(pathToString(path.getParent()));
-        FileStatus stratInfo = null;
+        int targetID = fileSystem.getFileId(pathToString(path));
         for (int x = 0; x < tachyInfo.size(); x++) {
-            if (tachyInfo.get(x).path.equals(pathToString(path))) {
-                stratInfo = new TachyonFileStatus(tachyInfo.get(x), name);
+            if (tachyInfo.get(x).id==targetID) {
+                return new TachyonFileStatus(tachyInfo.get(x), name);
             }
         }
-        return stratInfo;
+        return null;
     }
 
     @Override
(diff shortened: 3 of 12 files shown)


stratosphere-addons/tachyon/src/test/java/eu/stratosphere/core/fs/tachyon/TachyonFileSystemTest.java
```diff
@@ -0,0 +1,143 @@
...
+        out = fs.create(testFilePath2, true, 0, (short) 0, 8);
+        out.write("i-am-to-big-for-one-block".getBytes());
+        out.close();
+
+        info = fs.getFileBlockLocations(fs.getFileStatus(testFilePath2), 0, 15);
+        assertEquals(2, info.length);
+    }
+
+    @Test
+    public void testOpen_Path_int() throws Exception {
```
uce: Could you use Java CamelCase capitalization here and in the other methods?

uce:
Thank you for this PR! I read over it and didn't notice anything major.

We would have to test it in a distributed setup though before merging.

stratosphere-addons/tachyon/src/main/java/eu/stratosphere/core/fs/tachyon/TachyonBlockLocation.java
```diff
@@ -0,0 +1,43 @@
+package eu.stratosphere.core.fs.tachyon;
+
+import eu.stratosphere.core.fs.BlockLocation;
+import java.io.IOException;
+
+/**
+ * Implementation of the {@link BlockLocation} for the
+ * Tachyon File System. *
```
uce: We use 120 as line length and you should remove the last `*` (also in other comments).

stratosphere-addons/tachyon/src/main/java/eu/stratosphere/core/fs/tachyon/TachyonFileSystem.java
```diff
@@ -0,0 +1,253 @@
...
+import java.util.List;
+import tachyon.client.ReadType;
+import tachyon.client.TachyonFS;
+import tachyon.client.WriteType;
+import tachyon.thrift.ClientBlockInfo;
+import tachyon.thrift.ClientFileInfo;
+import tachyon.thrift.NetAddress;
+
+/**
+ * Implementation of the {@link FileSystem} for the Tachyon File System.
```
uce: typo: {@link FileSystem} => {@link FileSystem} **interface**

stratosphere-addons/tachyon/src/test/java/eu/stratosphere/core/fs/tachyon/TachyonFileSystemTest.java
```diff
@@ -0,0 +1,143 @@
...
+
+        out = fs.create(testFilePath1, true, 0, (short) 0, 8);
+        out.write("i-fit".getBytes());
+        out.close();
+
+        info = fs.getFileBlockLocations(fs.getFileStatus(testFilePath1), 0, 1);
+        assertEquals(1, info.length);
+
+        out = fs.create(testFilePath2, true, 0, (short) 0, 8);
+        out.write("i-am-to-big-for-one-block".getBytes());
```
uce: Doesn't matter, but it's "too" big ;-)

uce:
How did you test it with WordCount? If it is not too much work (and makes sense), we could add the following:

If I'm not mistaken (from what I've seen in the test case), you can instantiate a local Tachyon FS at runtime and write data to it. So we could add something like TachyonITCase, which writes the word count test data to a local Tachyon FS and then runs WordCount on it.

zentol:
i started tachyon manually via console using a provided .sh skript and created a class that calls Wordcount.main(...).

initialize only connects to an existing tachyon file system, it does not create one.
but if one manages to get that skript running it is possible to write the test.

uce:
OK, makes sense. After we figure out if it is possible to start Tachyon easily from Java, we can just adapt https://github.com/amplab/tachyon/blob/master/src/main/java/tachyon/examples/BasicOperations.java for WordCount.

stratosphere-addons/tachyon/pom.xml
```diff
@@ -0,0 +1,32 @@
...
+    <dependencies>
+        <dependency>
+            <groupId>eu.stratosphere</groupId>
+            <artifactId>stratosphere-core</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.tachyonproject</groupId>
+            <artifactId>tachyon</artifactId>
+            <version>0.4.1-SNAPSHOT</version>
```
rmetzger: I would suggest to use the stable 0.4.0 version of tachyon in maven central.
http://search.maven.org/#artifactdetails%7Corg.tachyonproject%7Ctachyon%7C0.4.0%7Cjar

Somehow, our test server is not able to access the SNAPSHOT version .. setting the version to one that is available in central will solve the problem.

StephanEwen: Looks good initially. I'll take a closer look later.

The only thing that strikes me as a bit strange is that the packages are
part of "eu.stratosphere.core". Can you move them to something like
"eu.stratosphere.fs.tachyon" or "eu.stratosphere.addons.fs.tachyon"? I
personally have a bias towards the first option.

zentol:
@rmetzger : i just tried using the 0.4.0 version, and now my code is broken. :/ it seems that connecting to the filesystem works differently now, will look into it.

@StephanEwen : sure. this would mean though that there are 3 different places where filesystems are placed (core.fs / runtime.fs / the new one), and in case of the first option also at different depths. i dont know though whether that is relevant or not though.

rmetzger:
Is there also a 0.4.1 in central?

So that you do not have to change the code.

Sent from my iPhone

On 26.02.2014, at 20:27, zentol notifications@github.com wrote:

@rmetzger https://github.com/rmetzger : i just tried using the 0.4.0
version, and now my code is broken. :/ it seems that connecting the the
filesystem works differently now, will look into it.

@StephanEwen https://github.com/StephanEwen : sure. this would mean
though that there are 3 different places where filesystems are placed
(core.fs / runtime.fs / the new one), and in case of the first option also
at different depths. i dont know though whether that is relevant or not.

## 

GitHubhttps://github.com/stratosphere/stratosphere/pull/512#issuecomment-36165907
.

stratosphere-addons/tachyon/pom.xml
uce: I agree with @StephanEwen.

zentol:
@rmetzger : no, theres no 0.4.1. version available.

rmetzger:
Okay. It seems that the travis-tests are failing since tachyon is (obviously) not running.
There is either a way to start tachyon in a pseudo mode, or we have to disable the test for now. If you do not find a solution for starting Tachyon out of java, ask on the Tachyon mailing list.

We also have no automated tests for HDFS. (Which is bad as well).

zentol:
i started trying to start it from java, but ran very early into NoSuchMethodErrors (and Im quite clueless about fixing that) , so right now i cant say for sure whether it is possible or not. i have posted a question on the mailing list.

one thing i should (maybe) mention is that starting tachyon with the scripts always required me to enter the sudo password, would that be an issue?

rmetzger:
Okay, let's see what they say.

On Sat, Mar 1, 2014 at 6:49 PM, zentol notifications@github.com wrote:

> i started trying to start it from java, but ran very early into
> NoSuchMethodErrors (and Im quite clueless about fixing that) , so right now
> i cant say for sure whether it is possible or not. i have posted a question
> on the mailing list.
> 
> one thing i should (maybe) mention is that starting tachyon with the
> scripts always required me to enter the sudo password, would that be an
> issue?
> 
> ## 
> 
> Reply to this email directly or view it on GitHubhttps://github.com/stratosphere/stratosphere/pull/512#issuecomment-36431164
> .

## 

zentol:
just got a response, they have a class that does exactly what we want, that they use themselves for tests. 
i tried integrating it, but still hit NoSuchMethodErrors on my end. in case its just my system being the cause i still committed it to see what travis says. hope thats alright.

rmetzger:
Hi,
cool.
The error is this one, right?

```
14/03/01 21:10:41 ERROR : Failed to connect (1) to master localhost.localdomain/127.0.0.1:18998 : java.net.ConnectException: Connection refused
Exception in thread "Thread-2" java.lang.NoSuchMethodError: org.slf4j.spi.LocationAwareLogger.log(Lorg/slf4j/Marker;Ljava/lang/String;ILjava/lang/String;[Ljava/lang/Object;Ljava/lang/Throwable;)V
    at org.eclipse.jetty.util.log.JettyAwareLogger.log(JettyAwareLogger.java:607)
    at org.eclipse.jetty.util.log.JettyAwareLogger.warn(JettyAwareLogger.java:431)
    at org.eclipse.jetty.util.log.Slf4jLog.warn(Slf4jLog.java:69)
    at org.eclipse.jetty.util.component.AbstractLifeCycle.setFailed(AbstractLifeCycle.java:204)
    at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:74)
    at tachyon.web.UIWebServer.startWebServer(UIWebServer.java:83)
    at eu.stratosphere.fs.tachyon.util.Master.start(Master.java:188)
    at eu.stratosphere.fs.tachyon.util.LocalTachyonCluster$1.run(LocalTachyonCluster.java:150)
    at java.lang.Thread.run(Thread.java:744)
```

I'm guessing their webinterface is using a different version of the log4j package.
The `NoSuchMethodError` indicates that the class is present but has different methods.
When you start Tachyon using their scripts, they can put the right dependencies into the classpath. When you start it within our tests, they run with our dependencies .. which do not seem to match.

You might be able to resolve the issue by using a newer version of log4j in our project.... But I'm not sure about this. Have a look at the transitive dependencies.

zentol:
you are correct, that was the problem. i added the newer version as a test dependency.
i ended up removing the call to startWebServer() completely, since there were other issues with it aswell.

stratosphere-core/src/main/java/eu/stratosphere/core/fs/FileSystem.java
```diff
@@ -153,6 +155,7 @@ public int hashCode() {
 		FSDIRECTORY.put("hdfs", DISTRIBUTED_FILESYSTEM_CLASS);
 		FSDIRECTORY.put("file", LOCAL_FILESYSTEM_CLASS);
 		FSDIRECTORY.put("s3", S3_FILESYSTEM_CLASS);
+                FSDIRECTORY.put("tachyon", TACHYON_FILESYSTEM_CLASS);
```
rmetzger, line 158: When you put Tachyon here, you also have to add it to `stratosphere-dist` as a dependency, because otherwise, it would not be deployed with the Stratosphere binary.

rmetzger:
Ok. Did you copy the test-code from Tachyon into our project?
I think that there is no way to resolve the issue right now because Tachyon does not ship the test code.
I asked on their mailing list if we can fix it for them.

zentol:
yes, i copied the class that starts tachyon. i also had to add another class (Master) due to scope issues.

rmetzger:
Tachyon accepted a pull request (https://github.com/amplab/tachyon/pull/136) I made. Once they push a new release, the test-jar will be available and we do not need to copy the code anymore.

fhueske:
Excellent! :+1:

stratosphere-addons/tachyon/src/main/java/eu/stratosphere/fs/tachyon/TachyonBlockLocation.java
```diff
@@ -0,0 +1,42 @@
...
+    public long getLength() {
+        return length;
+    }
+
+    /**
+     * Compares two BlockLocations based on their offset.
+     */
+    @Override
+    public int compareTo(BlockLocation o) {
+        return (new Long(offset)).compareTo(o.getOffset());
```
rmetzger: Why are you creating a new Long here? I think a cast should suffice

stratosphere-addons/tachyon/src/main/java/eu/stratosphere/fs/tachyon/TachyonFileSystem.java
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    /**
+     * Returns the current working directory. 
+     * This directory is always the root directory.
+     * @return working directory
+     */
+    @Override
+    public Path getWorkingDirectory() {
+        return new Path(pathToString(workingDirectory));
```
rmetzger: Why can't you just return `this.workingDirectory` ?

stratosphere-addons/tachyon/src/main/java/eu/stratosphere/fs/tachyon/TachyonFileSystem.java
```diff
@@ -0,0 +1,250 @@
...
+     * @return FileStatus for given file
+     * @throws IOException 
+     */
+    @Override
+    public FileStatus getFileStatus(Path path) throws IOException {
+        List<ClientFileInfo> tachyInfo = fileSystem.listStatus(pathToString(path.getParent()));
+        FileStatus stratInfo = null;
+        for (int x = 0; x < tachyInfo.size(); x++) {
+            if (tachyInfo.get(x).path.equals(pathToString(path))) {
+                stratInfo = new TachyonFileStatus(tachyInfo.get(x), name);
```
rmetzger: I think you can save some iterations by immediately returning. Are you sure that it only hits once?

stratosphere-addons/tachyon/src/main/java/eu/stratosphere/fs/tachyon/TachyonFileSystem.java
```diff
@@ -0,0 +1,250 @@
...
+     * @throws IOException 
+     */
+    @Override
+    public FSDataOutputStream create(Path path, boolean overwrite) throws IOException {
+        String filePath = pathToString(path);
+        int fileID;
+        if (fileSystem.exist(filePath)) {
+            if (overwrite) {
+                fileSystem.delete(filePath, true);
+                fileID = fileSystem.createFile(filePath);
```
rmetzger, line 238: I have the feeling that you copy-pasted code here.

rmetzger, line 238: Shit. sorry. Its not ;)

stratosphere-addons/tachyon/src/main/java/eu/stratosphere/fs/tachyon/TachyonFileSystem.java
zentol: good point, changed.

stratosphere-addons/tachyon/src/main/java/eu/stratosphere/fs/tachyon/TachyonFileSystem.java
zentol: i cant think of way for multiple hits, that would require files with the same path (which includes the filename).

rmetzger: Then its okay.

stratosphere-addons/tachyon/src/main/java/eu/stratosphere/fs/tachyon/TachyonBlockLocation.java
zentol: changed

rmetzger:
By the way: I found out why you/we were unable to access the tachyon web interface from outside the virtual machine. The address was bound to "localhost", therefore, you could only access it from the local machine. Changing it to the hostname of the machine allows you to externally access it (for example outside virtual box).

rmetzger:
Hey,

I think the default behavior of HDFS is to create the output directories if they do not exist.
@fhueske: You've recently changed the behavior there: How can @zentol implement this?

```
vagrant@master:~/stratosphere/stratosphere-dist/target/stratosphere-yarn-0.5-SNAPSHOT$ ./bin/stratosphere run -j ./examples/stratosphere-java-examples-0.5-SNAPSHOT-WordCount.jar -a 2 tachyon://master.local:19998/input/gpl.txt tachyon://master.local:19998/output/wc
Found a .yarn-jobmanager file, using "hadoop1.local:6123" to connect to the JobManager
ERROR: The job was not successfully submitted to the nephele job manager: java.lang.RuntimeException: I/O Error while accessing file
    at eu.stratosphere.pact.runtime.task.DataSinkTask.getMaximumNumberOfSubtasks(DataSinkTask.java:392)
    at eu.stratosphere.nephele.jobgraph.AbstractJobVertex.getMaximumNumberOfSubtasks(AbstractJobVertex.java:723)
    at eu.stratosphere.nephele.executiongraph.ExecutionGraph.createVertex(ExecutionGraph.java:514)
    at eu.stratosphere.nephele.executiongraph.ExecutionGraph.constructExecutionGraph(ExecutionGraph.java:273)
    at eu.stratosphere.nephele.executiongraph.ExecutionGraph.<init>(ExecutionGraph.java:174)
    at eu.stratosphere.nephele.jobmanager.JobManager.submitJob(JobManager.java:551)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:622)
    at eu.stratosphere.nephele.ipc.RPC$Server.call(RPC.java:416)
    at eu.stratosphere.nephele.ipc.Server$Handler.run(Server.java:944)
Caused by: java.io.IOException: FileDoesNotExistException(message:/output)
    at tachyon.master.MasterClient.listStatus(MasterClient.java:264)
    at tachyon.client.TachyonFS.listStatus(TachyonFS.java:806)
    at eu.stratosphere.fs.tachyon.TachyonFileSystem.getFileStatus(TachyonFileSystem.java:73)
    at eu.stratosphere.core.fs.FileSystem.exists(FileSystem.java:340)
    at eu.stratosphere.core.fs.FileSystem.initOutPathDistFS(FileSystem.java:541)
    at eu.stratosphere.pact.runtime.task.DataSinkTask.getMaximumNumberOfSubtasks(DataSinkTask.java:381)
    ... 11 more
Caused by: FileDoesNotExistException(message:/output)
    at tachyon.thrift.MasterService$liststatus_result$liststatus_resultStandardScheme.read(MasterService.java:6718)
    at tachyon.thrift.MasterService$liststatus_result$liststatus_resultStandardScheme.read(MasterService.java:6676)
    at tachyon.thrift.MasterService$liststatus_result.read(MasterService.java:6610)
    at org.apache.thrift.TServiceClient.receiveBase(TServiceClient.java:78)
    at tachyon.thrift.MasterService$Client.recv_liststatus(MasterService.java:335)
    at tachyon.thrift.MasterService$Client.liststatus(MasterService.java:322)
    at tachyon.master.MasterClient.listStatus(MasterClient.java:260)
    ... 16 more
```

rmetzger:
What did I do wrong?

I have a Tachyon cluster with 2 machines.

```
ERROR: The program execution failed: java.io.IOException: Error opening the Input Split tachyon://master.local:19998/input/gpl.txt [17574,17573]: The machine does not have any local worker.
    at eu.stratosphere.api.common.io.FileInputFormat.open(FileInputFormat.java:549)
    at eu.stratosphere.api.common.io.DelimitedInputFormat.open(DelimitedInputFormat.java:425)
    at eu.stratosphere.api.common.io.DelimitedInputFormat.open(DelimitedInputFormat.java:44)
    at eu.stratosphere.pact.runtime.task.DataSourceTask.invoke(DataSourceTask.java:134)
    at eu.stratosphere.nephele.execution.RuntimeEnvironment.run(RuntimeEnvironment.java:345)
    at java.lang.Thread.run(Thread.java:701)
Caused by: java.io.IOException: The machine does not have any local worker.
    at tachyon.client.BlockOutStream.<init>(BlockOutStream.java:75)
    at tachyon.client.RemoteBlockInStream.<init>(RemoteBlockInStream.java:65)
    at tachyon.client.BlockInStream.get(BlockInStream.java:48)
    at tachyon.client.TachyonFile.getInStream(TachyonFile.java:66)
    at eu.stratosphere.fs.tachyon.TachyonFileSystem.open(TachyonFileSystem.java:130)
    at eu.stratosphere.api.common.io.FileInputFormat$InputSplitOpenThread.run(FileInputFormat.java:687)
```

fhueske:
I did not change the behavior, I added an option to let the user decide.

But `TachyonFilesystem.getFileStatus(Path path)` is not correctly implemented. 
JavaDoc says:

```
/**
     * Return a file status object that represents the path.
     * 
     * @param f
     *        The path we want information from
     * @return a FileStatus object
     * @throws FileNotFoundException
     *         when the path does not exist;
     *         IOException see specific implementation
     */
```

It must throw a `FileNotFoundException` and not a `FileDoesNotExistException`.
Otherwise `Filesystem.exists()` will not work.

zentol:
i made a change so that a FileNotFoundException is thrown whenever an IOException is passed from tachyon. Lets see if that changes anything.

fhueske:
Not sure if that's the correct approach as it might swallow some serious IOExceptions.
Isn't it sufficient to throw a `FileNotFoundException` if a `FileDoesNotExistException` is caught?

zentol:
it wasn't supposed to solve the problem long term, just a quick change to see whether the behavior changes. but here's a better version :)

rmetzger:
Hey, I think you have to add our license headers to your files (we recently introduced a maven plugin to check for the headers).

Please let me know when you consider the pull request as "ready for testing" so that I can test it on my little Tachyon cluster.

StephanEwen:
Moved to stratosphere-artifacts

rmetzger:
See: https://github.com/stratosphere/stratosphere-experimental/pull/2

Issue #519, opened 2014-03-01
Add setValue methods to Value types


fhueske:
StringValue provides a function StringValue.setValue(StringValue s) to set its value to the value of another StringValue.

The other value types (e.g., IntValue, DoubleValue) do not provide such a method. 
For consistency reasons, these types should be extended by the corresponding functions.

MikhailErofeev:
Hello, I want to realize this as my first contribution to project. 
Would it be nice to add interface MutableValue<V> with setValue(V) and implement it to basic types? Should this interface contains more methods?

rmetzger:
Hi,
great idea! We are always happy about contributions!

I think introducing an interface `MutableValue<T> for a`setValue(T)` method is a good idea! Somebody might come up with a better name for the interface but renaming afterwards should not be a big problem ;)

MikhailErofeev:
can't link pull request with this issue:( https://github.com/stratosphere/stratosphere/pull/525
Why don't you use Null-checkers annotations in project? They are very helpful in code analysis with IDE, imo.

MikhailErofeev:
Looks like it' bad fix. It's broke compability with existing code in cases, when autoboxing falls (int -> Double, for example).

StephanEwen:
Fixed in 2006980be27debb604b40dc8f5c0ea229a4ae0ac via #525

PR #525, opened 2014-03-02
#519 Add setValue methods to Value types

MikhailErofeev:
add ResettableValue interface to base values and refactor setValue() in StringValue


MikhailErofeev:
oops, Travis gone red at 'stratosphere-tests', with strange error. 

"eu.stratosphere.client.program.ProgramInvocationException: The program execution failed: java.lang.NoSuchMethodError: eu.stratosphere.types.IntValue.setValue(I)V
    at eu.stratosphere.examples.scala.datamining.KMeansForTest$UDTSerializerImpl$2.serialize(KMeansForTest.scala:79)
    at eu.stratosphere.examples.scala.datamining.KMeansForTest$UDTSerializerImpl$2.serialize(KMeansForTest.scala)
    at eu.stratosphere.examples.scala.datamining.KMeansForTest$$anon$8.readRecord(KMeansForTest.scala:79)
    at eu.stratosphere.examples.scala.datamining.KMeansForTest$$anon$8.readRecord(KMeansForTest.scala:79)
    at eu.stratosphere.api.common.io.DelimitedInputFormat.nextRecord(DelimitedInputFormat.java:449)
    at eu.stratosphere.pact.runtime.task.DataSourceTask.invoke(DataSourceTask.java:151)
    at eu.stratosphere.nephele.execution.RuntimeEnvironment.run(RuntimeEnvironment.java:345)
    at java.lang.Thread.run(Thread.java:744)"

why KMeansForTest commented? what is the UDTSerializerImpl? where are in code invocation of  setValue() ?

fhueske:
Thanks a lot for your pull request and welcome!

The Scala API automatically generates data structures (List, Records, etc.) based on the basic data types (IntValue, StringValue, ...). The UDTSerializerImpl is auto-generated code to read and write one of these data structures if I am not mistaken.

I think we should keep the setValue methods for the primitive types (`setValue(int i)`, `setValue(double d)`, ...) as well for performance reasons. Otherwise, one would need to create (or let autoboxing create) an object just to update a value. Since these methods tend to be called very often, this could increase the pressure on garbage collection which we try to avoid as much as possible. Adding the methods for the basic types should fix the issue with Scala data type deserializer.

Also, the issue #519 did mean to add methods like `setValue(IntValue i)` and `setValue(DoubleVal)` such that the value of an IntValue can be set using another IntValue.

Best, Fabian

StephanEwen:
The error comes from a special test whose classes are in a jar file under
/src/test/resources. We have them not in the normal compile/test scope to
verify the behavior of classloading with user-defined classes; that is why
the test is commented. If you change fundamental signatures, you need to
rebuild that test.

MikhailErofeev:
oops, i misunderstand issue, and implement it to java primitives wrappers instead of Value. fix it

fhueske:
OK, thanks a lot! :-)

stratosphere-core/src/main/java/eu/stratosphere/types/StringValue.java
```diff
@@ -137,17 +138,7 @@ public String getValue() {
 	 * @param value The new string value.
 	 */
 	public void setValue(CharSequence value) {
-		if (value == null)
```
StephanEwen, line 140: I would suggest to keep this check, it gives a more specific error message.

StephanEwen:
Looks good. One question though: You removed a few checks for null. The old checks gave a specific error message. When now someone submits a null value, the `value.value` statements will all throw an unspecific null pointer exception.

MikhailErofeev:
ok, I replace this with apache Validate, forgot about old checking, then thought that exception will occurs few lines below and remove checking. Should i add validators in other setters?

StephanEwen:
I think it looks good. I will merge it.

StephanEwen:
Merged in 2006980be27debb604b40dc8f5c0ea229a4ae0ac

Thanks!

Issue #516, opened 2014-02-28
Detect Pipelined Dataflows that might lead to Deadlocks


StephanEwen:
Certain pipelined flows may lead to deadlocks, in which case we need to make sure the pipelines are broken or made elastic enough to prevent that.

This is only relevat to pipelined data flows where one operator has more than one consumers (successors in the flow).

Most cases are caught by the general logic that deals with branching/joining flows. The following cases need additional checks:

```
                    <build>
(source1) ------ (join)
          \    /    <probe>
           \  /
            \/
            /\
           /  \
          /    \    <probe>
(source2) ------(join)
                    <build>
```

Since both sources pipeline their data into a build and a probe side, they get stalled by  the back pressure from the probe side (which waits for the build side to complete) and never finish the build side.

We can model dependencies of pipelined / materialized connections and do a reguar cyclic dependencies check to detect such situations. Pipelined connections have a dependency from sender to receiver, non-pipelined (fully dammed) connections have a dependency from receiver to sender.

A [PlanNode](https://github.com/stratosphere/stratosphere/blob/master/stratosphere-compiler/src/main/java/eu/stratosphere/compiler/plan/PlanNode.java) and a [Channel](https://github.com/stratosphere/stratosphere/blob/master/stratosphere-compiler/src/main/java/eu/stratosphere/compiler/plan/Channel.java) are considered pipelined unless the following holds:
- The channel has a [LocalStrategy](https://github.com/stratosphere/stratosphere/blob/master/stratosphere-runtime/src/main/java/eu/stratosphere/pact/runtime/task/util/LocalStrategy.java) where `LocalStrategy#dams()`returns true.
- The channel has a [TempMode](https://github.com/stratosphere/stratosphere/blob/master/stratosphere-compiler/src/main/java/eu/stratosphere/compiler/dag/TempMode.java) where `TempMode#breaksPipeline()` is true.
- The PlanNode has a [DriverStrategy](https://github.com/stratosphere/stratosphere/blob/master/stratosphere-runtime/src/main/java/eu/stratosphere/pact/runtime/task/DriverStrategy.java) where the [DamBehavior](https://github.com/stratosphere/stratosphere/blob/master/stratosphere-runtime/src/main/java/eu/stratosphere/pact/runtime/task/DamBehavior.java)  as returned by `firstDam()` or `secondDam()` for the respective input is a `FULL_DAM`.

[...]

PR #527, opened 2014-03-03
Compiler deadlock detection

markus-h:
I implemented a deadlock detection inside the compiler. For details see #516. Basically I build a dependency graph and check for cycles. If there is a cycle I try to fix the deadlock by first trying to switch build and probe side of hash joins. If that does not work I insert pipeline breakers and force materialization.

I am not 100% sure that all cases of deadlocks in combinations with iterations are detected. Especially I couldn't take care of deadlocks that occure in combination with the termination criterion of the BulkIteration, since it is not yet part of the master.

### DIFF
--- a/stratosphere-compiler/src/main/java/eu/stratosphere/compiler/PactCompiler.java
+++ b/stratosphere-compiler/src/main/java/eu/stratosphere/compiler/PactCompiler.java
@@ -30,13 +30,13 @@ import org.apache.commons.logging.LogFactory;
 
 import eu.stratosphere.api.common.Plan;
 import eu.stratosphere.api.common.operators.BulkIteration;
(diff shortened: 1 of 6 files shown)


StephanEwen:
Looks good to me! I have two minor suggestions:
1. Can you rename the subpackage `deadlock` to `deadlockdetect` ?
2. Can you make the deadlock detector its own class? I know the other steps are realized as internal classes, but I would suggest to go away from that pattern a bit. The compiler class is too huge already.

markus-h:
I integrated your proposed changes.
Moreover I tested the detection with our big testjob and it is working. The manual compiler hint that was needed before ist now superfluous.

StephanEwen:
That is amazing :-)

StephanEwen:
Merged in 32b218535f0bd4ebc99126a092bdea53ad26fee1

Issue #504, opened 2014-02-25
Bleeding menu on Downloads and Documentation web pages


nullusDefectus:
This is probably low priority, but I notice that the menu in the left sidebar on the Downloads and the Documentation page bleeds on to content when the page is scrolled down.

The page is responsive for the most part, except when I dock the page left/right on the screen. Notice that the menu isn't collapsed to the right-hand top corner of the page as it should.

My monitor is 24" set to 1920x1080 (native) resolution. I am running Chrome "33.0.1750.117 m" on Windows 8.1. 

![image](https://f.cloud.github.com/assets/6477490/2264720/00640024-9e76-11e3-945d-3e0f2c1eb363.png)

uce:
Thank you for posting this. I will look into it after I'm back from holidays, if nobody else finds the time.

I think the left sub navigation needs a proper fix, where it collapses earlier and for the top navigation we should think about moving some links to a dropdown, for example Events and Contact to the existing Project dropdown?

rmetzger:
Thanks for reporting the issue. I fixed it in commit: 3cf916d3b1083177fbd22474de4263e76df3106e

nullusDefectus:
Thanks @rmetzger.

I notice a minor problem with the fix. When the screen is narrow and the left menu is collapsed, everything looks fine, but not when the left menu is expanded, about a half of the first entry in the left menu is hidden. I believe the difference is the changing height of the top menu, illustrated below with four different screen (progressively smaller) widths.

## Screen wide enough for full top-menu

![image](https://f.cloud.github.com/assets/6477490/2304108/6cd062a2-a204-11e3-8eb6-c048cd937acb.png)

## Narrower screen, but top menu not yet collapsed

![image](https://f.cloud.github.com/assets/6477490/2304109/83dd3830-a204-11e3-8785-5076a3cb1186.png)

## Even narrower screen (~20%?), top menu not yet collapsed

![image](https://f.cloud.github.com/assets/6477490/2304113/9bc2f0a2-a204-11e3-9188-fcba43d006b4.png)

## Screen narrow enough for top menu to collapse

![image](https://f.cloud.github.com/assets/6477490/2304114/bb56d9a6-a204-11e3-9db7-246f54dc63e4.png)

rmetzger:
Hi, thanks for your feedback.
I'll ask somebody to fix the remaining issues with the website.

nullusDefectus:
[...]

PR #530, opened 2014-03-03
Fixed Issue #504

### LINKED ISSUE #5: Configurable Iteration Aggregators
Aggregators are currently hard wired into the bulk iteration runtime code. Make them configurable on the Bulk-Iteration plan construct.

Aggregators must be hosted by the iteration head task, communicated to the convergence criterion and communicated back, such that the tasks can access the previous superstep's aggregation result.

It is open to discussion whether to keep the entire history of aggregation results, only the previous iteration's aggregate result, or whether this should be made configurable.

MoeweX:
Hi,

I managed to find a way to fix this issue my increase the size where the navbar starts to collapse. This is needed because we have to many items in our top-navigation bar ... . .
bildschirmfoto 2014-03-03 um 19 01 56

fixes #5

### DIFF
--- a/css/main.css
+++ b/css/main.css
@@ -433,6 +433,33 @@ section {
     font-weight: bold;
 }*/
 
+@media (max-width: 991px) {
+    .navbar-header {
+        float: none;
+    }
+    .navbar-toggle {
+        display: block;
+    }
+    .navbar-collapse {
+        border-top: 1px solid transparent;
+        box-shadow: inset 0 1px 0 rgba(255,255,255,0.1);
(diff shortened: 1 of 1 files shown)


rmetzger:
Thank you for your pull request.
There is a new issue with a width between 880 - 991 px (inclusive)

bug

rmetzger:
Please reopen the pull request against the new website repository (https://github.com/stratosphere/stratosphere.github.io) See also https://github.com/stratosphere/stratosphere/issues/487

Issue #455, opened 2014-02-04
Implement HadoopDataSink into hadoop-compatibility package


rmetzger:
The `HadoopDataSink` is similar to the `HadoopDataSource`. 
See here for an example: https://github.com/stratosphere/stratosphere/blob/master/stratosphere-addons/hadoop-compatibility/src/main/java/eu/stratosphere/hadoopcompat/example/WordCount.java#L117 and the issue: https://github.com/stratosphere/stratosphere/issues/382

We also need a  `HadoopDataSink` that supports any kind of Hadoop-OutputFormat for Stratosphere.

API Example:

``` java
HadoopDataSink out = new HadoopDataSink(new SequenceFileOutputFormat<LongWritable, Text>(), "Write SequenceFile");
SequenceFileOutputFormat.setOutputPath(out.getJobConf(), new Path(output));
```

rmetzger:
Hey @qmlmoon, have you started working on this task? Is the work in progress somewhere in GitHub? Somebody on the mailing list was asking for this: https://groups.google.com/d/msg/stratosphere-dev/qYvJRSoMYWQ/zPf_nGHp_M0J

qmlmoon:
I push it to my Github. HadoopDataSink works in local mode for hadoop 1.2.1. Still need to delete temporary directory generated by hadoop and test cluster mode.

PR #531, opened 2014-03-04
HadoopDataSink works both for hadoop 1.2.1 and 2.2.0

qmlmoon:
issue: #455
Need to delete temporary output folder when job finishes.


rmetzger:
Thank you for your pull request. I'll take a look as soon as I find time.

Could you add a test case that tests both the Hadoop Input / Output format?

rmetzger:
Have you noticed the failed builds? https://travis-ci.org/stratosphere/stratosphere/builds/20147138
Log file: https://s3.amazonaws.com/archive.travis-ci.org/jobs/20147139/log.txt

stratosphere-addons/hadoop-compatibility/src/main/java/eu/stratosphere/hadoopcompatibility/HadoopOutputFormatWrapper.java
```diff
@@ -0,0 +1,139 @@
...
+
+	public FileOutputCommitterWrapper fileOutputCommitterWrapper;
+
+	public HadoopOutputFormatWrapper(org.apache.hadoop.mapred.OutputFormat<K,V> hadoopFormat, JobConf job, StratosphereTypeConverter<K,V> conv) {
+		super();
+		this.hadoopOutputFormat = hadoopFormat;
+		this.hadoopOutputFormatName = hadoopFormat.getClass().getName();
+		this.converter = conv;
+		// merge hadoopConf into jobConf. This is necessary for the hdfs configuration
+		org.apache.hadoop.conf.Configuration hadoopConf = DistributedFileSystem.getHadoopConfiguration();
```
rmetzger: I think this code is copy-pasted from the HadoopInputFormatWrapper.
Can you move it to method used by both Wrappers?

rmetzger:
You said `Need to delete temporary output folder when job finishes.`.

Isn't it possible to move the finished parts in the close() method to the final destination (HDFS) ? Once the job has finished, all files should be gone in the tmp dir?

stratosphere-addons/hadoop-compatibility/src/main/java/eu/stratosphere/hadoopcompatibility/example/WordCountWithHadoopOutputFormat.java
```diff
@@ -0,0 +1,158 @@
...
+		}
+	}
+
+	/**
+	 * Sums up the counts for a certain given key. The counts are assumed to be at position <code>1</code>
+	 * in the record. The other fields are not modified.
+	 */
+	@Combinable
+	@ConstantFields(0)
+	public static class CountWords extends ReduceFunction implements Serializable {
```
rmetzger, line 82: Isn't it possible to re-use the Mapper and Reducer from the regular WordCount example?

rmetzger, line 82: Ah, no, probably not because of the dependencies

qmlmoon:
The problem is the tmp dir itself cannot be deleted in close() if still
some tasks are running.

2014-03-06 12:55 GMT+01:00 Robert Metzger notifications@github.com:

qmlmoon:
Ah. The failed builds is due to the dependency. That's also something i
want to ask:). There's no hadoop-compatibility dependency under
sratosphere-test package, should I put the test file also in
hadoop-compatibility package or?

2014-03-06 12:49 GMT+01:00 Robert Metzger notifications@github.com:

rmetzger:
Hi,
yes, put the tests into the "hadoop-compatibility" package.
For spargel, we added the tests like this: (https://github.com/stratosphere/stratosphere/blob/master/stratosphere-addons/spargel/pom.xml)

``` xml
        <dependency>
            <artifactId>stratosphere-tests</artifactId>
            <scope>test</scope>
            <type>test-jar</type>
        </dependency>
```

StephanEwen:
@rmetzger Is this good to merge?

rmetzger:
I want to test it first on the cluster. I try to do that tomorrow.

stratosphere-addons/hadoop-compatibility/src/test/java/eu/stratosphere/test/hadoop-compatibility/HadoopInputOutputTest.java
```diff
@@ -0,0 +1,35 @@
+package eu.stratosphere.test.hadoopcompatibility;
```
rmetzger: The directory name (`hadoop-compatibility`) is not the same as the package name `hadoop-compatibility`.

rmetzger:
I started working on this. I have to fix a few things since the code does not check the user input and is causing null pointer exceptions at runtime (even in local mode).
I will work on this on Monday and then push the change.

rmetzger:
I merged this pull request in cc03f84450169bbcb36630569f4e70258e60bb51.

I also added code to check if the user has properly set the output directory (d61d87f3df85eefe4848c3fd97ba9fb2495d748e) so that we throw the exception at compile time (and not at runtime).

Issue #463, opened 2014-02-08
Implement new Java API for Stratosphere


rmetzger:
The Stratosphere team is going to implement a new Java API. We will keep the old java API, calling it the "java-record-api".

This issue is for planning and discussion of the new API. I will update it if new requirements come in.

@StephanEwen is going to lead the implementation effort. (Current working branch https://github.com/StephanEwen/stratosphere/tree/japi) (See the examples https://github.com/StephanEwen/stratosphere/blob/japi/stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/wordcount/WordCount.java)

**Requirements**:
- Support for any java object as types.
- a typesafe tuple-model.
- Predefined aggregations for sum, min, max etc.
- Key-extraction using an extractor function.
- A more fluent program description that feels native to java developers (a DataSet model, rather than a dataflow-model)
- reduce-function that returns the reduce-key

We are planning to implement more features. These requirements are just for the first version.

**Timeline**:
- New Java API is the main new feature of Release 0.5
- First working prototype: week of 17. Feb.
- 0.5 Release Candidate 1: 7. March 2014.

sscdotopen:
Why not throw away the old API? Keeping it only creates legacy that needs
to be maintained.
Am 08.02.2014 16:45 schrieb "Robert Metzger" notifications@github.com:

rmetzger:
I think we are going to throw it away once the new API is stable and adopted. Since the new API is in a separate package, we can change the quickstarts and the documented maven dependencies so that new users don't see the old one. 

But I think we have to keep it for a few months. The new API will probably also be a bit slower since it takes us a while to implement sophisticated object serialization with codegen etc. Therefore, "high performance" users can choose the record-api as an option.
We just need to make sure that we do not end up in the `mapred`, `mapreduce`-hell ;)

UPDATE: I deleted the other part of the message again. I should think before posting ;)

sscdotopen:
If highend users choose the old api, you _cannot_ throw it away
afterwards...
Am 08.02.2014 16:59 schrieb "Robert Metzger" notifications@github.com:

StephanEwen:
I think there is not really a problem here for the following reasons:
1. I expect that the new api will not be slower than the record api, if
   used in a similar way, i.e. a tuple-like fashion with our own value types.
   On the contrast, I expect it to be faster even. In that sense there should
   be no reason to stick with the record api, at least not performance wise.
[...]

PR #535, opened 2014-03-06
Clean up LICENSE and NOTICE files

rmetzger:
This commit is from @uce.
1. Added LICENSE and NOTICE files to the source tree
   => we now have two files, one for the source, one for the dist
2. Renamed LICENSE.txt and NOTICE.txt without the .txt 
   => This seems to be the common naming.
3. Removed swt-visualization project, because it used LGPL 2.1 dependencies and nobody really uses it. (We do not support it)
4. Removed the notices about the usage of other Apache licensed SW, because I am not sure if we need it


rmetzger:
I think we should keep the contents of the old NOTICE file.

StephanEwen:
The old Notice file had an incomplete list of license dependencies. Ufuk's new LICENSE file has a complete list.

I'll add a note in the NOTICE file that a list of dependencies with their licenses is in the LICENSE file

StephanEwen:
Merged in 551e83f8b04be549d5750477da814370e2def541 and 2d0975b028dc457f9b1f36bce6cb538a7d484967

rmetzger:
Why did you merge it into `staging` ?
I though the branch is only for the new JavaAPI ?
Actually, `master` is kind of a staging area for releases. If we have the feeling that too many users are using `master` we should release a new version.

fhueske:
The idea of the staging branch was to preserve the stability of the master branch, mainly because of the complex PR #545. @StephanEwen asked to do all PRs and merges against the staging branch to avoid merge conflicts when staging goes into master.
Once staging is merged with master, we will again work directly on master.

rmetzger:
Oh, I got that wrong. I merged some PR's into `master` recently.

What's the problem with rebasing `staging` before merging it to master ?

I'm against having a `staging` branch. It makes our development process too complicated.
New features should be developed "somewhere" until they are stable enough, then, they should come into the mainline.
If the new Java API is not stable enough for `master` we should have kept it in @StephanEwen's repo  and make the pull requests there.

fhueske:
The problem with rebasing staging is the hugh changelist by PR #545.
As I said, we will continue to work on master once staging is merged.

I think it's OK to have unstable new features in master but we should be careful with things that people are currently relying on (therefore the staging branch). 
What would be the difference of having a branch in @StephanEwen's repository (to which only he can commit) and a temp branch in the @stratosphere repository?

StephanEwen:
It's only to avoid merge conflicts. We should merge staging tomorrow, if
Tobias' Tests work. Staging will then have existed for a week only.

Not all people were aware of the separate branch in my repo.

rmetzger:
> What would be the difference of having a branch in @StephanEwen's repository (to which only he can commit)

That we follow a development process that most of the other open source projects follow. (Something like a design pattern)
The main repo only contains code that is considered stable and is reviewed.
Adding contributors to a repository is really easy ;) 
We could also create a separate repository in @stratosphere for the new Java API.

> We should merge staging tomorrow, if Tobias' Tests work.

Who is Tobias?

> Not all people were aware of the separate branch in my repo.

I understand. But there is an issue (https://github.com/stratosphere/stratosphere/issues/463) for planning the new Java API and you could have used that to promote your branch and find contributors. (Sorry, I don't want to annoy you. It is about reflecting our development process)

PR #540, opened 2014-03-07
Reuse of already generated OptimizerNodes in the recursive descent step of GraphCreatingVisitor

tillrohrmann:
I tried to tackle the problem of generating multiple OptimizerNodes for one and the same Operator during recursive descent steps of the GraphCreatingVisitor as they happen for the step function generation of iterations. This poses a problem, because flows with multiple sinks, where one of the inputs to the step function is outputted as well, will be wrongly recognised as an unconnected data flow. Moreover, in the case of an Operator which produces random results, the generated data of this Operator inside and outside of the iteration would be distinct and thus false.

My solution is to forward the generated OptimizerNodes to the recursive GraphCreatingVisitor. This implies, however, to consider the step function for the computation of the unclosed branch stack. Furthermore, the branch plans of the step function has to be taken account of as well when merging the branch plans of the iteration plan nodes. For the WorksetIterationPlanNode, I could only come up with a nasty hack to disable the mergeBranchPlanMaps of the DualInputPlanNode.

So far, all test cases succeed locally but I'm not sure whether I found all code positions requiring a change for this modification. Would be great if you could review it to make sure that I didn't forget anything.

### DIFF
--- a/stratosphere-compiler/src/main/java/eu/stratosphere/compiler/PactCompiler.java
+++ b/stratosphere-compiler/src/main/java/eu/stratosphere/compiler/PactCompiler.java
@@ -787,11 +787,16 @@ public class PactCompiler {
 
 		
 		private GraphCreatingVisitor(int maxMachines, int defaultParallelism) {
-			this(null, false, maxMachines, defaultParallelism);
+			this(null, false, maxMachines, defaultParallelism, null);
 		}
 		
-		private GraphCreatingVisitor(GraphCreatingVisitor parent, boolean forceDOP, int maxMachines, int defaultParallelism) {
-			this.con2node = new HashMap<Operator, OptimizerNode>();
+		private GraphCreatingVisitor(GraphCreatingVisitor parent, boolean forceDOP, int maxMachines,
+                                     int defaultParallelism, HashMap<Operator, OptimizerNode> closure) {
+            if(closure == null){
+                con2node = new HashMap<Operator, OptimizerNode>();
+            }else{
+                con2node = closure;
+            }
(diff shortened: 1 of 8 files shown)


rmetzger:
Hi @tillrohrmann,
thank you for your contribution.
We'll have a look at your pull request as soon as we find time.

StephanEwen:
Thank you, the pull request looks good.

There are two things where I am wondering about implications:
- The logic that deals with branching and rejoining data flows may be affected by this. To verify that, we would need as test with a data flow as shown further below.
- The visualization of iteration plans might be broken for such plans. This issue is less critical, as we are going to replace the visualization library anyways in the near future.

```
                 +----Iteration------+
                 |                   |
        /--map--< >----\             |
       /         |      \         /-< >---sink
src-map          |     join------/   |
       \         |      /            |
        \        +-----/-------------+
         \            /
          \--reduce--/
```

tillrohrmann:
I added the test case for the bulk iteration and the delta iteration. They pass on my computer.

StephanEwen:
That is good news :-) It think it is good to merge then. Can you include
the test in your pull request? You only need to update the branch from
which you created the pull request.

tillrohrmann:
They should be included. I appended them to the class BranchingPlansCompilerTest.

StephanEwen:
Very nice work.

The only comment I have is to not use wildcard imports. We try to keep them out of the core code to prevent name clashes that may be overlooked.

Merging it now and testing...

StephanEwen:
Merged in f05d24ad61dcda29affe67bacd61926a657bb77c

PR #542, opened 2014-03-07
Bug fixes for and extensions of new Java API

fhueske:
- Fix for Map operator translation (introduced by merge of 0038c9d0773e90676)
- Fix for `TupleComparator` (normalized key writing) @stephanewen please verify
- Fix for `BasicTypeComparator` (inverted sort order) @StephanEwen please verify
- initial support for Group Sorting (position keys and Reduce operator only)
- extended `TypeExtractor` and `TupleTypeInfo` to work with subclasses of Tuple classes.
- added join function for semi joins. Support in runtime required to enable this feature.

PR #544, opened 2014-03-07
Fix bug in MapNode

aljoscha:
MapNode is responsible for PlainMapOperatorBase and not for
MapOperatorBase, but this is a mess right now.

Maybe we should think about cleaning the situation a bit. CollectorMap is the same as FlatMap, only the function in the interface is called map instead of flatMap. I know that this is because of historic reasons but it is still a mess because we handle all three types through the layers. For example, some code in PactCompiler looks like this:

``` java
else if (c instanceof MapOperatorBase) {
    n = new CollectorMapNode((MapOperatorBase<?>) c);
}
else if (c instanceof PlainMapOperatorBase) {
    n = new MapNode((PlainMapOperatorBase<?>) c);
}
else if (c instanceof FlatMapOperatorBase) {
    n = new FlatMapNode((FlatMapOperatorBase<?>) c);
}
```

### DIFF
--- a/stratosphere-compiler/src/main/java/eu/stratosphere/compiler/dag/MapNode.java
+++ b/stratosphere-compiler/src/main/java/eu/stratosphere/compiler/dag/MapNode.java
@@ -37,8 +37,8 @@ public class MapNode extends SingleInputNode {
 	}
 
 	@Override
-	public MapOperatorBase<?> getPactContract() {
-		return (MapOperatorBase<?>) super.getPactContract();
+	public PlainMapOperatorBase<?> getPactContract() {
+		return (PlainMapOperatorBase<?>) super.getPactContract();
 	}
 
 	@Override


fhueske:
I had a fix for this in #542 as well ;-)

stratosphere-compiler/src/main/java/eu/stratosphere/compiler/dag/MapNode.java
```diff
@@ -37,8 +37,8 @@ public MapNode(PlainMapOperatorBase<?> pactContract) {
 	}
 
 	@Override
-	public MapOperatorBase<?> getPactContract() {
-		return (MapOperatorBase<?>) super.getPactContract();
+	public PlainMapOperatorBase<?> getPactContract() {
+		return (PlainMapOperatorBase<?>) super.getPactContract();
```
fhueske, line 41: Correct me if I am wrong, but won't this break with the Record API where Map extends MapOperatorBase?

rmetzger:
Damn. 
I suggest to have faster merge cycles for JAPI related (notice the new label) pull requests. As there are more and more developers working on it, we should sort out bugs as fast as possible.

Can we agree to a 
2 x +1 (from a committer or @markus-h)  AND no complaints ==> merge
voting scheme ?

stratosphere-compiler/src/main/java/eu/stratosphere/compiler/dag/MapNode.java
aljoscha, line 41: Yes, that's what I meant by historic reasons. We could change FlatMap and call the method there map() as well...

fhueske:
I agree if the two +1 do not include the PR author.

rmetzger:
Okay. Makes sense.

stratosphere-compiler/src/main/java/eu/stratosphere/compiler/dag/MapNode.java
fhueske, line 41: My fix removed the method as it was done before the merge and the build passed Travis.

aljoscha, line 41: This works as well, because MapOperatorBase is converted to a CollectorMapNode, not this MapNode here.

fhueske:
I agree with cleaning up the situation. It also took me quite some time to figure out how the different MapOperators relate to each other...

stratosphere-compiler/src/main/java/eu/stratosphere/compiler/dag/MapNode.java
fhueske, line 41: Alright then ;-)

StephanEwen, line 41: I agree that this needs urgent cleanout. I'll try and look over the PRs
soon...

fhueske:
The bug was fixed by another commit.
I'll open a new issue for the clean-up discussion and close this PR.

Issue #457, opened 2014-02-05
Fix tests failing with java 8


rmetzger:
Oracle is planning to announce Java 8 in March this year.
Stratosphere does currently not work with Java 8.

Someone should look into the issues and fix them until then. We should be prepared for this event if early adopters are using the new Java on their deskop systems for development.

zentol:
I'll look into this.

rmetzger:
Amazing!

zentol:
i changed a few things and it compiles now on linux (at least for me),
would be nice if someone else could try this:

https://github.com/zentol/stratosphere/tree/java8

rmetzger:
Let travis do the job.
You can add java8 to the `.travis.yml` file (See http://blog.travis-ci.com/2013-11-26-test-your-java-libraries-on-java-8/).
Then, open a pull request and it will test everything

PR #549, opened 2014-03-09
Fix tests failing with java 8 #457 

zentol:
Compatibility fixes for java 8, see #457.

Note: preliminary version, this is mostly to see whether i found all issuess

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,8 @@
 
 language: java
 jdk:
+  - oraclejdk8
   - oraclejdk7
-  - openjdk7
   - openjdk6
 env:
     global: 
--- a/pom.xml
+++ b/pom.xml
@@ -234,6 +234,36 @@
 				</pluginManagement>
 			</build>
 		</profile>
+                <profile>
+                        <id>jdk8</id>
+                        <activation>
+                                <activeByDefault>false</activeByDefault>
+                                <jdk>1.8</jdk>
+                        </activation>
+                <build>
+                        <plugins>
+                                <plugin>
+                                        <groupId>org.apache.maven.plugins</groupId>
+                                        <artifactId>maven-javadoc-plugin</artifactId>
+                                        <version>2.9.1</version>
+                                        <configuration>
+                                                <quiet>true</quiet>
+                                        </configuration>
+                                        <executions>
+                                                <execution>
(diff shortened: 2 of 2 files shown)


zentol:
looks like compiling on java8 works fine.

only issue remaining is that using -Xdoclint:none breaks previous versions of java.
(it prevent errors from malformed javadoc comments)

rmetzger:
Mh. Another user on Stackoverflow has the same issue we do: http://stackoverflow.com/a/16743137/568695

I have no idea what to do in this case.

zentol:
looks like it works. :)

rmetzger:
How did you resolve it?

StephanEwen:
We have a regular javadoc build plugin now and one conditionally for java 8 that overrides the regular one?

zentol:
@StephanEwen correct.

StephanEwen:
God job :-)

Out of curiosity, what is the javassist dependency for?

zentol:
there's a transitive dependency in stratosphere-core for a javaassist version that is not compatible with java 8. guess i could encapsulate it as well in an extra profile.

StephanEwen:
Okay. Can you solve it with the dependency management entries? I think that would be the "clean" way to do it.

Is that dependency actually used? Or is it a dependency that for example hadoop adds that is never used, because we only use hadoop's hdfs client?

zentol:
Ill add it to the dependency management.

i checked the dependendies again; javaassist actually comes up in multiple parts.
(clients, runtime and addons aswell)

maven marks it as used in clients and runtime.

rmetzger:
I think the pull request is good to merge.

StephanEwen:
Should we exclude one of the Java7 profiles? With Hadoop 1 and 2, we now have 8 builds to test. How about we test
- oraclejdk8
- oraclejdk7
- openjdk6

fhueske:
Agree :+1:

rmetzger:
@zentol Can you update the `.travis.yml` and remove the `openjdk7` ?
Then I'll merge your PR.

zentol:
@rmetzger Done.

rmetzger:
Merged in 6bf7d5a2451f9fd0b1773a51785b24f7aeb83304

PR #550, opened 2014-03-10
includeFields support for new Java API's CsvReader

twalthr:

PR #552, opened 2014-03-10
Relational example for new Java API (TPC-H Query 10)

twalthr:


rmetzger:
Mh. Why is only the hadoop2 build is running out of time?

twalthr:
sorry, I forgot to mention that #550 is required for the example to work properly

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/relational/TPCHQuery10.java
```diff
@@ -0,0 +1,206 @@
...
+				.includeFields("1000011010000000").types(Integer.class, Double.class, Double.class, String.class);
+
+		// read in nation table file
+		// nation: nationkey, name
+		DataSet<Tuple2<Integer, String>> nations = env.readCsvFile(nationPath).fieldDelimiter('|').includeFields("1100")
+				.types(Integer.class, String.class);
+
+		// orders filtered by year: orderkey, custkey
+		DataSet<Tuple2<Integer, Integer>> ordersFilteredByYear = orders
+				.flatMap(new FlatMapFunction<Tuple3<Integer, Integer, String>, Tuple2<Integer, Integer>>() {
```
fhueske: Can you use the built-in filter() operator for filters to demonstrate their use?
Also the code will be easier and filter() will be tested.

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/relational/TPCHQuery10.java
```diff
@@ -0,0 +1,206 @@
...
+						int year = Integer.parseInt(t.T3().substring(0, 4));
+						if (year > YEAR_FILTER) {
+							out.collect(new Tuple2<Integer, Integer>(t.T1(), t.T2()));
+						}
+					}
+				});
+
+		// lineitems filtered by flag: orderkey, extendedprice, discount
+		DataSet<Tuple3<Integer, Double, Double>> lineitemsFilteredByFlag = lineitems
+				.flatMap(new FlatMapFunction<Tuple4<Integer, Double, Double, String>, Tuple3<Integer, Double, Double>>() {
```
fhueske: Another filter.

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/relational/TPCHQuery10.java
```diff
@@ -0,0 +1,206 @@
...
+					@Override
+					public Tuple3<Integer, Double, Double> join(Tuple2<Integer, Integer> o, Tuple3<Integer, Double, Double> l) {
+						return new Tuple3<Integer, Double, Double>(o.T2(), l.T2(), l.T3());
+					}
+				});
+
+		// aggregate for revenue
+		// custkey, revenue
+		DataSet<Tuple2<Integer, Double>> revenueOfCustomerKey = lineitemsOfCustomerKey.groupBy(0).reduceGroup(
+				new GroupReduceFunction<Tuple3<Integer, Double, Double>, Tuple2<Integer, Double>>() {
```
fhueske: Can you rewrite this into a map that does the projection, i.e., computes revenue \* (1-discount) and a reduce function (not a groupReduce). 
Reduce should be preferred over GroupReduce because it can be more efficiently executed in many situations  (streaming hash-based instead of blocking sort-based) and implicitly allows for Combining.

fhueske:
#550 has been merged into staging. 
I think, the PR is ready for merge with staging as well once the changes have been done.

Thanks!

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/relational/TPCHQuery10.java
```diff
@@ -0,0 +1,206 @@
...
+		DataSet<Tuple2<Integer, Integer>> ordersFilteredByYear = orders
+				.flatMap(new FlatMapFunction<Tuple3<Integer, Integer, String>, Tuple2<Integer, Integer>>() {
+					private static final long serialVersionUID = 1L;
+					private static final int YEAR_FILTER = 1990;
+
+					@Override
+					public void flatMap(Tuple3<Integer, Integer, String> t, Collector<Tuple2<Integer, Integer>> out) throws Exception {
+						int year = Integer.parseInt(t.T3().substring(0, 4));
+						if (year > YEAR_FILTER) {
+							out.collect(new Tuple2<Integer, Integer>(t.T1(), t.T2()));
```
fhueske: Filter does not modify tuples. This can be done using the map() function (not flatMap). 
map() returns exactly one tuple for each input tuple.

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/relational/TPCHQuery10.java
```diff
@@ -0,0 +1,206 @@
...
+			throw new IllegalArgumentException("Invalid number of parameters: [customer.tbl] [orders.tbl] [lineitem.tbl] [nation.tbl]");
+		} else {
+			customerPath = args[0];
+			ordersPath = args[1];
+			lineitemPath = args[2];
+			nationPath = args[3];
+		}
+
+		final ExecutionEnvironment env = ExecutionEnvironment.getExecutionEnvironment();
+		env.setDegreeOfParallelism(1);
```
uce: Independent of this example: should we add a cons and getter, which directly sets the DOP?

StephanEwen: The LocalEnvironment and the RemoteEnvironment have that. The "
ExecutionEnvironment.getExecutionEnvironment()" has no such thing yet. If
you do not set anything, the DOP will be what is provided by the
environment.

I think DOP should actually not be configured inside a program by default,
this is a bit messy in many cases. It should come from the shell-script (as
a parameter) or the environment or so.

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/relational/TPCHQuery10.java
twalthr: I will use the new `aggregate()` function instead of a reduce to demonstrate its use. OK?

twalthr:
I have implemented your mentioned comments and tested the job with the staging branch, unfortunately, the TypeExtractor is buggy. I will try to fix it myself on monday.

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/relational/TPCHQuery10.java
fhueske: Right, `aggregate()` is even better, but not implemented yet.
Do you want to do this?
I opened an issue for this #581.

fhueske:
Can you also add some tests for the type extractor?
See #579.

StephanEwen:
Can you also remove the setDegreeOfParallelism?

twalthr:
I implemented your comments, I hope everything is fine now ;)

#585 solves the issues I had with the TypeExtractor.

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/relational/TPCHQuery10.java
```diff
@@ -0,0 +1,216 @@
...
+import eu.stratosphere.api.java.tuple.Tuple4;
+import eu.stratosphere.api.java.tuple.Tuple5;
+import eu.stratosphere.api.java.tuple.Tuple6;
+
+/**
+ * This program implements a modified version of the TPC-H query 10.
+ * 
+ * The original query can be found at
+ * http://www.tpc.org/tpch/spec/tpch2.16.0.pdf (page 45).
+ *
```
fhueske, line 40: Can you briefly state what the modifications are?

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/relational/TPCHQuery10.java
```diff
@@ -0,0 +1,216 @@
...
+ * 
+ */
+public class TPCHQuery10 {
+
+	public static void main(String[] args) throws Exception {
+		
+		String customerPath = (args.length > 0 ? args[0] : TPCHQuery10.class.getResource("/Testdata/customer.tbl").toString());
+		String ordersPath =  (args.length > 1 ? args[1] : TPCHQuery10.class.getResource("/Testdata/orders.tbl").toString());
+		String lineitemPath =  (args.length > 2 ? args[2] : TPCHQuery10.class.getResource("/Testdata/lineitem.tbl").toString());
+		String nationPath =  (args.length > 3 ? args[3] : TPCHQuery10.class.getResource("/Testdata/nation.tbl").toString());
```
fhueske: We should make sure that the input data resource files are also put into the job jar file and check that a distributed execution of the job leads to correct results.

fhueske:
As said before, we need to make sure that the resource files are properly accessed when the job is packaged as a Jar file and executed in a distributed setting.

Also, you create a lot of new Tuple objects (basically one in each function call). You can use mutable objects to reduce the GC footprint. The `setFields()` method proposed in #600 will make the use of mutable objects more comfortable.

Apart from that, the example job looks good!

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/relational/TPCHQuery10.java
twalthr: How can I make a new Java API program ready for distributed execution? Currently, the entire code is within the main() method. With the record API I had a getPlan() method and returned the plan. How to do that with new API?

StephanEwen: No need to do anything. Main method is fine. If you want cluster execution,
make sure you use the `ExecutionEnvironment.getEnvironment()`
Am 18.03.2014 10:02 schrieb "twalthr" notifications@github.com:

twalthr:
I have changed the pom.xml so that maven creates a separate jar for the TPCH query and includes the testdata. Do you know how I can use files of jars as input for CsvReader?

At the moment the path is incorrect:
<code>
    Caused by: java.lang.IllegalArgumentException: java.net.URISyntaxException: Relative path in absolute URI: jar:file:/home/twalthr/stratosphere-java-examples-0.5-SNAPSHOT-TPCHQuery10.jar!/Testdata/customer.tbl
</code>

Do you think simply copy the files to tmp and read them from there is ok?

StephanEwen:
To use files in jars directly, we would need a ResourceFileSystem, similar
to the local file system. It gets a stream through
ClassLoader#getResourceAsStream
Am 18.03.2014 11:55 schrieb "twalthr" notifications@github.com:

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/relational/TPCHQuery10.java
```diff
@@ -0,0 +1,243 @@
...
+		OutputStream os = new FileOutputStream(tmpFile);
+		int read = 0;
+		byte[] bytes = new byte[1024];
+ 
+		while ((read = is.read(bytes)) != -1) {
+			os.write(bytes, 0, read);
+		}
+		is.close();
+		os.close();
+		return "file://"+tmpFile.toString();
```
StephanEwen, line 241: Can you make this cross platform compatible (use File.toUri().toString()) ?

StephanEwen:
Can you move the path of the test data from `Testdata` to `testdata/tpch`? That way we can include more testdata for other algorithms in the future...

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/relational/TPCHQuery10.java
```diff
@@ -0,0 +1,243 @@
...
+ * Compared to the original TPC-H query this version does not print 
+ * c_phone and c_comment, only filters by years greater than 1990 instead of  
+ * a period of 3 months, and does not sort the result by revenue.
+ * 
+ */
+public class TPCHQuery10 {
+	
+	public static void main(String[] args) throws Exception {
+		
+		String customerPath = (args.length > 0 ? args[0] : copyJarFileToTmp("/Testdata/customer.tbl", "customer.tbl"));
```
StephanEwen, line 77: We have this kind of parameter pattern a lot. I have come to find it a bit weird, though, because it allows you to do a strange combination of having 2 specified file inputs and some default test data. Is that intended? Otherwise you could simply check if args.length < 4 and go with test data in that case.

StephanEwen:
@twalthr Can you do a few cleanups here, then we can merge it,
- Please use aggregations where available (such as for average), no need to implement you r own reducer for that
- I would simply add a `@SuppressWarning("serial")` to the top program and remove all serial version UIDs. That makes it a bit cleaner.
- You can also remove the `throws Exception` from all functions.
- Use either command line parameters, of test data. But potentially mixing both gets confusing

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/relational/TPCHQuery10.java
```diff
@@ -0,0 +1,243 @@
...
+					private static final long serialVersionUID = 1L;
+
+					@Override
+					public boolean filter(Tuple4<Integer, Double, Double, String> t)
+							throws Exception {
+						return t.T4().equals("R");
+					}
+				})
+				// remove flag as it is not necessary anymore
+				.map(new MapFunction<Tuple4<Integer, Double, Double, String>, Tuple3<Integer, Double, Double>>() {
```
fhueske, line 141: Can you replace this by a project transformation?

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/relational/TPCHQuery10.java
```diff
@@ -0,0 +1,243 @@
...
+					private static final int YEAR_FILTER = 1990;
+					
+					@Override
+					public boolean filter(Tuple3<Integer, Integer, String> t) throws Exception {
+						int year = Integer.parseInt(t.T3().substring(0, 4));
+						return year > YEAR_FILTER;
+					}
+				})
+				// remove date as it is not necessary anymore
+				.map(new MapFunction<Tuple3<Integer,Integer,String>, Tuple2<Integer, Integer>>() {
```
fhueske, line 119: Can you replace this by a project transformation?

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/relational/TPCHQuery10.java
```diff
@@ -0,0 +1,243 @@
...
+					public Tuple3<Integer, Double, Double> map(Tuple4<Integer, Double, Double, String> t) throws Exception {
+						return new Tuple3<Integer, Double, Double>(t.T1(), t.T2(), t.T3());
+					}
+				});
+
+		// join orders with lineitems
+		// custkey, extendedprice, discount
+		DataSet<Tuple3<Integer, Double, Double>> lineitemsOfCustomerKey = ordersFilteredByYear.joinWithHuge(lineitemsFilteredByFlag)
+				.where(0).equalTo(0)
+				.with(new JoinFunction<Tuple2<Integer, Integer>, Tuple3<Integer, Double, Double>, Tuple3<Integer, Double, Double>>() {
```
fhueske, line 154: Replace UDF join with project join.

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/relational/TPCHQuery10.java
```diff
@@ -0,0 +1,243 @@
...
+					}
+				});
+
+		// join customer with nation
+		// custkey, name, address, nationname, acctbal
+		DataSet<Tuple5<Integer, String, String, String, Double>> customerWithNation = customers
+				.joinWithTiny(nations)
+				.where(3)
+				.equalTo(0)
+				.with(new JoinFunction<Tuple5<Integer, String, String, Integer, Double>, Tuple2<Integer, String>, Tuple5<Integer, String, String, String, Double>>() {
```
fhueske, line 192: Replace by project join.

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/relational/TPCHQuery10.java
```diff
@@ -0,0 +1,243 @@
...
+					}
+				});
+
+		// join customer (with nation) with revenue
+		// custkey, name, address, nationname, acctbal, revenue
+		DataSet<Tuple6<Integer, String, String, String, Double, Double>> customerWithRevenue = customerWithNation
+				.join(revenueOfCustomerKey)
+				.where(0)
+				.equalTo(0)
+				.with(new JoinFunction<Tuple5<Integer, String, String, String, Double>, Tuple2<Integer, Double>, Tuple6<Integer, String, String, String, Double, Double>>() {
```
fhueske, line 208: Replace by project join.

fhueske:
Can you rebase to the current master and use projection and join projection to reduce the code?

We should also decide how to consistently handle test data and output for all examples.

StephanEwen:
I will merge this, but will not include the generated TPC-H data for now.

I have sent a mail to the TPC asking for clearance to ship sample data. Their response was that they are updating their terms in the next months to accommodate for such things, but it will take a while.
For now, it is safer to not ship these files.

StephanEwen:
Merged in a947361c5b5e2ec3b023b923a5f9c72e6ef14334

PR #569, opened 2014-03-13
Fix for termination criterion issues

tillrohrmann:
I encountered some minor problems while testing the new termination criterion for bulk iterations and I tried to fix them.
1. NullPointerException in OptimizerNode.mergeLists because child2open is not checked for being not null
2. Faulty chaining of root of termination criterion and root of step function. This happens if one uses the next partial solution as the termination criterion
3. BulkIterationNode tries to generate alternative plans for the termination criterion even if there are no plans for the next partial solution. If the termination criterion depends on the next partial solution, then the algorithm won't find a plan fulfilling the requirements and thus throw an exception.
4. Some of the termination criterion nodes are instantiated as a RegularPactTasks instead of IntermediateIterationTasks. This is caused by not setting the dynamic paths of the termination criterion.

I hope that I haven't broken anything while trying to fix the issues.

### DIFF
--- a/stratosphere-compiler/src/main/java/eu/stratosphere/compiler/PactCompiler.java
+++ b/stratosphere-compiler/src/main/java/eu/stratosphere/compiler/PactCompiler.java
@@ -989,7 +989,11 @@ public class PactCompiler {
 				this.numMemoryConsumers += recursiveCreator.numMemoryConsumers;
 				
 				// go over the contained data flow and mark the dynamic path nodes
-				rootOfStepFunction.accept(new StaticDynamicPathIdentifier(iterNode.getCostWeight()));
+				StaticDynamicPathIdentifier identifier = new StaticDynamicPathIdentifier(iterNode.getCostWeight());
+				rootOfStepFunction.accept(identifier);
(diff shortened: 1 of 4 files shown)


markus-h:
Thanks for your testing! Your fixes look good for me. 
I don't fully understand the 3rd issue. Can you describe the case where there are no plans for next partial solution and candidates.size() == 0?

tillrohrmann:
The 3rd issue always arises if all generated plans for the next partial solution are discarded because they don't met the properties (instantiateCandidate method of BulkIterationNode). If all plans are discarded then there is no plan for the required properties and thus there is no need to look for a plan for the termination criterion. But since the termination criterion often depends on the next partial solution, it requires the next partial solution to have at least one plan. If not, then the system complains wrongly about the termination criterion plan which cannot be instantiated. But in the case where the next partial solution has no plans for given properties, we should never try to find a plan for the termination criterion.

markus-h:
Ok, that makes sense, thanks for your explanation!

StephanEwen:
Merged in 525cb03668b6f6a6bdb4c5c91d350d2e1670c319

PR #572, opened 2014-03-13
Scala termination criterion

tillrohrmann:
I tried to extend the scala api by the newly introduced termination criterion for bulk iterations. Since one has neither direct access to the partial solution nor to the next partial solution one has to specify a termination function of type (previousState: DataSet[T], currentState: DataSet[T]) => DataSet[U] which is supplied with the partial and next partial solution. I hope this is flexible enough to cover many termination caluclations.

Additionally I extended the scala api by a name parameter for the write method of DataSet. It makes debugging a lot easier if one can name the data sinks properly.

I also included a small example which has however only instructional value.

### DIFF
--- a/stratosphere-examples/stratosphere-scala-examples/src/main/scala/eu/stratosphere/examples/scala/iterative/TerminationCriterion.scala
+++ b/stratosphere-examples/stratosphere-scala-examples/src/main/scala/eu/stratosphere/examples/scala/iterative/TerminationCriterion.scala
@@ -1,0 +1,57 @@
+package eu.stratosphere.examples.scala.iterative
+
+import eu.stratosphere.client.LocalExecutor
+import eu.stratosphere.api.common.Program
+import eu.stratosphere.api.common.ProgramDescription
+
+import eu.stratosphere.api.scala._
+import eu.stratosphere.api.scala.operators._
+import eu.stratosphere.types.DoubleValue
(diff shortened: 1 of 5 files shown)


stratosphere-examples/stratosphere-scala-examples/src/main/scala/eu/stratosphere/examples/scala/iterative/TerminationCriterion.scala
```diff
@@ -0,0 +1,57 @@
+package eu.stratosphere.examples.scala.iterative
```
rmetzger, line 14: Your change is not passing our integration tests because this file is missing the LICENSE header.

rmetzger, line 14: See: https://travis-ci.org/stratosphere/stratosphere/builds/20732152

tillrohrmann, line 14: Thanks, I just added the License header.

StephanEwen:
Merged in 1aa1d26b9370088cf77f5fc86f506377bf46bfef

Issue #563, opened 2014-03-12
New Java API: reduce function does not allow to return input object.


fhueske:
The reduce function of the new Java API does not work if the an input object is returned, i.e.,

New example:

```
public T reduce(T value1, T value2) throws Exception {
  value1.setVal(value1.getVal() + value2.getVal());
  return value1;
}
```

This is due to the internal mutable object runtime.

Wrong example:

```
public T reduce(T value1, T value2) throws Exception {
  value1 = value1 + value2;
  return value1;
}
```

StephanEwen:
Can you elaborate a bit more, what does happen in this case? In the
example, it looks like the value1 reference (or value) is reassigned to
value1+value2. That does not modify any internal object, or am I
overlooking something?

fhueske:
I observed the problem with the `ReferenceWrappingReducer` in [PlanReduceOperator](https://github.com/stratosphere/stratosphere/blob/master/stratosphere-java/src/main/java/eu/stratosphere/api/java/operators/translation/PlanReduceOperator.java) but expect this to have the same effect in [ReduceFunction](https://github.com/stratosphere/stratosphere/blob/master/stratosphere-java/src/main/java/eu/stratosphere/api/java/functions/ReduceFunction.java).

When looping over the iterator, the final hasNext() call in the wrapping `reduce(Iterator<Reference<T>> in, Collector<Reference<T>> out)` overwrites the curr object.
I think the problem is that return values are not _safe_ when returned from the user function as in the other function which serialize output values or immediately forward them to the next function.

I am currently working on a fix that copies the result of the function call into another mutable object using a Serializer.

fhueske:
Damn, stupid example.
It should be something like:

```
public T reduce(T value1, T value2) throws Exception {
  value1.setVal(value1.getVal() + value2.getVal());
  return value1;
}
```

StephanEwen:
Let's see if we can solve this without copying... Otherwise that kills
performance...

StephanEwen:
Can you post a concrete example where it fails? Maybe add a unit test that fails due to this problem?

fhueske:
I added a unit test for this to PR #578.

fhueske:
In commit cf0eff3b1386486ed656aed1c1e8260c0cba5a09 I added JavaDocs to `T ReduceFunction.reduce(T value1, T value2)` function that tell the user that the second input object (`value2`) can be safely returned while the first input object (`value1`) must not be returned (checked by an integration test).
[...]

PR #578, opened 2014-03-14
Fix for ReduceOperator mutable object return bug.

fhueske:
This PR proposes a fix for #563.

Performance implications should be tolerable. For Sort-Group strategy only the first value of a group is copied. For Hash-Aggregate strategy (not supported yet), there will be no performance implications.

The fix assumes that the object of the first input value (`val1`) of a `T reduce(T val1, T val2)` may be returned and that the object of the second input value (`val2`) may NOT be returned.

### DIFF
--- a/stratosphere-java/src/test/java/eu/stratosphere/api/java/functions/ReduceFunctionTest.java
+++ b/stratosphere-java/src/test/java/eu/stratosphere/api/java/functions/ReduceFunctionTest.java
@@ -1,0 +1,104 @@
+/***********************************************************************************************************************
+ *
+ * Copyright (C) 2010-2013 by the Stratosphere project (http://stratosphere.eu)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ **********************************************************************************************************************/
+package eu.stratosphere.api.java.functions;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+import eu.stratosphere.api.common.typeutils.TypeComparator;
+import eu.stratosphere.api.common.typeutils.TypeSerializer;
+import eu.stratosphere.api.java.functions.util.CollectionOutputCollector;
+import eu.stratosphere.api.java.functions.util.KeyGroupedIterator;
+import eu.stratosphere.api.java.functions.util.MutableObjectIteratorWrapper;
+import eu.stratosphere.api.java.tuple.Tuple2;
+import eu.stratosphere.api.java.typeutils.BasicTypeInfo;
+import eu.stratosphere.api.java.typeutils.TupleTypeInfo;
+import eu.stratosphere.util.Collector;
+import eu.stratosphere.util.MutableObjectIterator;
+
+public class ReduceFunctionTest {
+
+	@Test
+	public void reduceTest() throws Exception {
+		
+		// input and output data
+		List<Tuple2<Integer, Integer>> inData = new ArrayList<Tuple2<Integer, Integer>>();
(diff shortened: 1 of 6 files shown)


stratosphere-java/src/test/java/eu/stratosphere/api/java/functions/ReduceFunctionTest.java
```diff
@@ -0,0 +1,112 @@
...
+				new ListOutputCollector<Tuple2<Integer, Integer>>(outData, serializer);
+		
+		// run reduce function on input data
+		try {
+			while (kgi.nextKey()) {
+				testFunc.reduce(kgi.getValues(), out);
+			}
+		} catch(IOException ioe) {
+			Assert.fail();
+		} catch(Exception e) {
```
rmetzger: Why do you have this construct here? Why are you not printing a stack trace on an IOException?
Isn't there a way to hand over the exception to JUnit?

stratosphere-java/src/test/java/eu/stratosphere/api/java/functions/util/KeyGroupedIterator.java
```diff
@@ -0,0 +1,212 @@
...
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+
+import eu.stratosphere.api.common.typeutils.TypeComparator;
+import eu.stratosphere.api.common.typeutils.TypeSerializer;
+import eu.stratosphere.util.MutableObjectIterator;
+
+/**
+ * This class is an exact copy of the KeyGroupedIterator in stratosphere-runtime.
+ * Stratosphere-runtime cannot be added as a dependency because this would cause cyclic dependencies.
```
rmetzger, line 27: Can't you move this class to the Stratosphere-core?
I think both stratosphere-java and stratosphere-runtime would have access to it.

stratosphere-java/src/test/java/eu/stratosphere/api/java/functions/util/ListOutputCollector.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.util.List;
+
+import eu.stratosphere.api.common.typeutils.TypeSerializer;
+import eu.stratosphere.util.Collector;
+
+/**
+ * Implements an output collector that stores its collected elements in a supplied list.
+ */
+public class ListOutputCollector<E> implements Collector<E> {
```
rmetzger: Why did you not create a `CollectionOutputCollector` that expects Collection instead of a list?
The array you are creating to validate the test would still have an order but others might be happy to use this code as well (even if they don't need a list, but just a collection).
(Its just a suggestion)

stratosphere-java/src/test/java/eu/stratosphere/api/java/functions/util/MutableObjectIteratorWrapper.java
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Wraps a regular iterator and emulates a MutableObjectIterator.
+ */
+public class MutableObjectIteratorWrapper<E> implements MutableObjectIterator<E> {
+	private final Iterator<E> source;
+	private final TypeSerializer<E> serializer;
+	
+	public MutableObjectIteratorWrapper(Iterator<E> source, TypeSerializer<E> serializer)
+	{
```
rmetzger, line 31: Checkstyle will soon reject pull requests with brackets in the next line (https://github.com/stratosphere/stratosphere/issues/596)

fhueske, line 31: I guess that's still up for discussion ;-)

stratosphere-java/src/test/java/eu/stratosphere/api/java/functions/util/ListOutputCollector.java
fhueske: Good idea

stratosphere-java/src/test/java/eu/stratosphere/api/java/functions/ReduceFunctionTest.java
fhueske: Thanks, will update.

stratosphere-java/src/test/java/eu/stratosphere/api/java/functions/util/KeyGroupedIterator.java
fhueske, line 27: Not sure about this one because it is only a test dependency. 
Will think about it.

fhueske:
PR was rebased and updated as #614.

PR #577, opened 2014-03-14
Fix issue #505 removing trailing \r from lines in Windows text files

tuantrieu:
Window uses \r\n to split lines. so, for text files from Windows, when
\n is used as delimiter, \r is still appear at the end of lines.

The code checks to see if \n is used as delimiter and end of a line is
\r, then \r is removed from the line

PR #582, opened 2014-03-15
Fix issue about Windows line break in TextInputFormat

tuantrieu:
Window uses \r\n to split lines. so, for text files from Windows, when \n is used as delimiter, \r is still appear at the end of lines.

The code checks to see if \n is used as delimiter and end of a line is \r, then \r is removed from the line

The change is the same as in https://github.com/stratosphere/stratosphere/pull/577, but this time I move it into the TextInputFormat. https://github.com/stratosphere/stratosphere/pull/577 failed the test case for terasort because the input file uses DelimitedInputFormat but it is not a text file.

### DIFF
--- a/stratosphere-java/src/main/java/eu/stratosphere/api/java/io/TextInputFormat.java
+++ b/stratosphere-java/src/main/java/eu/stratosphere/api/java/io/TextInputFormat.java
@@ -34,12 +34,12 @@ public class TextInputFormat extends DelimitedInputFormat<String> {
 	/**
 	 * Code of \r, used to remove \r from a line when the line ends with \r\n
 	 */
-	private static final byte CARRIAGE_RETURN = 13;
+	private static final byte CARRIAGE_RETURN = (byte) '\r';
 
 	/**
 	 * Code of \n, used to identify if \n is used as delimiter
 	 */
-	private static final byte NEW_LINE = 10;
+	private static final byte NEW_LINE = (byte) '\n';
 
 	
 	// --------------------------------------------------------------------------------------------
--- a/stratosphere-java/src/main/java/eu/stratosphere/api/java/io/TextInputFormat.java
+++ b/stratosphere-java/src/main/java/eu/stratosphere/api/java/io/TextInputFormat.java
@@ -31,6 +31,17 @@ public class TextInputFormat extends DelimitedInputFormat<String> {
 	
 	private transient Charset charset;
 
+	/**
+	 * Code of \r, used to remove \r from a line when the line ends with \r\n
+	 */
+	private static final byte CARRIAGE_RETURN = 13;
+
+	/**
+	 * Code of \n, used to identify if \n is used as delimiter
+	 */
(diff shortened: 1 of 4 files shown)


StephanEwen:
Looks like good start :-)

We would need a test case, though, before we can merge the patch. The test case needs to cover cases where
- `\n` is used
- `\r\n` is used
- a different line delimiter is used

StephanEwen:
Looking at this specific functionality...

Does it make sense to decouple the TextInputFormat from the DelimitedInputFormat? The following arguments would speak for that
- I think text inputs always have `n` (or`\r\n`) as delimiters.
- We could be sure the delimiter is always one character (simplify logic)
- Other cases would still work with a DelimitedInputFormat, or a 1-field CSV input format

StephanEwen:
The TextValueInputFormat is a variant of the TextInputFormat made for mutable string objects. It has the same limitation as the TextInputFormat had and would also benefit from the here added functionality. This could be a separate issue, though,

stratosphere-java/src/main/java/eu/stratosphere/api/java/io/TextInputFormat.java
```diff
@@ -31,6 +31,17 @@
 	
 	private transient Charset charset;
 
+	/**
+	 * Code of \r, used to remove \r from a line when the line ends with \r\n
+	 */
+	private static final byte CARRIAGE_RETURN = 13;
```
StephanEwen: If you use `private static final byte CARRIAGE_RETURN = (byte) '\r';`, it becomes more intuitive.

stratosphere-java/src/main/java/eu/stratosphere/api/java/io/TextInputFormat.java
```diff
@@ -31,6 +31,17 @@
...
+
+	/**
+	 * Code of \n, used to identify if \n is used as delimiter
+	 */
+	private static final byte NEW_LINE = 10;
```
StephanEwen: Same comment as above

tuantrieu:
Thank for your suggestion. I will make test cases for this change.

In your comment about decoupling TextInputFormat from the DelimitedInputFormat, do you mean that this change could be done in DelimitedInputFormat ?

I am not sure if I understand what you mean, but I was thinking that files that can use delimiters must be text files, so I made this change in DelimitedInputFormat. However, it failed the TeraSortITCase test case  where input file is a text file with lines containing 100 characters including '\r\n', the test case expects exactly 99 characters for lines ( after delimiting lines by '\n'), 10 for key + 89 characters (including '\r') for values. So I moved this change to TextInputFormat to pass this test case.

tuantrieu:
Could you take a look at my test cases ?

fhueske:
Tests look good to me. Thank you very much!

StephanEwen:
Merged in 4b6af0b11831b68aba4b50c69563ca2264ddd85e

PR #586, opened 2014-03-15
Suggested Name change for the WriteMode

StephanEwen:
Got as feedback that CREATE as default write mode does not convey an intuitive meaning. Here is a suggested change of names.

CREATE --> NO_OVERWRITE
OVERWRITE FORCE_OVERWRITE

### DIFF
--- a/stratosphere-clients/src/main/java/eu/stratosphere/client/LocalExecutor.java
+++ b/stratosphere-clients/src/main/java/eu/stratosphere/client/LocalExecutor.java
@@ -332,6 +332,14 @@ public class LocalExecutor extends PlanExecutor {
 		LogUtils.initializeDefaultConsoleLogger(lvl);
 	}
 	
+	/**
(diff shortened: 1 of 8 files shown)


fhueske:
I agree that CREATE is not an intuitive naming and should be changed to NO_OVERWRITE.

But why do you want to change OVERWRITE to FORCE_OVERWRITE?
I think OVERWRITE is shorter and has a smaller chance of misinterpretation (What does FORCE mean? Is there OVERWRITE without FORCE?).

StephanEwen:
Yes, I see the point. The prefix force came from the unix tools, which often use `-f`or `--force` to overwrite. So I guess the options are 

I will change it to `NO_OVERWRITE` & `OVERWRITE`.

StephanEwen:
Adjusted and merged in a65ee8f600d16596f6193af444265de515ec80c4

PR #614, opened 2014-03-20
Fix for ReduceOperator mutable object return bug.

fhueske:
This PR proposes a fix for #563 and is a rebased version of PR #578.

Performance implications should be tolerable. For Sort-Group strategy only the first value of a group is copied. For Hash-Aggregate strategy (not supported yet), there will be no performance implications.

The fix assumes that the object of the first input value (`val1`) of a `T reduce(T val1, T val2)` may be returned and that the object of the second input value (`val2`) may NOT be returned.

### DIFF
--- a/stratosphere-java/src/main/java/eu/stratosphere/api/java/functions/ReduceFunction.java
+++ b/stratosphere-java/src/main/java/eu/stratosphere/api/java/functions/ReduceFunction.java
@@ -18,30 +18,56 @@ import java.util.Iterator;
 
 import eu.stratosphere.api.common.functions.AbstractFunction;
 import eu.stratosphere.api.common.functions.GenericGroupReduce;
+import eu.stratosphere.api.common.typeutils.TypeSerializer;
 import eu.stratosphere.util.Collector;
 
 
 public abstract class ReduceFunction<T> extends AbstractFunction implements GenericGroupReduce<T, T> {
 	
 	private static final long serialVersionUID = 1L;
-
-
-	public abstract T reduce(T value1, T value2) throws Exception;
 	
+	private TypeSerializer<T> serializer;
+	private T curr; 
+
+	/**
+	 * Combines two values into one.
+	 * The reduce function is consecutively applied to all values of a group until only a single value remains.
+	 * In functional programming, this is known as a fold-style aggregation.
+	 * 
+	 * Important: It is fine to return the first value (value1) as result from this function. 
+	 * You may NOT return the second value from this function.
+	 * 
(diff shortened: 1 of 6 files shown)


StephanEwen:
I think this is a flavor of the problem found in #131.

If I am not mistaken, then the problem is the following: Modifying a value may potentially overwrite the key in that value and cause further KeyGroupIterator comparisons to chunk the stream into key groups to fail.

If that is the only problem, then I think we can fix it without copying by introducing a third value in the KeyGroupedIterator.

Are there other problems that I overlooked?

fhueske:
Well, it is related to #131 because it is a problem in the context of the mutable object iterator of the Reduce operator. However, the problem itself is different. In #131, I proposed to return the same object instance for every `next()` call of the iterator. That would make this issue even more problematic...

The problem of this issue is, that an implementation of `T reduce(T t1, T t2)` might return `t1` or `t2` (either the same one every time or picking one depending on the context)
Let's have a look at the different cases:
1. **The function returns `t1` in every call.** In this case, the object in which the result is "aggregated" will be the same for the whole group because `ReduceFunction` performs the next call of `reduce(t1, t2)` with `t1` being the result of the previous call (or the first element of the group). If the mutable object iterator alternates between two objects, every second call of `hasNext()` will cause that the "aggregated" result will be overwritten by the next element of the group (or even the first element of the next group). Adding another mutable object to the iterator will only increase the interval but not fix this issue. This PR proposes to copy the first element of the group into a dedicated object which will hold the aggregated result while the whole group is processed and thereby fix the issue for this case.
2. **The function returns `t2` in every call.** In this case we should be fine, because the object that holds the aggregated value is always different from the object that will hold the next element of the group. I have not verified this with a unit test. This would contradict with the Java Doc comment added by this PR but would not require any code changes for this case.
3. **The function returns either `t1` or `t2`.** In this case, copying the first object does not help because the dedicated object might get replaced by one of the iterators mutable objects. 

So this PR does not solve the problem but what can we do?
1. Leave the code as it is (after checking that it works) and add a Java Doc that never the first parameter (`t1`) might be returned by the function. This requires that the programmers read and follow the java docs.
2. Try to make the code work for any user function by checking if the returned object is the same as the object which is returned next by the iterator and make a copy in this case. This will cause high overhead due to checks and potential copy operations.
3. Change the signature of reduce to `void reduce(T t1, T t2, T out)` and explicitly provide the object that should hold the result of reduce.
4. Something else???

fhueske:
Commit cf0eff3b1386486ed656aed1c1e8260c0cba5a09 goes for the first option (leave code as it is and add documentation).

Issue #566, opened 2014-03-13
Add project operator to new JavaAPI


fhueske:
Inspired by #552, I thought a project operator for Tuple DataSets would be really convenient.

Can we add a project operator to the new JavaAPI like this?

```
DataSet<Tuple4<Long, Integer, Double, String>> dataSet = ...
DataSet<Tuple2<String, Integer>> projectedDS = dataSet.project(3,1);
```

Where the fields 1 and 3 of the original tuple are moved into a Tuple2.
An operator like this would only work on DataSet of Tuples and not on arbitrary objects.

Do you think that could be done and is it worthwhile even if it is only working for Tuple Datasets?

uce:
I think it's perfectly fine to have it only working with Tuple DataSets as we should aim at making it as convenient as possible for users. :+1: 

I leave the question whether it can be done to @StephanEwen.

fhueske:
Added with #617.

PR #617, opened 2014-03-20
Add project operator to new Java API

fhueske:
Adds a project operator to the new Java API (Issue #566)
Project works only on Tuple datasets and is used like:

``` java
DataSet<Tuple3<Integer, Double, String>> ds = ...
DataSet<Tuple2<String,Integer>> pds = ds.project(2,0).types(String.class, Integer.class);
DataSet<Tuple1<Double>> pds2 = ds.project("FTF").types(Double.class);
DataSet<Tuple2<Integer, String>> pds3 = ds.project(true, false, true).types(Integer.class, String.class);
```

Also added unit tests and integration tests for the project operator.

### DIFF
--- a/stratosphere-java/src/main/java/eu/stratosphere/api/java/DataSet.java
+++ b/stratosphere-java/src/main/java/eu/stratosphere/api/java/DataSet.java
@@ -113,7 +113,7 @@ public abstract class DataSet<T> {
 		if(fieldIndexes.length == 0) {
 			throw new IllegalArgumentException("project() needs to select at least one (1) field.");
 		} else if(fieldIndexes.length > 22) {
-			throw new IllegalArgumentException("project() may to select at most twenty-two (22) fields.");
+			throw new IllegalArgumentException("project() may select only up to twenty-two (22) fields.");
 		}
 		
 		int maxFieldIndex = ((TupleTypeInfo<?>)this.getType()).getArity();
--- a/stratosphere-java/src/main/java/eu/stratosphere/api/java/DataSet.java
+++ b/stratosphere-java/src/main/java/eu/stratosphere/api/java/DataSet.java
@@ -15,7 +15,7 @@
 package eu.stratosphere.api.java;
 
 import org.apache.commons.lang3.Validate;
-
+import java.util.Arrays;
 import eu.stratosphere.api.common.io.FileOutputFormat;
 import eu.stratosphere.api.common.io.OutputFormat;
 import eu.stratosphere.api.java.aggregation.Aggregations;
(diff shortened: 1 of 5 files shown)


stratosphere-java/src/main/java/eu/stratosphere/api/java/DataSet.java
```diff
@@ -79,7 +94,87 @@ public ExecutionEnvironment getExecutionEnvironment() {
...
+	public Projection<T> project(int... fieldIndexes) {
+		
+		if(!(this.getType() instanceof TupleTypeInfo)) {
+			throw new UnsupportedOperationException("project() can only be applied to DataSets of Tuples.");
+		}
+		
+		if(fieldIndexes.length == 0) {
+			throw new IllegalArgumentException("project() needs to select at least one (1) field.");
+		} else if(fieldIndexes.length > 22) {
+			throw new IllegalArgumentException("project() may to select at most twenty-two (22) fields.");
```
rmetzger: Typo? 
"project() may to select at most twenty-two (22) fields."
"up to 22 fields" ?

fhueske:
Updated PR.

stratosphere-java/src/main/java/eu/stratosphere/api/java/operators/translation/PlanProjectOperator.java
```diff
@@ -0,0 +1,75 @@
...
+	public static final class MapProjector<T, R extends Tuple> 
+		extends AbstractFunction
+		implements GenericMap<T, R>
+	{
+		private static final long serialVersionUID = 1L;
+		
+		private final int[] fields;
+		private final R outTuple;
+		
+		private MapProjector(int[] fields, TypeSerializer<R> serializer) {
```
StephanEwen: Can you directly pass an instance of the tuple (or derived tuple) here, rather than doing the detour through the serializer?

stratosphere-java/src/main/java/eu/stratosphere/api/java/DataSet.java
```diff
@@ -79,7 +94,87 @@ public ExecutionEnvironment getExecutionEnvironment() {
...
+	 * Selects a subset of fields of a Tuple to create new Tuples. 
+	 * 
+	 * @param fieldIndexes The field indexes select the fields of an input tuple that are kept in an output tuple.
+	 * 					   The order of fields in the output tuple depends on the order of field indexes.
+	 * @return Returns a Projection. Call the types() method with the correct number of class parameters 
+	 *         to create a ProjectOperator. 
+	 */
+	public Projection<T> project(int... fieldIndexes) {
+		
+		if(!(this.getType() instanceof TupleTypeInfo)) {
```
StephanEwen: For the other operators, we put the logic into the operator class (with different constructors). That keeps the already large DataSet class more compact and readable.

I would vote to do the same here.

StephanEwen:
Looks very good. I have to minor comments inline.

fhueske:
@StephanEwen Thanks for the improvement suggestions!
I updated the PR and rebased it to the current master.

fhueske:
Will merge this PR once final tests passed.

Issue #427, opened 2014-01-23
DecimalTextIntParser has no static method that simply returns an int


sscdotopen:
DecimalTextIntParser has no static method that simply returns an int (in contrast to DecimalTextLongParser). That makes it hard to use.

StephanEwen:
I agree, we should add this.

fhueske:
Some of the other parsers are also lacking a static method.

StephanEwen:
Fixed in a1555ce84ae47edb8cd696f672935cbeffd3481b

PR #618, opened 2014-03-20
Added static method for DecimalTextIntParser

gruay:
I added the static method for DecimalTextIntParser (issue #427) the same way is done in DecimalTextLongParser.
I'm planning on adding static methods for the rest of the parsers.

### DIFF
--- a/stratosphere-core/src/main/java/eu/stratosphere/types/parser/DecimalTextIntParser.java
+++ b/stratosphere-core/src/main/java/eu/stratosphere/types/parser/DecimalTextIntParser.java
@@ -75,4 +75,38 @@ public class DecimalTextIntParser extends FieldParser<IntValue> {
 	public IntValue getLastResult() {
 		return this.result;
 	}
+	
+	public static final int parseField(byte[] bytes, int startPos, int length, char delim) {
+		long val = 0;
+		boolean neg = false;
+		
+		
+		if (bytes[startPos] == '-') {
+			neg = true;
+			startPos++;
+			length--;
+			// check for empty field with only the sign
+			if (length == 0 || bytes[startPos] == delim) {
+				throw new NumberFormatException("Orphaned minus sign.");
+			}
+		}
+		
+		for (; length > 0; startPos++, length--) {
+			if (bytes[startPos] == delim) {
+				return neg ? -(int)val : (int)val;
+			}
+			if (bytes[startPos] < 48 || bytes[startPos] > 57) {
+				throw new NumberFormatException();
+			}
+			val *= 10;
+			val += bytes[startPos] - 48;
+			
+			if (val > OVERFLOW_BOUND && (!neg || val > UNDERFLOW_BOUND)) {
+				throw new NumberFormatException("Number format Overlfow/Underflow");
+			}
+		}
+		return neg ? -(int)val : (int)val;
+	}
+
+	
 }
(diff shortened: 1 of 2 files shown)


rmetzger:
Hey,
thank you for your pull request.
Is there a way to use only one method for the parsing? I have the impression that `public int parseField()` and `public static final int parseField()` are quite similar.

gruay:
Right now the non static one is returning the position in the byte array where it ended parsing the int (or -1 if something went wrong) and then the parsed number is put in a parameter given, whereas the static one simply returns the parsed int.
So, they both do the same but the non-static one gives more information.
I'm not sure if that answer your question now that I re-read it

rmetzger:
I was just wondering if there would be a way to avoid duplicate code, for example by introducing an "internal" method that can handle both cases?

gruay:
Sure, I could move the non-static code to a new static method that would do the same. Then the non-static method would call that one for getting both the ending position and the value and the static method would just care about the parsed value.
Do you mean something like this?

gruay:
Hey, I just created the new static method that is called from both methods.
Did you mean something like this?

stratosphere-core/src/main/java/eu/stratosphere/types/parser/DecimalTextIntParser.java
```diff
@@ -28,19 +28,40 @@
...
+			int end = parseIntField(bytes, startPos, limit, delimiter, reusable);
+			this.result = reusable;
+			return end;
+		} 
+		catch (NumberFormatException e) {
+			if (e.getMessage() == null) {
+				setErrorState(ParseErrorState.NUMERIC_VALUE_ILLEGAL_CHARACTER);
+				return -1;
+			}
+			else if (e.getMessage().equals("Orphaned minus sign.")) {
```
rmetzger, line 41: Differentiating between exceptions by their messages is very bad design.
Imagine someone changes the exception messages. (You could use constants for that)
BUT: a) you could throw different kinds of exceptions.
b) you could implement a custom Exception class with an Enum to represent the different error states.

StephanEwen:
I agree with Robert. That is very fragile code, the way the error messages and return codes are handled. Also, throwing an exception and then turning that into a state seems reversed. Why not parse and set the state, and then react with an exception to the state (if it is not the "OKAY" state)?

StephanEwen:
I am not sure about the use case of this method, but I have a feeling that the delimiter is odd. I think @sscdotopen 's initial case was that it is odd to come up with some delimiter only to have a standalone array parsed.

The original parse method is used to parse a field as part of a CSV row. The standalone method ,seems to me, it rather intended to parse an isolated value inside an array.

@sscdotopen Can you comment on this?

StephanEwen:
What do we do with this pull request? I don't want to merge it in the state that it is in, I think it needs to be improved.

Is anything going to happen here?

fhueske:
I think the problem is, that it is not really clear what issue #427 is about...

StephanEwen:
I think this issue simply asks for a utility to parse binary text data (as
read from a file stream) directly into a primitive type.

Is it fair to ask whoever really needs it should implement it? I think it
is not a priority thing right now.

StephanEwen:
I will close this pull request, because the code is not in shape to be merged and the functionality has by now been subsumed by a1555ce84ae47edb8cd696f672935cbeffd3481b including tests.

Issue #600, opened 2014-03-18
Add SetFields() Method to Tuples


fhueske:
Add a `setFields()` method to tuples to set all fields of a tuple in one method call:
`public void Tuple3.setFields(T1 value1, T2 value2, T3 value3)`

twalthr:
I will implement it.

fhueske:
You know that you only need to change TupleGenerator for that, right?

twalthr:
yes ;-)

fhueske:
Fixed in b6edba26c3443ebe33a3de4f482900fa424430f2

Issue #599, opened 2014-03-18
Tuple field access indexes are off by one


fhueske:
Tuple fields start at 1 (`Tuple1._1`) but are accessed with indexes starting at 0:
`Tuple1.getField(0)` and `Tuple1.T1()` return the same field `Tuple1._1`.

I find this confusing as would make it consistent for example by starting the `getField()` at 1 as well.

fhueske:
The same applies for the key definitions, e.g., `DataSet.groupBy(0)` and `JoinOperatorSets.where(0)` (and so on) all refer to `Tuple._1`

twalthr:
I also ran into this issue at the beginning. I would prefer to start all indexes at 0 as it is usual in computer science.

fhueske:
I am fine either way as long as it is consistent :smirk:

StephanEwen:
The tuples are inspired by the scala tuples. Those start at one. So the
choice is not easy...
Am 18.03.2014 09:25 schrieb "Fabian Hueske" notifications@github.com:

fhueske:
Right, but I think we should have a decision rather soon, i.e., before we announce / release the API. Otherwise, we might silently break existing stuff.

rmetzger:
I'm also in favor of zero-indexed tuples.
I would say Scala is inconsistent and we should not support that. The majority of our users are Java users, and not Scala users.

rmetzger:
Rename it to "v0, v1, ..."

twalthr:
Sorry, I need to ask again. I'm currently working in the TupleGenerator.
So v0,v1,.. was our final decision? I thought f0, f1,..
But _1,_2,.. should still exist for scala developers, right?

fhueske:
I am in favor of replacing `private T1 _1;` by `public T0 f0;` and so on.
Also remove the getter `public T1 T1()` and setter `public void T1(T1 value)`.

This is a Java API and we should not think too much about the 1% Scala developers out there. These guys are probably smart enough and know that other languages start counting at 0 :smirk:

fhueske:
Fixed in b6edba26c3443ebe33a3de4f482900fa424430f2

Issue #597, opened 2014-03-17
Extend the TupleGenerator for generating Tuple-dependent code in other classes


twalthr:
Some classes also depend on the number of tuples. 
E.g. the `types()` of `CsvReader` or `CLASSES` of `TupleTypeInfo`

@StephanEwen already suggested to extend the TupleGenerator. I would like to do this but I'm not sure how.

What do you think about adding a special comment to these classes so that the tuple generator knows where to put Tuple-dependent code, instead of generating the whole class?

```
class CsvReader {
    // Constructor, methods etc.
    // ......

    // BEGIN_OF_TUPLE_DEPENDENT_CODE

    // END_OF_TUPLE_DEPENDENT_CODE
}
```

alexff91:
Hi, maybe better to do this:
Check if there the same methods, if no add tuple-methods to end of class.
This will help if someone will forgot to add // BEGIN_OF_TUPLE_DEPENDENT_CODE at first to the class.

fhueske:
Fixed in b6edba26c3443ebe33a3de4f482900fa424430f2

PR #619, opened 2014-03-20
Several changes in the TupleGenerator

twalthr:
```
- For Tuple Fields:
    - start with index 0
    - are public (no getters and setters anymore)
    - can be accessed via f0,f1, etc. instead of a method T1(), T2(), etc.
    - setFields() allows for setting all fields at once
- For CsvReader, ProjectOperator, TupleTypeInfo:
    - TupleGenerator generates code for classes which depend on the tuple arity.
    - A special String indicates where to put Tuple dependent code
```

This PR implements #600, #599, #597. The TupleGenerator already supports #617 if the indicator is added.


fhueske:
PR looks good. Will merge.

fhueske:
Merged in b6edba26c3443ebe33a3de4f482900fa424430f2

Issue #602, opened 2014-03-18
Collecting Output Format for new Java API (only for local executor)


rmetzger:
``` java
List<String> resultsCollected = new ArrayList<String>();
result.output(resultsCollected);
```

qmlmoon:
When dop = 2, then we'll get 2 collection outputs. How to merge them? They have no idea about how the record fields are grouped. So, do we just simply concat them or leave them like  `List<List<String>> resultCollected` ?

fhueske:
Good observation.
In most cases, you can simply concatenate them since the order of elements is not important.
If the user specifies sorted output, going for nested lists would be the best thing to do.

For now, I would go with the concatenated lists as it is easier to handle for the user.

rmetzger:
This is related: https://github.com/stratosphere/stratosphere/issues/478 (for an implementation into the runtime)

fhueske:
Issue solved in b032ff3afb10f9e0158e133a8f5c447ad89a90b6
Thanks!

PR #623, opened 2014-03-21
Add CollectionOutputFormat for local executor

qmlmoon:
implement issue #602

### DIFF
--- a/stratosphere-java/src/main/java/eu/stratosphere/api/java/DataSet.java
+++ b/stratosphere-java/src/main/java/eu/stratosphere/api/java/DataSet.java
@@ -25,7 +25,6 @@ import eu.stratosphere.api.java.functions.GroupReduceFunction;
 import eu.stratosphere.api.java.functions.KeySelector;
 import eu.stratosphere.api.java.functions.MapFunction;
 import eu.stratosphere.api.java.functions.ReduceFunction;
-import eu.stratosphere.api.java.io.CollectionOutputFormat;
 import eu.stratosphere.api.java.io.CsvOutputFormat;
 import eu.stratosphere.api.java.io.PrintingOutputFormat;
 import eu.stratosphere.api.java.io.TextOutputFormat;
@@ -36,9 +35,6 @@ import eu.stratosphere.api.java.tuple.Tuple;
 import eu.stratosphere.api.java.typeutils.InputTypeConfigurable;
 import eu.stratosphere.api.java.typeutils.TypeInformation;
 import eu.stratosphere.core.fs.Path;
-
-import java.io.Serializable;
-import java.util.Collection;
 
 /**
  *
@@ -198,10 +194,7 @@ public abstract class DataSet<T> {
 	public void writeAsText(String filePath) {
 		output(new TextOutputFormat<T>(new Path(filePath)));
 	}
-
-	public void writeAsCollection(Collection<T> out) {
-		output(new CollectionOutputFormat<T>(out, this.type.createSerializer()));
-	}
+	
 	
 	public void writeAsCsv(String filePath) {
 		writeAsCsv(filePath, CsvOutputFormat.DEFAULT_LINE_DELIMITER, CsvOutputFormat.DEFAULT_FIELD_DELIMITER);
--- a/stratosphere-java/src/main/java/eu/stratosphere/api/java/DataSet.java
+++ b/stratosphere-java/src/main/java/eu/stratosphere/api/java/DataSet.java
@@ -25,6 +25,7 @@ import eu.stratosphere.api.java.functions.GroupReduceFunction;
 import eu.stratosphere.api.java.functions.KeySelector;
 import eu.stratosphere.api.java.functions.MapFunction;
 import eu.stratosphere.api.java.functions.ReduceFunction;
+import eu.stratosphere.api.java.io.CollectionOutputFormat;
 import eu.stratosphere.api.java.io.CsvOutputFormat;
 import eu.stratosphere.api.java.io.PrintingOutputFormat;
 import eu.stratosphere.api.java.io.TextOutputFormat;
@@ -35,6 +36,9 @@ import eu.stratosphere.api.java.tuple.Tuple;
 import eu.stratosphere.api.java.typeutils.InputTypeConfigurable;
 import eu.stratosphere.api.java.typeutils.TypeInformation;
 import eu.stratosphere.core.fs.Path;
+
+import java.io.Serializable;
+import java.util.Collection;
 
 /**
  *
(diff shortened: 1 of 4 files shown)


stratosphere-java/src/main/java/eu/stratosphere/api/java/io/CollectionOutputFormat.java
```diff
@@ -0,0 +1,71 @@
...
+	public void configure(Configuration parameters) {
+	}
+
+
+	@Override
+	public void open(int taskNumber, int numTasks) throws IOException {
+		try {
+			this.taskResult = (Collection<T>) resultWrapper.getClass().newInstance();
+		} catch (InstantiationException e) {
+			e.printStackTrace();
```
rmetzger: Can you re-throw the exceptions as IOExceptions?

stratosphere-java/src/main/java/eu/stratosphere/api/java/DataSet.java
```diff
@@ -191,7 +195,10 @@ public ExecutionEnvironment getExecutionEnvironment() {
 	public void writeAsText(String filePath) {
 		output(new TextOutputFormat<T>(new Path(filePath)));
 	}
-	
+
+	public <X extends Serializable> void writeAsCollection(Collection<X> out) {
```
fhueske: Shouldn't this be `writeAsCollection(Collection<T> out)` to ensure type safety of the collection?

qmlmoon: Yeah, it should be. But to copy the record object in `writeRecord(T record)`, I used `SerializationUtil.clone`

fhueske: You can get a type preserving serializer from TypeInformation of DataSet and set it in the constructor of ColloctionOutputFormat.

stratosphere-java/src/main/java/eu/stratosphere/api/java/io/CollectionOutputFormat.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.Collection;
+
+/**
+ *  An output format that writes record into collection
+ */
+public class CollectionOutputFormat<T> implements OutputFormat<T> {
+
+	private static final long serialVersionUID = 1L;
+
+	private static Object resultWrapper;   //since generic collection is not allowed to be static
```
StephanEwen: I think you can make this a `Collection<?>`.

stratosphere-java/src/main/java/eu/stratosphere/api/java/io/CollectionOutputFormat.java
```diff
@@ -0,0 +1,77 @@
...
+	public void writeRecord(T record) throws IOException {
+		T recordCopy = this.typeSerializer.createInstance();
+		this.typeSerializer.copy(record, recordCopy);
+		this.taskResult.add(recordCopy);
+	}
+
+
+	@Override
+	public void close() throws IOException {
+		Collection<T> result = (Collection<T>) this.resultWrapper;
```
StephanEwen: I think this is problematic in the parallel case. If many parallel tasks are in `close()`at the same time, they will concurrently add elements, which causes a `ConcurrentModificationException`.

You need to synchronize around the target list here.

StephanEwen:
It is a good start. There are two things I would like to improve:
- I don't think we should have a dedicated method for this on `DataSet`. because this will only work for local testing. I would prefer to use it only through `result.output(new CollectionOutputFormat(list))`.
- This will not work for jobs with multiple sinks, because you have only one static reference. If you had two CollectionOutputFormats in the same program, both would get the union of the results. I think you can work around that by giving each CollectionOutputFormat a generated id and have a static map of collections, rather than only one static field.

stratosphere-java/src/main/java/eu/stratosphere/api/java/io/CollectionOutputFormat.java
```diff
@@ -0,0 +1,77 @@
...
+
+
+	@Override
+	public void configure(Configuration parameters) {
+	}
+
+
+	@Override
+	public void open(int taskNumber, int numTasks) throws IOException {
+		try {
```
StephanEwen: I think you can just create a new `ArrayList` here. The collection in which you gather the elements needs not be of the same type as the collection into which you out the elements at the end. That spares you the reflection instantiation, which may fail if no public nullary constructor exists.

qmlmoon:
`CollectionOutputFormat` needs a `TypeSerializer` from `DataSet` to make a copy of record. To remove the method on `DataSet`, I would do something like `CollectionOutputFormat<T extends Tuple>` to use `SerializationUtil.clone()`, which forces the element of the collection to be a tuple. But it's not good for type safty as Fabian pointed.

StephanEwen:
I see. But I don't think the workaround can be to add a method to data set for every possible output format.
Let's add an interface called `TypeConfigurable` or so. Then the output() function can call a method on corresponding output formats and pass the type.

qmlmoon:
What about to use the collection like `result.output(new CollectionOutputFormat(list, result.getType().createSerializer))` ,just specify the serialiser manually

stratosphere-java/src/main/java/eu/stratosphere/api/java/io/CollectionOutputFormat.java
```diff
@@ -0,0 +1,91 @@
...
+
+	@Override
+	public void open(int taskNumber, int numTasks) throws IOException {
+		this.taskResult = new ArrayList<T>();
+	}
+
+	@Override
+	public void writeRecord(T record) throws IOException {
+		T recordCopy = this.typeSerializer.createInstance();
+		this.typeSerializer.copy(record, recordCopy);
```
StephanEwen: I think this needs to be `recordCopy = this.typeSerializer.copy(record, recordCopy);`

stratosphere-java/src/main/java/eu/stratosphere/api/java/io/CollectionOutputFormat.java
```diff
@@ -0,0 +1,91 @@
...
+	public void close() throws IOException {
+		synchronized (this.resultWrapper) {
+			Collection<T> result = (Collection<T>) this.resultWrapper.get(this.id);
+			result.addAll(this.taskResult);
+		}
+	}
+
+	@Override
+	public void setInputType(TypeInformation<?> type) {
+		this.typeSerializer = (TypeSerializer<T>)type.createSerializer();
```
StephanEwen: Can you suppress the "unchecked" warnings here?

StephanEwen:
Looks nice to me :-)

I think we should rename it, just to make sure that it is obvious that this thing is only intended for local executor use. Maybe call it `LocalCollectionOutputFormat` or so.

fhueske:
Good to merge, thanks! :+1:

fhueske:
Will merge this now.

PR #627, opened 2014-03-25
Termination Criterion and PageRank for the new Java API

markus-h:
I added the termination criterion for BulkIterations to new new Java API.
Additionally I wrote the SimplePageRank algorithm in the new API and added it as testcase.

### DIFF
--- a/stratosphere-compiler/src/main/java/eu/stratosphere/compiler/postpass/JavaApiPostPass.java
+++ b/stratosphere-compiler/src/main/java/eu/stratosphere/compiler/postpass/JavaApiPostPass.java
@@ -81,6 +81,12 @@ public class JavaApiPostPass implements OptimizerPostPass {
 			if (iterationNode.getRootOfStepFunction() instanceof NAryUnionPlanNode) {
 				throw new CompilerException("Optimizer cannot compile an iteration step function where next partial solution is created by a Union node.");
 			}
+			
+			// traverse the termination criterion for the first time. create schema only, no utilities. Needed in case of intermediate termination criterion
+			if (iterationNode.getRootOfTerminationCriterion() != null) {
+				SingleInputPlanNode addMapper = (SingleInputPlanNode) iterationNode.getRootOfTerminationCriterion();
+				traverseChannel(addMapper.getInput());
+			}
 
 			PlanBulkIterationOperator<?> operator = (PlanBulkIterationOperator<?>) iterationNode.getPactContract();
 
--- a/stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/pagerank/SimplePageRank.java
+++ b/stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/pagerank/SimplePageRank.java
@@ -1,0 +1,155 @@
+/***********************************************************************************************************************
+ *
+ * Copyright (C) 2010-2013 by the Stratosphere project (http://stratosphere.eu)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ **********************************************************************************************************************/
+package eu.stratosphere.example.java.pagerank;
+
+
+import java.util.ArrayList;
+import java.util.Iterator;
+
+import eu.stratosphere.api.java.DataSet;
+import eu.stratosphere.api.java.ExecutionEnvironment;
+import eu.stratosphere.api.java.IterativeDataSet;
+import eu.stratosphere.api.java.functions.FlatMapFunction;
+import eu.stratosphere.api.java.functions.GroupReduceFunction;
+import eu.stratosphere.api.java.functions.JoinFunction;
+import eu.stratosphere.api.java.functions.ReduceFunction;
+import eu.stratosphere.api.java.tuple.Tuple2;
+import eu.stratosphere.util.Collector;
+
+public class SimplePageRank {
+	
+	public static final class OutgoingEdgeCounter extends GroupReduceFunction<Tuple2<Long, Long>, Tuple2<Long, Long[]>> {
+
+		private static final long serialVersionUID = 1L;
+		
+		@Override
+		public void reduce(Iterator<Tuple2<Long, Long>> values,
+				Collector<Tuple2<Long, Long[]>> out) throws Exception {
+
+			long id = 0;
+			ArrayList<Long> neighbors = new ArrayList<Long>();
+			
+			while(values.hasNext()) {
+				Tuple2<Long, Long> n = values.next();
+				id = n.f0;
+				neighbors.add(n.f1);
+			}
+			
+			out.collect(new Tuple2<Long, Long[]>(id, neighbors.toArray(new Long[neighbors.size()])));
+		}
+
+		
+	}
+	
+	public static final class JoinVertexWithEdgesMatch extends FlatMapFunction<Tuple2<Tuple2<Long, Double>, Tuple2<Long, Long[]>>, Tuple2<Long, Double>> {
+
+		private static final long serialVersionUID = 1L;
+
+		@Override
+		public void flatMap(
+				Tuple2<Tuple2<Long, Double>, Tuple2<Long, Long[]>> value,
+				Collector<Tuple2<Long, Double>> out) throws Exception {
+
+			double rankToDistribute = value.f0.f1 / (double) (value.f1.f1.length);
+				
+			for(int i = 0; i < value.f1.f1.length; i++) {
+				out.collect(new Tuple2<Long, Double>(value.f1.f1[i], rankToDistribute));
+			}
+			
+		}
+	
+	}
+	
+	public static final class AggregatingReduce extends ReduceFunction<Tuple2<Long, Double>> {
+
+		private static final long serialVersionUID = 1L;
+
+		@Override
+		public Tuple2<Long, Double> reduce(Tuple2<Long, Double> rec1,
+				Tuple2<Long, Double> rec2) throws Exception {
+
+			double rankSum = rec1.f1;
+			rankSum += rec2.f1;
+			
+			return new Tuple2<Long, Double>(rec1.f0, rankSum);
+		}
+		
+	}
+	
+	public static final class JoinOldAndNew extends JoinFunction<Tuple2<Long, Double>, Tuple2<Long, Double>, Tuple2<Long, Double>> {
+
+		private static final long serialVersionUID = 1L;
+
+		@Override
+		public Tuple2<Long, Double> join(Tuple2<Long, Double> first,
+				Tuple2<Long, Double> second) {
+			
+			double epsilon = 0.05;
+			double criterion = first.f1 - second.f1;
+			
+			if(Math.abs(criterion) > epsilon)
+			{
+				return second;
+			}
+			
+			return null;
+		}
+		
+	}
+	
+	public static void main(String[] args) throws Exception {
+		
+		if (args.length < 5) {
+			System.err.println("Usage: SimpePageRank <DOP> <pageWithRankInputPath> <adjacencyListInputPath> <outputPath> <numIterations>");
+			return;
+		}
+		
+		final int dop = Integer.parseInt(args[0]);
+		final String pageWithRankInputPath = args[1];
+		final String adjacencyListInputPath = args[2];
+		final String outputPath = args[3];
+		final int maxIterations = Integer.parseInt(args[4]);
+		
+		run(dop, pageWithRankInputPath, adjacencyListInputPath, outputPath, maxIterations, true);
+		
+	}
+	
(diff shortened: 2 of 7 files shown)


StephanEwen:
Looks good to me.

Concerning the dummy-serializer after the mapper that counts records and
updates the aggregators: Right now, using a generic serializer for Object
instantiates an Avro serializer, creates a reflection schema, ... Having a
simple NullSerializer would be more elegant and efficient there, but that
is a detail optimization.

markus-h:
Thanks for your review! I added a DummySerializerFactory for that purpose.

stratosphere-compiler/src/main/java/eu/stratosphere/compiler/postpass/JavaApiPostPass.java
```diff
@@ -63,6 +64,13 @@ else if (node instanceof BulkIterationPlanNode) {
 			if (iterationNode.getRootOfStepFunction() instanceof NAryUnionPlanNode) {
 				throw new CompilerException("Optimizer cannot compile an iteration step function where next partial solution is created by a Union node.");
 			}
+			
+			// traverse the termination criterion for the first time. create schema only, no utilities. Needed in case of intermediate termination criterion
+			if (iterationNode.getRootOfTerminationCriterion() != null) {
+				SingleInputPlanNode addMapper = (SingleInputPlanNode) iterationNode.getRootOfTerminationCriterion();
```
StephanEwen, line 87: I am bit confused here and do not understand how this can work.

The "addMapper" is the attached mapper that increments the aggregator for each element that it finds. As such, it really needs to be executed in the runtime. You give that mapper's input the null serializer. So no record can ever arrive at the mapper?

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/pagerank/SimplePageRank.java
```diff
@@ -0,0 +1,158 @@
...
+import eu.stratosphere.util.Collector;
+
+public class SimplePageRank {
+	
+	public static final class JoinVertexWithEdgesMatch extends CoGroupFunction<Tuple2<Long, Double>, Tuple2<Long, Long>, Tuple2<Long, Double>> {
+
+		private static final long serialVersionUID = 1L;
+
+		@Override
+		public void combineFirst(Iterator<Tuple2<Long, Double>> records,
```
StephanEwen: Can you add empty implementations for the combine methods to the CoGroupFunction? Otherwise every cogroup implementer needs to write (empty) combiners.

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/pagerank/SimplePageRank.java
```diff
@@ -0,0 +1,158 @@
...
+
+		@Override
+		public void combineSecond(Iterator<Tuple2<Long, Long>> records,
+				Collector<Tuple2<Long, Long>> out) throws Exception {
+		}
+
+		@Override
+		public void coGroup(Iterator<Tuple2<Long, Double>> first,
+				Iterator<Tuple2<Long, Long>> second,
+				Collector<Tuple2<Long, Double>> out) throws Exception {
```
StephanEwen, line 63: You can always remove the `throws Exception` if the method throws no Exception.

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/pagerank/SimplePageRank.java
```diff
@@ -0,0 +1,158 @@
...
+			 
+			for(Long l : neighbours) {
+				out.collect(new Tuple2<Long, Double>(l, rankToDistribute));
+			}
+			
+		}
+	
+	}
+	
+	public static final class AggregatingReduce extends GroupReduceFunction<Tuple2<Long, Double>, Tuple2<Long, Double>> {
```
StephanEwen: Can this be a regular reduce, rather than a group reduce? Reduce is preferable over GroupReduce because it is faster. Implicitly combinable and it fits execution via hash aggregation.

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/pagerank/SimplePageRank.java
```diff
@@ -0,0 +1,158 @@
...
+				rankSum += rec.f1;
+			}
+			
+			out.collect(new Tuple2<Long, Double>(rec.f0, rankSum));
+			
+		}
+		
+	}
+	
+	public static final class JoinOldAndNew extends JoinFunction<Tuple2<Long, Double>, Tuple2<Long, Double>, Tuple2<Long, Double>> {
```
StephanEwen, line 91: Would it look nicer if you used a default join and a filter?

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/pagerank/SimplePageRank.java
```diff
@@ -0,0 +1,158 @@
...
+		if (args.length < 5) {
+			System.err.println("Usage: SimpePageRank <DOP> <pageWithRankInputPath> <adjacencyListInputPath> <outputPath> <numIterations>");
+			return;
+		}
+		
+		final int dop = Integer.parseInt(args[0]);
+		final String pageWithRankInputPath = args[1];
+		final String adjacencyListInputPath = args[2];
+		final String outputPath = args[3];
+		final int numIterations = Integer.parseInt(args[4]);
```
StephanEwen: Can you call this `maxIterations`, as the actual number is rather determined by the termination criterion?

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/pagerank/SimplePageRank.java
```diff
@@ -0,0 +1,158 @@
...
+		
+		run(dop, pageWithRankInputPath, adjacencyListInputPath, outputPath, numIterations, true);
+		
+	}
+	
+	public static void run(int dop, String pageWithRankInputPath, String adjacencyListInputPath, String outputPath, int numIterations, boolean terminationCriterion) throws Exception {
```
StephanEwen: If you make this method a method that accepts DataSets with the inputs, rather than path strings, then you can run this in the tests with with collection sources, rather than temp files.

Similarly, if you return a DataSet with the result, rather than writing a file, you can check the result directly from a collection sink, rather than from result temp files.

stratosphere-java/src/main/java/eu/stratosphere/api/java/typeutils/runtime/DummySerializerFactory.java
```diff
@@ -0,0 +1,41 @@
+/***********************************************************************************************************************
```
StephanEwen: I am surprised that this works. If the factory is ever used, it will at some point be used to create a serializer. It then returns null and causes a NullPointerException. I think you really need a NullSerializer (or discarding serializer), rather than a factory that returns null.

stratosphere-compiler/src/main/java/eu/stratosphere/compiler/postpass/JavaApiPostPass.java
markus-h, line 87: I have to confess that I did not invest that much thinking in these lines. I just added the serializer where it was missing and it worked. (Actually it is almost the same as the current solution for the record API).
There are records passed to the mapper, in fact that is exactly the criterion for termination we are checking for. Can it be connected to the fact that the mapper is chained anyway and there is no serialization at all?

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/pagerank/SimplePageRank.java
markus-h: I wondered about this before and thought there would be some reason for the fact that these methods always need to be implemnted manually. Of couse I can add a default implementation in CoGroupFunction.

stratosphere-compiler/src/main/java/eu/stratosphere/compiler/postpass/JavaApiPostPass.java
StephanEwen, line 87: The chaining sounds like the solution. Lucky that his bug never occurred.
This should be fixed nonetheless for the new api. The null serializer would
make sense for after the mapper, but I think we never needed one there
anyways. Before the mapper, we need a proper serializer.

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/pagerank/SimplePageRank.java
markus-h: Yes, I think it can be replaced. I will do it tomorrow.

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/pagerank/SimplePageRank.java
markus-h, line 91: I'm not sure, for me it is not much of a difference. Has somebody else an opinion on that?

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/pagerank/SimplePageRank.java
markus-h: I tried to stick to the actual existing testing framework for the new Java API (JavaProgramTestBase, see eu.stratosphere.test.javaApiOperators).
When I change to DataSets I think I would have to write new logic for comparing the result instead of using the existing method for that.

stratosphere-java/src/main/java/eu/stratosphere/api/java/typeutils/runtime/DummySerializerFactory.java
markus-h: I think this issue is connected with your first comment. Can it be the case that because of chaining there is in no case a serialization and so no serializer is created anyway?

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/pagerank/SimplePageRank.java
```diff
@@ -0,0 +1,142 @@
...
+	public static final class JoinVertexWithEdgesMatch extends CoGroupFunction<Tuple2<Long, Double>, Tuple2<Long, Long>, Tuple2<Long, Double>> {
+
+		private static final long serialVersionUID = 1L;
+
+		@Override
+		public void coGroup(Iterator<Tuple2<Long, Double>> first,
+				Iterator<Tuple2<Long, Long>> second,
+				Collector<Tuple2<Long, Double>> out) {
+
+			if(!first.hasNext() || !second.hasNext())
```
StephanEwen: This code here really looks like you are imitating an inner join. You throw out cases where either side has no input. You manually build the cross produce within a group. This is exactly what the join would do for you.

Why not use a join directly?

StephanEwen: If you need to know what the number of outgoing edges per node is, you can add a reduce over the edges where you build the adjacency list: (source-id, targat-ids[])

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/pagerank/SimplePageRank.java
```diff
@@ -0,0 +1,142 @@
...
+			 
+			for(Long l : neighbours) {
+				out.collect(new Tuple2<Long, Double>(l, rankToDistribute));
+			}
+			
+		}
+	
+	}
+	
+	public static final class AggregatingReduce extends ReduceFunction<Tuple2<Long, Double>> {
```
StephanEwen, line 75: The system has built-in aggregations. If you use them, the code becomes simpler and more readable.

Also, where do you apply the dampening factor formula?

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/pagerank/SimplePageRank.java
markus-h: Yes it is possible to build an adjacency list with a reduce. But I don't see how I can replace the coGroup with a join. The new JoinFunction interface does not provide a collector as output but can only emit one tuple. I think we need an additional "FlatJoinFunction" or something with an collector. I opened an issue to discuss this: #662

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/pagerank/SimplePageRank.java
markus-h, line 75: What do you mean with dampening factor formula? I oriented on the old PageRank examples and cannot find such a factor there.

markus-h:
I have redone the PageRank example to use arrays. This change relies on #670, without it the tests will fail.

markus-h:
I rebased to current master. Lets see if Travis does pass now.

StephanEwen:
Merged in 4244e6d8b11dcaf80fb2a3e9897f39ac24452852

Issue #626, opened 2014-03-25
Set write mode in new Japi


markus-h:
I'm writing the output of a job using [...].writeAsCsv(outputPath);
When the file specified by the outputPath does not exist I get the following error:
 _Stream to output file could not be opened: Opening the file output stream failed: File or directory already exists. Existing files and directories are not overwritten in NO_OVERWRITE mode. Use OVERWRITE mode to overwrite existing files and directories._

As far as I can see I can only set the the writeMode in the FileOutputFormat class. But this class is instantiated in the DataSet class. Is there a clean way to switch to OVERWRITE in the new Japi that I have overseen? If not, what would (in your opinion) be a clean way to set this configuration in the new API?

StephanEwen:
Good point. I think in these cases you would need to go with `dataset.write(new CsvOutputFormat(), filepath, writeMode)`

The writeMode parameter does not currently exist for the write(...) function, but we should add it in my opinion. And remove the redundant write(...) variant that takes a path as parameter.

Another option is to return the DataSink from the `write(...)` or `writeAsCsv(...)`method. That would allow you to fluently attach further options, like `setOverwrite()`.

markus-h:
I think I like the first no fluid variant better. I will create a pull request shortly.

markus-h:
Implemented in #630.

StephanEwen:
Fixed in #630 via 5cc81cdd1f71c740669c7c4284524499d6e8617a

PR #630, opened 2014-03-26
small change to write function for new java api to set the write mode

markus-h:
See #626


StephanEwen:
The way this pull request changes the behavior is that the system always uses the `NO_OVERWRITE` mode unless something else is specified. The configured default write mode is therefore always overridden and never used.

I think the method that takes no write mode parameter should not set a write mode. If the write mode is not set, the configured default write mode (in stratosphere-conf.yml) of the executing environment is used. If you throw it to the cluster, then the cluster decided whether it accepts overwriting files.

markus-h:
Thanks for your feedback, Stephan! I changed the implementation accordingly.

StephanEwen:
Merged in 5cc81cdd1f71c740669c7c4284524499d6e8617a

PR #632, opened 2014-03-26
Cost estimator

### LINKED ISSUE #613: Cost estimation problem for branched plans
I encountered the problem of negative cumulative costs for a plan which contained branched nodes. The problem is that the costs for branched nodes are not correctly calculated if one branched node x is contained in a cycle of another branched node y. Assuming that y is closed first, then closing of x will calculate too high double counted costs. To illustrate this case consider the following flow:

<pre>
     A ---- D
   /  \   /  \
SRC     \   SINK
   \  /   \  /
     B ---- C
</pre>

Assuming that the costs of all nodes except for SRC is 0 and the cost of SRC is 1, then SINK should have cumulative costs of 1 at the end. But executing the algorithm will calculate the following sequence.

SRC = 1
A = SRC = 1
B = SRC = 1
D = A + B - SRC (as a joined branching node) = 1
C = A + B - SRC = 1
SINK = D + C - A (as a joined branching node) - B (as a joined branching node) = 0

Thus the problem originates from a wrong accounting of branching nodes. A possible solution for this problem could be to split the predecessor costs according to the out degree of the predecessor node. By doing this, we wouldn't need to subtract double counted costs. But I'm not aware of any side effects this could have....

tillrohrmann:
This fixes #613. First the Costs members are now doubles to support divisions. The cumulative costs of a node are now computed in the following way: Costs for the node + adjusted costs of all predecessor nodes, where adjusted costs means that the costs of the predecessor node is divided by its out degree. This achieves that the costs of a node is propagated equally among its successors.


StephanEwen:
I like the principle idea of dividing costs. I am a bit skeptical concerning the switch to double actually. 

With the large numbers that we represent in the doubles, the domain starts to become very sparse and it may be that computations like `costs += 1` do not change the value, because that specific value could not be represented. We may need to do such computations to introduce biases, though that's not clear yet.

On the other hand, the same would occur when we divide the integer valued costs by the number of outputs. Although I guess that is okay, as we never compare plans across different candidates at such "branching points".

All in all I think that representing the costs as `long` is a bit safer right now.

What do you think?

tillrohrmann:
You are right that the problem of absorption might occur if the costs get really huge. And you are also right that using `long` would not be a problem if one reasons about one group of plans with the same interesting properties. Only the best plan is kept at the branching point and thus prior to the actual division of the costs. 

But what about the case where one has two groups of plans with different interesting properties at a branching node `x`. Let's assume their costs are `p` for interesting properties A and `p+1` for interesting properties B. Let's assume the node costs at the successor node `y` is `c` for both interesting properties A and B. Furthermore `x` has a branching degree of 2. Then the cumulative costs of `y` for A would be `c +  p/2` and for B would be `c+(p+1)/2`. Then you couldn't distinguish anymore which plan is the cheapest. (I hope that I got the pruning mechanism right....) I see the problem of this scenario that there cannot be two groups of plans at the end of the getAlternativePlans method, but I haven't found the source code location yet where this is assured. Does it have any effect if one doesn't know anymore which plan is the cheapest?

Another problem might be that other programmers could stumble upon this design decision wondering why one can forget about the decimal fraction. This can of course be alleviated by some explanatory comments.

How big does the costs actually get? Using doubles, one has roughly 16 digits until the problem with `costs+1` occurs. A long is of course better with roughly 20 digits but if one talks about numbers of this magnitude the jump from doubles to longs might not be sufficient and one could end up with a overflow when using longs.

StephanEwen:
During pruning, plans are not compared across different candidates at branching points. Branch points act like an additional 'super' interesting property.

Plans across candidates at branch points are only compared at the point where the branch has been closed. At that point, it may happen that plan costs appear equal due to rounding loss.

I think the problem will occur in both ways, though at different points. It will most likely not matter for the operator costs themselves, only for small values we add to introduce a bias.

The costs actually get fairly big. Especially the heuristic costs (see cost estimator classes).

StephanEwen:
I think we can go ahead with this for now.

tillrohrmann:
Do you mean going ahead with `double` or `long`?

StephanEwen:
Double
Am 30.03.2014 23:23 schrieb "Till Rohrmann" notifications@github.com:

StephanEwen:
Merged in 84c2bcb0d8cb4c2ce91f93686cf646a786a3fc60

PR #635, opened 2014-03-27
Add PojoTypeInformation, PojoComparator and Expressions for Keys

aljoscha:
This enables code like:

``` java
class WC {
    public String word;
    public int count;
}
...
DataSet<WC> result = tokenized
        .groupBy("word")
        .reduce(new ReduceFunction<WC>() {
            public WC reduce(WC value1, WC value2) {
                return new WC(value1.word, value1.count + value2.count);
            }
        });
```

### DIFF
--- a/stratosphere-addons/avro/src/main/java/eu/stratosphere/api/java/io/AvroInputFormat.java
+++ b/stratosphere-addons/avro/src/main/java/eu/stratosphere/api/java/io/AvroInputFormat.java
@@ -17,6 +17,7 @@ package eu.stratosphere.api.java.io;
 
 import java.io.IOException;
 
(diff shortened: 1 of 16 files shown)


rmetzger:
Amazing! Whats the plan regarding unit tests for that?

aljoscha:
We should have some, but this is still an early implementation. I could add a WordCountITCase and a RelQuery using this feature with an ITCase.

StephanEwen:
I think the type mapping also needs quite a few "Unit" tests, meaning
testing different scenarios independent of program execution.

aljoscha:
Make new pull request from pojo-feature branch.

PR #636, opened 2014-03-27
maven checkstyle refactoring

alexff91:
There's only stratosphere-java refactored for now.

PR #640, opened 2014-03-31
added annotation support for new java api

skunert:
The operators now read the annotations and wrap them in semanticproperties. This implementation should work fine for operator implementation which use tuples.

### DIFF
--- a/stratosphere-java/src/main/java/eu/stratosphere/api/java/operators/translation/PlanCogroupOperator.java
+++ b/stratosphere-java/src/main/java/eu/stratosphere/api/java/operators/translation/PlanCogroupOperator.java
@@ -15,8 +15,14 @@
 package eu.stratosphere.api.java.operators.translation;
 
 import eu.stratosphere.api.common.functions.GenericCoGrouper;
+import eu.stratosphere.api.common.operators.DualInputSemanticProperties;
+import eu.stratosphere.api.common.operators.SingleInputSemanticProperties;
 import eu.stratosphere.api.common.operators.base.CoGroupOperatorBase;
+import eu.stratosphere.api.common.operators.util.UserCodeObjectWrapper;
+import eu.stratosphere.api.common.operators.util.UserCodeWrapper;
 import eu.stratosphere.api.java.functions.CoGroupFunction;
+import eu.stratosphere.api.java.functions.GroupReduceFunction;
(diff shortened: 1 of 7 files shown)


stratosphere-java/src/main/java/eu/stratosphere/api/java/operators/translation/PlanCogroupOperator.java
```diff
@@ -35,6 +41,10 @@ public PlanCogroupOperator(
 		this.inType1 = inType1;
 		this.inType2 = inType2;
 		this.outType = outType;
+		
+		UserCodeWrapper<CoGroupFunction<IN1, IN2, OUT>> tmp = new UserCodeObjectWrapper<CoGroupFunction<IN1, IN2, OUT>>(udf);
+        DualInputSemanticProperties sp = FunctionAnnotation.readDualConstantAnnotations(tmp);
```
fhueske, line 46: Can you replace the spaces by tabs for indention?
We will have codestyle checks in place once PR #636 is merged.
Please check also the other lines you added.

stratosphere-java/src/main/java/eu/stratosphere/api/java/operators/translation/PlanCogroupOperator.java
```diff
@@ -35,6 +41,10 @@ public PlanCogroupOperator(
 		this.inType1 = inType1;
 		this.inType2 = inType2;
 		this.outType = outType;
+		
+		UserCodeWrapper<CoGroupFunction<IN1, IN2, OUT>> tmp = new UserCodeObjectWrapper<CoGroupFunction<IN1, IN2, OUT>>(udf);
```
fhueske, line 45: You can get a UserCodeWrapper for `udf` by calling `this.getUserCodeWrapper()` and don't need the `tmp` object.
It is initialized and set by the super() constructor call.
Same applies (probably) for the other operators.

fhueske:
What happens if the annotations are incorrectly used for UDFs with non-Tuple types? 
Should we allow annotations only on UDFs with Tuple types, check that at program assembly time, give a warning, and ignore the annotations?

fhueske:
Hi @skunert, what happened to this PR?
Are you reworking it?

skunert:
Yeah, I talked to stephan and am reworking it at this moment.

PR #638, opened 2014-03-28
First Version of Spargel (Pregel style API) for Java

StephanEwen:
To get a feel of the api, check out this example (connected components)

```
public class SpargelConnectedComponents {

    public static void main(String[] args) throws Exception {
        ExecutionEnvironment env = ExecutionEnvironment.getExecutionEnvironment();

        DataSet<Long> vertexIds = env.generateSequence(0, 10);
        DataSet<Tuple2<Long, Long>> edges = env.fromElements(new Tuple2<Long, Long>(0L, 2L), new Tuple2<Long, Long>(2L, 4L), new Tuple2<Long, Long>(4L, 8L),
                                                            new Tuple2<Long, Long>(1L, 5L), new Tuple2<Long, Long>(3L, 7L), new Tuple2<Long, Long>(3L, 9L));

        DataSet<Tuple2<Long, Long>> initialVertices = vertexIds.map(new IdAssigner());

        DataSet<Tuple2<Long, Long>> result = initialVertices.runOperation(SpargelIteration.withPlainEdges(edges, new CCUpdater(), new CCMessager(), 100));

        result.print();
        env.execute("Spargel Connected Components");
    }


    public static final class CCUpdater extends VertexUpdateFunction<Long, Long, Long> {
        @Override
        public void updateVertex(Long vertexKey, Long vertexValue, MessageIterator<Long> inMessages) {
            long min = Long.MAX_VALUE;
            for (long msg : inMessages) {
                min = Math.min(min, msg);
            }
            if (min < vertexValue) {
                setNewVertexValue(min);
            }
        }
    }

    public static final class CCMessager extends MessagingFunction<Long, Long, Long, NullValue> {
        @Override
        public void sendMessages(Long vertexId, Long componentId) {
            sendMessageToAllNeighbors(componentId);
        }
    }

    public static final class IdAssigner extends MapFunction<Long, Tuple2<Long, Long>> {

        @Override
        public Tuple2<Long, Long> map(Long value) throws Exception {
            return new Tuple2<Long, Long>(value, value);
        }
    }
}
```

From https://github.com/StephanEwen/stratosphere/blob/spargel/stratosphere-addons/spargel/src/main/java/eu/stratosphere/spargel/java/examples/SpargelConnectedComponents.java

Issue #607, opened 2014-03-18
Delta iterations in new Java API


rmetzger:

markus-h:
I'm on it.

StephanEwen:
I have implemented a port of Spargel to the new java API. It does not use the new Java API classes for delta iterations, but rather directly instantiates the `eu.stratosphere.api.common` classes in the background.

It does however need the java api postpass to be able to handle delta iteration plan operators. Any chance you have gotten to that part already?

markus-h:
No, sorry, not yet. But it should be very similiar to BulkIterations with
termination criterion, so you can try to do a quick copy paste.

2014-03-28 0:50 GMT+01:00 Stephan Ewen notifications@github.com:

markus-h:
This issue can be closed.
See #643 merged in 74ce8d3

PR #643, opened 2014-04-01
Delta Iterations for new Java API

markus-h:
I added Delta Iterations to the new Java API (#607). Moreover I implemented a simple Delta PageRank Example and created a testcase with it. The example also shows how the new syntax for delta iterations looks like.

My branch is based on PullRequest #638, since it already included some of the compiler post pass logic for delta iterations.


StephanEwen:
Merged in 74ce8d3b3a202aadb59b843582068c5ed21ca4d9

vasia:
Hey,
the deltas initialization can now actually be included into the same program (when I first wrote this delta version, that wasn't possible, so I would first run one separate bulk iteration to get the initial deltas input). 
So, now both the SimpleDeltaPageRank example and the test case can be simplified to receive only the initial ranks and the edges as input and not the deltas explicitly.
Not sure if you think it's important, but I could change it if you want :)

StephanEwen:
It would make a nice example, but it is not critical. Up to you ;-)

PR #646, opened 2014-04-01
Allow .yml file extension for config files (fixes #645)

### LINKED ISSUE #645: Add .yml as config file extension
Currently, `GlobalConfiguration` searches the config dir for files ending in XML or YAML.

We should also allow YML as file extension for YAML files as it is quite common.

uce:


stratosphere-core/src/main/java/eu/stratosphere/configuration/GlobalConfiguration.java
```diff
@@ -19,6 +19,7 @@
 import java.io.FilenameFilter;
 import java.io.IOException;
 import java.io.InputStreamReader;
+import java.util.Arrays;
```
rmetzger: My brain-based checkstyle plugin sees an unused import here.

rmetzger:
Good to merge (once the issue is resolved)

stratosphere-core/src/main/java/eu/stratosphere/configuration/GlobalConfiguration.java
```diff
@@ -658,10 +655,20 @@ private void includeConfigurationInternal(final Configuration conf) {
...
+		return filterFilesBySuffix(dir, new String[] { suffix });
+	}
+
+	private static File[] filterFilesBySuffix(final File dir, final String[] suffixes) {
 		return dir.listFiles(new FilenameFilter() {
 			@Override
 			public boolean accept(final File dir, final String name) {
-				return dir.equals(dir) && name != null && name.endsWith(suffix);
+				for (String suffix : suffixes) {
+					if (dir.equals(dir) && name != null && name.endsWith(suffix)) {
```
StephanEwen: This line seems weird. `dir.equals(dir)` is always true. I assume one `dir` is supposed to refer to the parameter of the closure, rather then the parameter in the function? I think you have to rename one of them to make them work properly.

uce: Thanks for catching this one! I missed it while copying from the old methods.

StephanEwen:
Merged in 1c1be8cb051f3b97e43a4b6d36f0bc35b459d63b

PR #647, opened 2014-04-01
JDBC In-Outputformat for new API

zentol:
I ported the JDBC Formats to the new API. (both formats, tests for each and an example).

The outputformat only supports a number of types, currently these

```
private enum supportedTypes {
        Boolean,
        Byte,
        Short,
        Integer,
        Long,
        String,
        Float,
        Double
    }
```

Are there others i should add, like ValueTypes?

### DIFF
--- a/stratosphere-addons/jdbc/src/main/java/eu/stratosphere/api/java/io/jdbc/JDBCInputFormat.java
+++ b/stratosphere-addons/jdbc/src/main/java/eu/stratosphere/api/java/io/jdbc/JDBCInputFormat.java
@@ -1,0 +1,336 @@
+/***********************************************************************************************************************
+ *
+ * Copyright (C) 2010-2013 by the Stratosphere project (http://stratosphere.eu)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ **********************************************************************************************************************/
+package eu.stratosphere.api.java.io.jdbc;
+
+import eu.stratosphere.api.common.io.InputFormat;
+import eu.stratosphere.api.common.io.statistics.BaseStatistics;
+import eu.stratosphere.api.java.tuple.Tuple;
+import eu.stratosphere.configuration.Configuration;
+import eu.stratosphere.core.io.GenericInputSplit;
+import eu.stratosphere.core.io.InputSplit;
+import eu.stratosphere.types.NullValue;
+import java.io.IOException;
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.ResultSet;
+import java.sql.ResultSetMetaData;
+import java.sql.SQLException;
+import java.sql.Statement;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+/**
+ * InputFormat to read data from a database and generate tuples.
+ * The InputFormat has to be configured using the supplied InputFormatBuilder.
+ * 
+ * @param <OUT>
+ * @see Tuple
+ * @see DriverManager
+ */
+public class JDBCInputFormat<OUT extends Tuple> implements InputFormat<OUT, InputSplit> {
+    private static final long serialVersionUID = 1L;
+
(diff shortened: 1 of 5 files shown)


stratosphere-addons/jdbc/src/main/java/eu/stratosphere/api/java/io/jdbc/JDBCInputFormat.java
```diff
@@ -0,0 +1,336 @@
...
+     * @throws java.io.IOException
+     */
+    @Override
+    public OUT nextRecord(OUT tuple) throws IOException {
+        try {
+            resultSet.next();
+            int column_count = resultSetMetaData.getColumnCount();
+
+            for (int pos = 0; pos < column_count; pos++) {
+                int type = resultSetMetaData.getColumnType(pos + 1);
```
fhueske: I think it would be better to extract the schema / return types only once and not for every tuple.

stratosphere-addons/jdbc/src/main/java/eu/stratosphere/api/java/io/jdbc/JDBCOutputFormat.java
```diff
@@ -0,0 +1,258 @@
...
+        } catch (IllegalArgumentException iae) {
+            close();
+            throw new IllegalArgumentException("writeRecord() failed", iae);
+        }
+    }
+
+    private void addValue(int index, OUT tuple) throws SQLException {
+        supportedTypes type;
+        try {
+            type = supportedTypes.valueOf(tuple.getField(index).getClass().getSimpleName());
```
fhueske: Can we do obtain the types only for the first tuple, store them, and reuse them for the remaining tuples?

stratosphere-addons/jdbc/src/test/java/eu/stratosphere/api/java/io/jdbc/JDBCInputFormatTest.java
```diff
@@ -0,0 +1,142 @@
...
+            Assert.assertEquals("Field 4 should be int", Integer.class, tuple.getField(4).getClass());
+
+            for (int x = 0; x < 5; x++) {
+                Assert.assertEquals(dbData[recordCount][x], tuple.getField(x));
+            }
+            recordCount++;
+        }
+        Assert.assertEquals(5, recordCount);
+    }
+
```
fhueske, line 188: Can you add a few more tests to check for proper exceptions in case of incorrect configurations? For example:
- Provided tuple is not wide enough (not enough fields) to hold query result tuples.
- JDBC class does not exist.
- Query is incorrect.
- ...

stratosphere-addons/jdbc/src/test/java/eu/stratosphere/api/java/io/jdbc/JDBCOutputFormatTest.java
```diff
@@ -0,0 +1,179 @@
...
+                Assert.assertEquals(dbData[recordCount][x], tuple.getField(x));
+            }
+
+            recordCount++;
+        }
+        Assert.assertEquals(5, recordCount);
+
+        jdbcInputFormat.close();
+    }
+}
```
fhueske, line 252: Similar to the InputFormatTest, some tests for invalid configurations would be good:
- Invalid query.
- Invalid connection credentials / driver class name.
- Tuple fields do not match parameters.
- Tuple field types do not match table column types.
- ...

fhueske:
Looks pretty good to me. 
I would extend the tests to check also for failure handling and not only for correct working. Also, type handling during read and write could be improved a bit.

Otherwise, very solid work, thanks!

stratosphere-addons/jdbc/src/main/java/eu/stratosphere/api/java/io/jdbc/JDBCOutputFormat.java
```diff
@@ -0,0 +1,269 @@
...
+    private void establishConnection() throws SQLException, ClassNotFoundException {
+        Class.forName(drivername);
+        if (username == null) {
+            dbConn = DriverManager.getConnection(dbURL);
+        } else {
+            dbConn = DriverManager.getConnection(dbURL, username, password);
+        }
+    }
+
+    private enum supportedTypes {
```
fhueske: Enums are classes and should follow their naming conventions, i.e., CamelCase: -> SupportedTypes
Enum instances are constants and should follow their conventions: ALLCAPITAL: -> BOOLEAN

fhueske:
Looks very good! Thanks for adding the tests.
I have one more comment in-line (Enum naming conventions) and a general comment.
With PR #636 we will have style checks in place. The project uses tabs for indention and not spaces. So it would be good, if you would convert your indention spaces to tabs.

After that, I think this PR is good to merge. Thanks!

fhueske:
Good to merge :+1:

fhueske:
Will merge this now.

Issue #615, opened 2014-03-20
Provide a python map / python reducer in java api.


rmetzger:
Current language binding code https://github.com/stratosphere/stratosphere/issues/528

StephanEwen:
At some point we have to be careful not to offer too many methods on
DataSet. Otherwise, people get lost.

How about we offer a specific java mapper (as part of the python project)
that implements the bridge between java and python. We could then use it
the following way:

```
DataSet<String> set = ...;

set.map(new PythonMapper("./path/to/the/python/script/")).writeTo(...);
```

zentol:
I have started working on this.

zentol:
..and ran into an issue.

This is the current state:
there is one generic PythonMapper class

```
public class PythonMapper<IN extends Tuple,OUT extends Tuple> extends MapFunction<IN, OUT> {
...
}
```

if you write a plan, you have to create a new class: (which really contains nothing else)

```
public static class PythonMapperCustom extends PythonMapper<Tuple1<String>,Tuple1<String>> {
    public PythonMapperCustom(String scriptPath) {
        super(scriptPath);
    }
}
```

(using PythonMapper directly results in type errors, this is what i tried first)

then you can (theoretically) do this:

```
data
   .map(new PythonMapperCustom(path))
   .print();
```

but now i get this error:

```
Exception in thread "main" java.lang.IllegalArgumentException: Generic function base class must be immediate super class.
    at eu.stratosphere.api.java.typeutils.TypeExtractor.getSuperParameterizedType(TypeExtractor.java:165)
    at eu.stratosphere.api.java.typeutils.TypeExtractor.getTemplateTypes(TypeExtractor.java:135)
    at eu.stratosphere.api.java.typeutils.TypeExtractor.getMapReturnTypes(TypeExtractor.java:31)
    at eu.stratosphere.api.java.operators.MapOperator.<init>(MapOperator.java:34)
    at eu.stratosphere.api.java.DataSet.map(DataSet.java:75)
    at eu.stratosphere.languagebinding.example.NewClass.main(NewClass.java:44)
```

so, i can't use the generic PythonMapper directly due to type issues, nor can i create a less generic subclass from it.

any help would be appreciated.

twalthr:
The current TypeExtractor implementation does not support this complex generic stuctures. I'm currently working on this.

Try my current working branch https://github.com/twalthr/stratosphere/tree/TypeExtractorTests it should fix this issue.

zentol:
[...]

PR #564, opened 2014-03-12
Add distributed cache to stratosphere

qmlmoon:
implement issue #506

PR #671, opened 2014-04-09
Python interface for new API (Map/Reduce)

zentol:
#615


zentol:
This PR now includes the flatMap-Operator aswell.

zentol:
now includes CoGroup and Join.

zentol:
now includes Filter/GroupReduce/Cross

rmetzger:
Can you add an example that includes a complex machine learning algorithm written in python (but in a library, such as http://scikit-learn.org/stable/) ?
I'm asking for this since we want to allow our users to use the excellent algorithms available as python libraries.

zentol:
I'll look into it.

rmetzger:
I've merged the DistributedCache you are depending on (https://github.com/stratosphere/stratosphere/pull/564).

zentol:
@rmetzger Can we add a more intuitive way to register these files? currently i have to write this:

```
l.executePlan(p);
```

whereas

```
env.registerCachedFile(path, "name");
env.execute();
```

would feel more in line with the new API.

rmetzger:
Yeah, I think the DC is not really integrated with the new Java API. 
Can you do this integration and open a pull request?

zentol:
sure.

zentol:
updated the examples for the changes to the distributed cache.
the usability issue of having to create custom functions was solved with @twalthr help. turns out that curly braces are magical things.

instead of

```
public static class CustomPythonMap<Integer, String> extends PythonMap<Integer, String>{
...constructor stuff...
}

data.map(new CustomPythonMap(..))
```

you can just write

```
data.map(new PythonMap<Integer, String(..){})
```

zentol:
I have uploaded the new version of the python interface.

Big changes:
- plans can be written completely in python (not every operation for environments/datasets are included yet though)
- the stratosphere package (including the protobuf package, which as a sideeffect blows up the additions/file count) are distributed automatically among the cluster (this is currently mandatory)

Small changes
- java side streamlined and uses threads
- functions implement the ResultTypeQueryable interface (no more curly braces!)

Questionable things (aka stuff that i need a confirmation for):
- The stratosphere python package will show up in the resource folder of the distribution. i needed a static place in order to run the python plan. I wonder if there is a more appropriate place.
- In order to get the path to the stratosphere package, i needed the location of the stratosphere distribution. for this i exported the STRATOSPHERE_ROOT_DIR variable in the scripts. (i guess i could use the already exported conf_dir variable and go from there, but that seems like jumping through hoops.)
- the protobuf package is now included in the distribution. it is licensed under http://opensource.org/licenses/BSD-3-Clause . Will this work? (and if it does, will i have to modify the licenses in the protobuf files in any way?)

This PR depends on:
#882

uce:
The PR branch is on halt. @zentol will re-open in the new repo with the current state.

PR #677, opened 2014-04-10
New Annotations for JAPI.

skunert:
Annotations for the new API were implemented. There is now a sepparate FunctionAnnotation class for the new japi because the code is much more different than the code for the record API. The annotations also have new features. It is now possible to specify which field of the input is copied where.

**_Example:**_

```
@ConstantFields(inTuplePos={0,1,2}, outTuplePos={2,0,1})
```

The example specifies that three fields of the input are copied but are on a different position in the output.  At the moment, this only works for tuples. If only the input positions are specified, it is assumed that the position of the specified fields remains the same. If a udf has other input and output, the annotations are ignored. Nevertheless, there is already some infrastructure for future work. You can annotate like this:

```
@ConstantFields(inCustomPos={"Field1", "Field2"}, outCustomPos={"Field2", "Field1"})
```

It is also now possible to hand semanticproperties to the operators themselves via the `withSemanticProperties()` method.

### DIFF
--- a/stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/wordcount/WordCountCollection.java
+++ b/stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/wordcount/WordCountCollection.java
@@ -17,14 +17,15 @@ package eu.stratosphere.example.java.wordcount;
 import eu.stratosphere.api.java.DataSet;
 import eu.stratosphere.api.java.ExecutionEnvironment;
 import eu.stratosphere.api.java.functions.FlatMapFunction;
+import eu.stratosphere.api.java.functions.FunctionAnnotationJapi.ConstantFields;
 import eu.stratosphere.api.java.tuple.*;
 import eu.stratosphere.util.Collector;
-
 import static eu.stratosphere.api.java.aggregation.Aggregations.*;
 
 @SuppressWarnings("serial")
+
 public class WordCountCollection {
-	
+	@ConstantFields(from={1,2,3}, to={0,1})
 	public static final class Tokenizer extends FlatMapFunction<String, Tuple2<String, Integer>> {
 
 		@Override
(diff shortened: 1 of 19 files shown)


StephanEwen:
Looks good, I really like where this is going!

Can you add the a `value` field to the annotation, to allow constructs like `@ConstantFields({1, 2})`, which would interpret them as staying constant as one and two?

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/wordcount/WordCountCollection.java
```diff
@@ -19,12 +19,11 @@
 import eu.stratosphere.api.java.functions.FlatMapFunction;
 import eu.stratosphere.api.java.tuple.*;
 import eu.stratosphere.util.Collector;
-
```
StephanEwen, line 22: Can you remove the changes to this file from the pull request? I know it may seem picky, but it should really be a general rule to not just commit any reformatting that is totally unrelated to the changes.

stratosphere-java/src/main/java/eu/stratosphere/api/java/functions/FunctionAnnotation.java
```diff
@@ -0,0 +1,523 @@
...
+
+/**
+ * This class defines the semantic assertions that can be added to functions.
+ * The assertions are realized as java annotations, to be added to the class declaration of
+ * the class that realized the user function. For example, to declare the <i>ConstantFieldsExcept</i> 
+ * annotation for a map-type function that realizes a simple absolute function,
+ * use it the following way:
+ * 
+ * <pre><blockquote>
+ * \@ConstantFieldsExcept(inTuple={1,2}, outTuple={2,1})
```
StephanEwen: The names here do not match the field names in the annotation.

StephanEwen:
It would be good to have something like `@AllConstant` or `@Unmodified` if a function forwards all fields

StephanEwen:
While these annotations fill the copy sets (or forward sets) in the semantic properties, we can also think about annotations that define the read set (which fields are actually accessed by the function).

skunert:
The annotation `@AllFieldsConstants` is still present.

StephanEwen:
I will merge this one now. It looks good.

For the future, can you review your pull requests with respect to formatting and unnecessary changes? The changes to the file https://github.com/stratosphere/stratosphere/pull/677/files#diff-17 for example look like artifacts that should be excluded.

StephanEwen:
The first part is merged here: a13a44d17d806378fee4b36b1e84cf98cdbe0079

For any subsequent changes, please make a new pull request.

Thanks for the patch!

PR #678, opened 2014-04-10
Add PojoTypeInformation, PojoComparator and Expressions for Keys

aljoscha:


StephanEwen:
Should we merge this now? I think the general tendency is to include this in 0.6, and not in 0.5  What do you think?

fhueske:
I would not add it to the 0.5 feature list.

StephanEwen:
@aljoscha Is this pull request still valid, or is it out of sync with the feature dev already?

aljoscha:
I'm not sure. It works, but is limited and slow because the comparators use introspection to get at the data.

StephanEwen:
We move the feature to 0.6 anyways, to that is fin. We will not merge it as of now. 0.5 is feature frozen ;-)

aljoscha:
Superseded by #908

PR #674, opened 2014-04-10
New java Api Union

filiphaase:
Corr. Issue: https://github.com/stratosphere/stratosphere/issues/621

A union operator for the new Java-API callable: dataSet1.union(dataSet2)
There exists a binary union operator from the java-api down to the optimizer level. 
This means that unions do not need to be handled over lists of operators by the commons-api anymore, but a cascade of union-operators can be used for this case. Therefore my next step is to clean the code from the operator lists to make the code "slimmer".

Issue #621, opened 2014-03-21
Add Union operator to new Java API


fhueske:
Add support to union two (or more) identically typed data sets.
This is possible in the former API, so we need to support it here as well.

rmetzger:
As far as I am aware, @filiphaase is working on this.

rmetzger:
Please always write in the issue or open one if you start working on something. We had at least two cases in the past where people were working on the same topic without knowing.

filiphaase:
Okay, thought assignment is enough ;)

fhueske:
I think either is fine.

rmetzger:
The problem with assignment is that there will be no email notification.
(And we do not really use the assignment feature).

fhueske:
@filiphaase please add tests for Union analogous to `stratosphere-tests/src/test/java/eu/stratosphere/test/javaApiOperators/FilterITCase.java` and `stratosphere-tests/src/test/java/eu/stratosphere/test/javaApiOperators/ReduceITCase.java`.

fhueske:
well, I do. ;-)
And I guess it is exactly for this purpose.

StephanEwen:
Fixed in e29cba57f0a03776287fea41046c21e6951b7fe0

PR #680, opened 2014-04-13
New-Java-API Union

filiphaase:
First PR: https://github.com/stratosphere/stratosphere/pull/674
Corr. Issue: https://github.com/stratosphere/stratosphere/issues/621

Now the old-java-api is also using the binary-union-operator by automatically creating a cascade of unions if needed. Therefore every operator in the common-package saves a single operator instead of a list of operators.

But the scala-api has some parts where it expects list of operators for getInputs() calls. Therefore there are getInputs() methods which are wrapping lists around the single operator and return the list. This is a little be ugly, but I assume we switch the scala-api to use the new java-api in the future, so we can remove them then.

### DIFF
--- a/stratosphere-compiler/src/main/java/eu/stratosphere/compiler/PactCompiler.java
+++ b/stratosphere-compiler/src/main/java/eu/stratosphere/compiler/PactCompiler.java
@@ -37,6 +37,7 @@ import eu.stratosphere.api.common.operators.DeltaIteration.WorksetPlaceHolder;
 import eu.stratosphere.api.common.operators.GenericDataSink;
 import eu.stratosphere.api.common.operators.GenericDataSource;
 import eu.stratosphere.api.common.operators.Operator;
+import eu.stratosphere.api.common.operators.Union;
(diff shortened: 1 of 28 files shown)


StephanEwen:
The first commit is merged in 4bcbd133ac43289c268ba776efdf9b7de88d1dca

filiphaase:
Fixed bug @StephanEwen commented on earlier.

fhueske:
@filiphaase Can you please also update the documentation on the [website](http://stratosphere.eu/docs/0.5/programming_guides/java.html#transformations). Union (at the end of the Data Transformation section) is marked as unsupported and not documented yet. Thanks!

filiphaase:
@fhueske Done in PR https://github.com/stratosphere/stratosphere.github.io/pull/31.

fhueske:
@filiphaase Thanks! Merged it.

StephanEwen:
Merged in e29cba57f0a03776287fea41046c21e6951b7fe0

PR #686, opened 2014-04-14
Added support for local distributed execution environments.

tillrohrmann:
I added support for local distributed execution environments similar to the LocalDistributedExecutor. For this purpose I generalized the LocalInstanceManager to be able to host multiple TaskManager threads. Local distributed execution should be useful for easy local testing of scheduler strategies. Moreover, this feature might be interesting for local execution on a multi-core system.


tillrohrmann:
I made the hardware description of the task manager a static variable to fix a bug in the memory calculation if multiple task manager run on the same jvm. Before the hardware description was calculated for each task manager individually. If multiple task manager run on the same jvm then the free memory is simply divided by their number. However, once created the memory manager, the free memory was directly reduced so that a subsequently instantiated task manager saw only the reduced free memory. Dividing this amount by the number of task managers running on the jvm leads now to a wrong result.

stratosphere-runtime/src/main/java/eu/stratosphere/nephele/instance/local/LocalInstanceManager.java
```diff
@@ -14,10 +14,7 @@
 package eu.stratosphere.nephele.instance.local;
 
 import java.io.File;
-import java.util.ArrayList;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
uce: Can you replace this with single imports? IntelliJ's default is to use \* imports if you include more than a configured number of imports from the same package... http://stackoverflow.com/questions/3587071/disable-intellij-starred-package-imports

With the new Maven check-style pluging this will not pass Travis.

stratosphere-tests/src/test/java/eu/stratosphere/test/exampleJavaPrograms/WordCountITCase.java
```diff
@@ -24,6 +25,10 @@
 	protected String textPath;
 	protected String resultPath;
 
+	public WordCountITCase(){
+		this.config.setInteger(ConfigConstants.LOCAL_INSTANCE_MANAGER_NUMBER_TASK_MANAGER, 2);
```
uce: Nice. Maybe we should think about setting the default value to something > 1 in order to always test the network stack with the IT cases. But this might result in a a problem with the time we have on Travis.

stratosphere-java/src/main/java/eu/stratosphere/api/java/LocalEnvironment.java
```diff
@@ -61,7 +67,7 @@ public boolean isLoggingEnabled() {
 	private void initLogging() {
 		LogUtils.initializeDefaultConsoleLogger(logging ? Level.INFO : Level.OFF);
 	}
-	
+
 	@Override
 	public String toString() {
 		return "Local Environment (DOP = " + (getDegreeOfParallelism() == -1 ? "default" : getDegreeOfParallelism()) + ") : " + getIdString();
```
uce: Let's add the number of TMs here?

uce:
I noticed some minor stuff (see comments). Very nice to have this. :-)

stratosphere-java/src/main/java/eu/stratosphere/api/java/ExecutionEnvironment.java
```diff
@@ -299,6 +300,18 @@ public static ExecutionEnvironment createRemoteEnvironment(String host, int port
 		rec.setDegreeOfParallelism(degreeOfParallelism);
 		return rec;
 	}
+
+	public static ExecutionEnvironment createLocalDistributedEnvironment(int numTaskManager){
```
uce: Should we just merge this with `createLocalEnvironment`, i.e. just overload the method with the extra parameter? I personally never liked the name local distributed executor, but maybe that's just me. ;-)

stratosphere-runtime/src/main/java/eu/stratosphere/nephele/instance/local/LocalInstanceManager.java
tillrohrmann: Never trust IntelliJ's auto import... it's too smart.

uce:
I just noticed a wrong error message (while doing s.th. else) in `createRemoteExecutor` of `PlanExecutor`:

``` java
catch (Throwable t) {
            throw new RuntimeException("An error occurred while loading the local executor (" + LOCAL_EXECUTOR_CLASS + ").", t);
}
```

Can you also fix this before we merge this?

stratosphere-java/src/main/java/eu/stratosphere/api/java/ExecutionEnvironment.java
tillrohrmann: Actually I did it at first like that, but then you have the problem that there is already another method with an int parameter, namely with degreeOfParallelism. This implies that one cannot independently specify degreeOfParallelism or numTaskManager.

uce:
Oh wait, I think it's not just the error message... it's a real problem. I'm calling

``` java
final ExecutionEnvironment env = ExecutionEnvironment.createRemoteEnvironment("localhost", 6123);
```

and get the following message:

```
Exception in thread "main" java.lang.RuntimeException: An error occurred while loading the local executor (eu.stratosphere.client.LocalExecutor).
    at eu.stratosphere.api.common.PlanExecutor.createRemoteExecutor(PlanExecutor.java:75)
    at eu.stratosphere.api.java.RemoteEnvironment.execute(RemoteEnvironment.java:51)
    at eu.stratosphere.example.java.wordcount.WordCount.main(WordCount.java:61)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:606)
    at com.intellij.rt.execution.application.AppMain.main(AppMain.java:120)
Caused by: java.lang.NoSuchMethodException: eu.stratosphere.client.LocalExecutor.<init>(java.lang.String, int, java.util.List)
    at java.lang.Class.getConstructor0(Class.java:2810)
    at java.lang.Class.getConstructor(Class.java:1718)
    at eu.stratosphere.api.common.PlanExecutor.createRemoteExecutor(PlanExecutor.java:72)
    ... 7 more
```

tillrohrmann:
I'll look into it.

tillrohrmann:
The error didn't occur on my machine. I could instantiate a remote environment.

I removed the local distributed executor environment from the user api. For testing purposes one can still use. See the WordCountITCase.

rmetzger:
I guess @uce's error is caused by some outdated maven dependencies? I think a `mvn clean install` will resolve the issue.

@tillrohrmann: Why can't we just remove the LocalDistributedExecutor class? I guess your implementation is better integrated into the system than the LDE impl.

tillrohrmann:
If I'm not mistaken, then the LocalDistributedExecutor is only used in the `PackagedProgramEndToEndITCase`. Thus we only need to replace there the LDE with the NepheleMiniCluster and it should be fine. Apart from that, everything should be working. I'll do the reworking.

tillrohrmann:
I removed the LDE.

StephanEwen:
I would like to merge this.

One question, though: Why configure the number of task managers through the config and not simply through a setter, like the other parameters (memory, ports, overwrite)?

tillrohrmann:
Initially, I wanted to integrate this feature as transparent as possible for the different components, because it should only be used for testing purposes. But you are right that it wouldn't actually be exposed to the user API if the NepheleMiniCluster gets a setter for the number of task managers. And additionally it is more elegant.

tillrohrmann:
Ok, I refactored it.

StephanEwen:
Sorry, to ping you again. Can you rebase this to the master? There have been a few changes to the Task Manager in the last week.

One thing that has an implication on this code is that the task managers do not get the "numTaskmanagers" parameter any more. The LocalDistributedExecuter in the curret HEAD configures them with memory that is scaled down by the number of TaskManagers to start.

I think you need something similar here...

tillrohrmann:
I rebased on the current master. Now, the memory size which one can define for the NepheleMiniCluster is considered to be the total available memory and thus is split among the TaskManagers. If the memory size is not set, then he takes the available free memory minus the memory space for the buffers.

StephanEwen:
Wow, great! That should work! Sorry for making you go though all the
changes...

StephanEwen:
Merged in 93c4b9ae8d633efd49d02dc1702185f670f9c6e7

Issue #482, opened 2014-02-17
Exceptions on TaskManagers (usercode, input formats etc) are not displayed in the JobManager log


rmetzger:
The JobManager log file should contain the exceptions of a failing task. 

This is currently only possible by starting a job using the `-w` argument of the stratosphere client.
The exception should always be in the JobManager log file!

ghost:
hello, 

I would like to work on this and provide a patch.

rmetzger:
Cool.
I would suggest to take the wordcount example and throw a Runtime Exception in the mapper. Then, run the Job on a cluster with the `-w` option. You'll see your exception.
As a next step, start a _distributed_ Stratosphere cluster (e.g. not  `start-local.sh`). Running the job again (without `-w`) will also fail the job, but this time, you won't see your Exception in the jobmanager log file, only in one of the taskmanager's log files (or in all TMs).

It would be great if the exception is also shown in the JobManager log file. As a awesome bonus, we would also like to show the Exception in the JobManager's webinterface (on port 8081). Once you've managed to fix the first issue, the webinterface thing will be easy.

StephanEwen:
The exceptions are already reported from task manager to the job Manager,
which forwards them to the client (and in -w Mode, the client prints the
exception).

If you can find the point where the job manager stores the exception for
the jobclient to pull the status (and the exception), you probably have a
good point to start. Logging it is most likely a quick fix then, forwarding
the exception to the web Frontend would be a very nice addition!

StephanEwen:
This was fixed as part of #688 (https://github.com/stratosphere/stratosphere/pull/688/files#diff-0808a97b4666fab09f9746c60e4a9b36R669)

Issue #483, opened 2014-02-17
Display Exceptions in the JobManager webinterface


rmetzger:
Currently, users have to scan through the log files to understand why a job has failed.

The web interface should show the exception of a failed job to ease debugging (Because these exceptions are usually quite helpful).

This enhancement depends on a fix for https://github.com/stratosphere/stratosphere/issues/482.

qmlmoon:
Is #482 fixed?

StephanEwen:
Fixed by #688 in a2650c5749d1c3eba40364123d397fdafbcf30e4

PR #688, opened 2014-04-14
Display Exceptions in the JobManager webinterface

qmlmoon:
fix issue #482 and #483

### DIFF
--- a/stratosphere-runtime/src/main/java/eu/stratosphere/nephele/jobmanager/web/JobmanagerInfoServlet.java
+++ b/stratosphere-runtime/src/main/java/eu/stratosphere/nephele/jobmanager/web/JobmanagerInfoServlet.java
@@ -249,7 +249,7 @@ public class JobmanagerInfoServlet extends HttpServlet {
 					} else {
 						wrt.write(",");
 					}
-					wrt.write("{\"node\": \"" + instanceName + "\"}");
+					wrt.write("{\"node\": \"" + instanceName + " - " + managementVertex.getOptMessage().replaceAll("\n", "<br\\>").replaceAll("\t", " ")+ "\"}");
 					map.add(instanceName);
 				}
 			}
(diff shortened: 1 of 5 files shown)


stratosphere-runtime/src/main/java/eu/stratosphere/nephele/jobmanager/web/JobmanagerInfoServlet.java
```diff
@@ -249,7 +249,7 @@ private void writeJsonForArchivedJob(PrintWriter wrt, RecentJobEvent jobEvent) {
 					} else {
 						wrt.write(",");
 					}
-					wrt.write("{\"node\": \"" + instanceName + "\"}");
+					wrt.write("{\"node\": \"" + instanceName + " - " + managementVertex.getOptMessage().replaceAll("\n", "<br\\>").replaceAll("\t", " ")+ "\"}");
```
markus-h: I think it would be nicer to not add the optional message to the instance name string, but to introduce an additional JSON attribute for this message.
The html part of the webinterface then needs to be adjusted to display this new attribute in a nice way.

StephanEwen: Please also use the StringUtils method to escape a String for HTML

uce:
This is what it looks like right now:

screen shot 2014-04-14 at 15 15 49

stratosphere-runtime/src/main/java/eu/stratosphere/nephele/jobmanager/web/JobmanagerInfoServlet.java
uce: Agree with @markus-h. Let's have an extra attribute and display it after a line break or something in the web interface.

StephanEwen:
@uce Did you already test it?
@markus-h Could  you review the final version of this pull request?

markus-h:
The code looks good for me.

uce:
Yes, I've tested it (locally).

StephanEwen:
I have started testing code locally with `startCluster.sh` instead of `startLocal.sh`. That starts a cluster of one machine, where JobManager and TaskManager are co-located, but in two different JVMs. This should cover already most of the cases that we encounter in clusters. An exception are HDFS and issues arising from the fact that in a distributes setup, the directory structure may be different between master and worker.

StephanEwen:
Merged in a2650c5749d1c3eba40364123d397fdafbcf30e4

Issue #588, opened 2014-03-16
Descending group sorting does not work.


StephanEwen:
The flag that describes whether comparisons should result in an ascending or descending order is not used by any of the basic type comparators.

twalthr:
You mean something like that in every Comparator, right?

<code>
@Override
    public int compare(DataInputView firstSource, DataInputView secondSource) throws IOException {
        double l1 = firstSource.readDouble();
        double l2 = secondSource.readDouble();
        int comp = (l1 < l2 ? -1 : (l1 > l2 ? 1 : 0));
        return ascendingComparison ? comp : -comp;
    }
</code>

StephanEwen:
Exactly

markus-h:
Is this fixed? I am working on testcases for groupReduce and the descending order seems to be not working. I think this is connected to this issue.

fhueske:
No, I don't think this is fixed. I have the same issue with a deactivated test in `ReduceITCase`.

twalthr:
I will do it, tomorrow. ok?

markus-h:
:+1:

fhueske:
I had to remove the GroupSorting test in the `ReduceITCase` because we do no longer support group sorting for the pair-wise `ReduceFunction`.
However, we should check if this solves the problem with GroupSorting (Ascending and Descending) in the `GroupReduceITCase`.

markus-h:
@twalthr and me are working on it. The simple change mentioned above seems not to work.

fhueske:
Great! Thanks! :smile:

markus-h:
Ok, here are my investigations:
What I do is a `groupBy(1).sortGroup(0,Order.DESCENDING)`. In the compiler post pass the localStrategyComparator of the channel to the reduce is correctly set to [0] LongComparator(asc) and [1] IntComparator(desc). Both comparators are instantiated, but the second one is never used. I set a breakpoint inside the compare() method and it is never called.

So I had a look at the ReduceDriver, more specifically to the KeyGroupedIterator and its ValueIterator and I don't see how there should be support for sorted values inside of a group.

Is there runtime support for sorted groups in the reduce?

fhueske:
[...]

Issue #548, opened 2014-03-08
Tests for the operators of the new Java API


fhueske:
The operators of the new Java API need to be tested for different data types (Tuples, Native Java Objects (Generic Avro Serialization), ...) and different key definitions (record fields, key extractors) in different combinations.

fhueske:
I thought about the operator tests and made a [wiki page of tests that we should run](https://github.com/stratosphere/stratosphere/wiki/Operator-Tests-New-Java-API).
Please review, comment, extend.

Since this is a larger effort, it would be cool, if some of you guys would pick one or two operators and write tests for them. 
I started with the tests for [**Filter**](https://github.com/stratosphere/stratosphere/blob/a7d10395058f21f1c66375857c92544ef2a78ce9/stratosphere-tests/src/test/java/eu/stratosphere/test/javaApiOperators/FilterITCase.java) and will continue with tests for **Reduce**.

If you decide to test an operator, have a look at the [filter tests](https://github.com/stratosphere/stratosphere/blob/a7d10395058f21f1c66375857c92544ef2a78ce9/stratosphere-tests/src/test/java/eu/stratosphere/test/javaApiOperators/FilterITCase.java) and drop a line, so that it is clear what is currently being worked on.

vasia:
Hey! I can work on this, e.g. for Map and FlatMap to start with. Is that OK?

fhueske:
Yes, definitely! That would be great! More tests a always very welcome! :smile:

vasia:
Hey @fhueske, I have pushed my Map operator test [here](https://github.com/vasia/ozone/blob/japi_operator_tests/stratosphere-tests/src/test/java/eu/stratosphere/test/javaApiOperators/MapITCase.java). 
Could you take a look and let me know if that's what you had in mind? If yes, I'll continue with the FlatMap operator :) Thanks!

fhueske:
Hi @vasia,
looks very good to me! Thank you!

I'll be working on tests for Join.

markus-h:
I started working on cogroup

markus-h:
@fhueske In your wiki page you wrote "check correctness of cogroup a tuple input with key field selector and a custom type input with key extractor"
Currently there is no support for mixing key extractor and field position keys. Should this be implemented?

fhueske:
Yes, I think we should support that.

markus-h:
I opened an issue for that: #675

vasia:
[...]

PR #695, opened 2014-04-15
GroupReduce testcases for the new java api

markus-h:
See #548.
I had to do a small fix since for some reason the TupleUnwrappingIterator was not instantiated.

The descending sorting of tuples inside of the groups is not working yet. I think this is connected to #588.

### DIFF
--- a/stratosphere-tests/src/test/java/eu/stratosphere/test/javaApiOperators/GroupReduceITCase.java
+++ b/stratosphere-tests/src/test/java/eu/stratosphere/test/javaApiOperators/GroupReduceITCase.java
@@ -1,0 +1,523 @@
+/***********************************************************************************************************************
+ *
+ * Copyright (C) 2010-2013 by the Stratosphere project (http://stratosphere.eu)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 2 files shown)


stratosphere-tests/src/test/java/eu/stratosphere/test/javaApiOperators/GroupReduceITCase.java
```diff
@@ -0,0 +1,523 @@
...
+	}
+
+	@Override
+	protected void testProgram() throws Exception {
+		expectedResult = GroupReduceProgs.runProgram(curProgId, resultPath);
+	}
+	
+	@Override
+	protected void postSubmit() throws Exception {
+		if(expectedResult.substring(0, 6).equals("SORTED"))
```
fhueske, line 66: Group sorting does not mean, that the groups are processed in the specified order.
Instead the iterator returns the elements **within** a group in a specified order. So the result tuples might not be ordered in key order.

fhueske, line 66: I fixed this with another commit and will merge the PR.
So no need to update it.

Thanks!

fhueske:
Merged in 9e5e73a0ae26f0471ede62025f2751fdb3685dea

PR #696, opened 2014-04-15
Csv reader tuple subclasses

skunert:
It is now possible to create DataSets of subclasses of Tuples via the CsvReader.

**_Example**_

```
DataSet<Lineitem> lineitems = env
    .readCsvFile(lineitemPath).fieldDelimiter('|')
    .includeFields("1000011000100000")
    .tupleType(Lineitem.class);
```


StephanEwen:
Looks good, but it still misses a test ;-)
Can you add a test case for this?

skunert:
There is now a simple test for the reader. Is this what you had in mind?

StephanEwen:
I think it would be good to improve that a bit
- It is not a unit test, for one, and would not be executed by the test cycle.
- It would be nice if it would assert on the parameterization of the csv format. A test that does not write a temp file and read it, but create a csv reader with a derived tuple and then check via the `getGenericFieldTypes()` method whether the correct types were extracted. That is more in the spirit of a unit test.
- A second test along the same line with a subclass of a subclass would be good as well. Best even, if the two subclasses parameterize each some of the generic parameters. That would check whether the right method of the type extractor is called. Have a look at the type extractor tests, they do that.
- A final test that checks whether the return type is actually the derived class, and not the tuple class.

skunert:
It is now a JUnit test.

StephanEwen:
Okay, I will merge this one.

StephanEwen:
Merged in 45bed7af76a9adfa27624979ad498ed8dfc41743 and 44ada15462f2bf22dcc4efcb62335116feeafcd2

PR #697, opened 2014-04-16
Moved SerializerTestBase to stratosphere-core and resolved Maven dependencies

fhueske:
Moved the SerializerTestBase from stratosphere-java to stratosphere-core where the abstract TypeSerializer class and all base type serializers are defined.
Had add some Maven stuff to get the dependency on test code working.


fhueske:
Was merged in 4fe742635f3a6dced7b9bf66f9c34ef7c1e053dc

Issue #694, opened 2014-04-15
Defining TypeInformation with a String


twalthr:
While implementing test cases for the TypeExtractor, I found it very inconvenient to define TypeInformation manually.

What do you think about a method like:

``` java
TypeInformation.parse("Tuple2<String,Tuple3<Boolean[],Integer,org.my.CustomObject>>");
```

I would like to implement it as it would help me at my current task (#683).

StephanEwen:
+1

fhueske:
Solved in #698

Issue #683, opened 2014-04-13
Type extractor should detect type mismatches


StephanEwen:
Currently, the type extractor does not check whether the input type(s) of functions and the TypeInformation of the input data sets match. It would be good to check that, in order to catch cases where programmers omit the generics on the data set and can compile programs where the input type does not fit.

The code below should not result in a successful test.

```
public void testFunctionDependingOnInputWithTupleInputWithTypeMismatch() {
    IdentityMapper2<Boolean> function = new IdentityMapper2<Boolean>();

    TypeInformation<Tuple2<Boolean, String>> inputType = new TupleTypeInfo<Tuple2<Boolean, String>>(BasicTypeInfo.BOOLEAN_TYPE_INFO,
            BasicTypeInfo.INT_TYPE_INFO);

    TypeInformation<?> ti = TypeExtractor.getMapReturnTypes(function, inputType);

    Assert.assertTrue(ti.isBasicType());
    Assert.assertEquals(BasicTypeInfo.BOOLEAN_TYPE_INFO, ti);
}

public class IdentityMapper2<T> extends MapFunction<Tuple2<T, String>, T> {
    private static final long serialVersionUID = 1L;

    @Override
    public T map(Tuple2<T, String> value) throws Exception {
        return null;
    }
}
```

twalthr:
I'm working on this.

twalthr:
This issue can be closed.

PR #698, opened 2014-04-16
Implements a parse() method to parse TypeInformation in string representation.

twalthr:
Explained in #694.


fhueske:
This is a cool feature!
Can you also add a test case?

twalthr:
Would be great if you could merge this PR soon, because then I can create a PR for #683.

fhueske:
I'll have a look at it.

fhueske:
Merged.

Issue #589, opened 2014-03-16
Add Tests for the basic type utils


StephanEwen:
We need thorough test for the basic type utils, as they are a core part of the API type system.

fhueske:
This issue refers to the untested serializers and comparators in the `stratosphere-core` module in package `eu.stratosphere.api.common.typeutils.base` for the basic data types:
- Boolean
- Byte
- Char
- Double
- Float
- Int
- Long
- Short
- String

The *ValueSerializer classes are for the data types of our old Java API which will be depreciated in the future. 

We need thorough unit-tests for all of these classes because they will be heavily used. The tests should be put in the test directory of the `stratosphere-core` module in the same package as the tested classes. 

I would start with tests for the *Serializer classes. and continue with the *Comparator classes.

StephanEwen:
I think we should retain the value types of mutable versions of javas
primitives.

StephanEwen:
Have a look at the string array serializer tests. It uses a serializer test
util that makes it easy to add serializer tests.

StephanEwen:
stratosphere-java/src/main/java/eu/stratosphere/api/java/typeutils/runtime/StringArraySerializer.java
and the associated test.

aaronchlam:
I will be working on *Serializer tests in this issue and then move on to do the *Comparator if someone else does not get to it first.

fhueske:
With  #745, the serializers for all nine primitive types are covered with unit tests.
Thanks @aaronchlam !

Tests for comparators are addressed by issue #744

PR #702, opened 2014-04-17
Add test for String Serializer for #589

aaronchlam:
#589

First test is the String Serializer. One thing that I'm wondering about the StringSerializer is whether it should be able to handle `null` values. I took the test strings from `stratosphere-core.src.test.java.eu.stratosphere.types.StringSerializationTest.java` as the test strings for the StringSerializer.

It was able to handle `new String[] {"a", "", "bcd", "jbmbmner8 jhk hj \n \t üäßß@µ", "", "non-empty"}`
 but not `new String[] {"a", null, "", null, "bcd", null, "jbmbmner8 jhk hj \n \t üäßß@µ", null, "", null, "non-empty"}`.

I am not sure if this is intended, so I have kept the array that does not include null for now.

### DIFF
--- a/stratosphere-core/src/test/java/eu/stratosphere/api/common/typeutils/base/StringSerializerTest.java
+++ b/stratosphere-core/src/test/java/eu/stratosphere/api/common/typeutils/base/StringSerializerTest.java
@@ -1,0 +1,47 @@
+/***********************************************************************************************************************
+ *
+ * Copyright (C) 2010-2013 by the Stratosphere project (http://stratosphere.eu)
+ *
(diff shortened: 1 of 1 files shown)


fhueske:
Hi @aaronchlam,

thanks for submitting this PR! 
Stratosphere does not support handling of null-valued data objects. So, it is fine that the tests do not include null-valued Strings.
I am currently on easter vacations and will have a detailed look at the PR next week.

StephanEwen:
Thanks, that one is good to merge.

fhueske:
I will merge this PR

fhueske:
Merged in 09632ed2372784806e5602f298ae5b7e7cb16259

Issue #675, opened 2014-04-10
Mixing key extractor and field position keys in new JAPI


markus-h:
Currently it is only possible to specify either two position keys or two key extractors in Join and CoGroup in the new JAPI.
It would be convenient to also support a mixture of both, for example like:

```
                DataSet<Tuple5<Integer, Long, Integer, String, Long>> ds = CollectionDataSets.get5TupleDataSet(env);
                DataSet<CustomType> ds2 = CollectionDataSets.getCustomTypeDataSet(env);
                DataSet<Tuple2<Integer, Long>> coGroupDs = ds.coGroup(ds2).where(0).equalTo(new KeySelector<CustomType, Integer>() {
                                    private static final long serialVersionUID = 1L;
                                    @Override
                                    public Integer getKey(CustomType in) {
                                        return in.myInt;
                                    }
                                }).with(new MixedCoGroup());
```

fhueske:
CoGroup has been done and merged into master. Join still needs to be done.

fhueske:
Done with #684 and #715.

PR #715, opened 2014-04-23
Mixed key types for join

markus-h:
See #675
For some reasons one of the tests is already running because copy() in the Avro Serializer is not called for it. I suppose it has something to do with one dataset being very small so that the optimizer decides for some different strategy.
The other test has copy() support for the Avro Serializer pending.


fhueske:
I checked the other tests by forcing the input with the custom type to be the build side of the join. That way, the custom types do not need to be copied.
The test passed.
I will merge this PR.

fhueske:
Merged in d056a43b13e23fb2d2e2db0cf0577f341c0daa09

PR #718, opened 2014-04-23
Various minor fixes:

rmetzger:
- better exceptions for faulty temporary directories
- better exception if client can not connect to JobManager

### DIFF
--- a/stratosphere-compiler/src/main/java/eu/stratosphere/compiler/PactCompiler.java
+++ b/stratosphere-compiler/src/main/java/eu/stratosphere/compiler/PactCompiler.java
@@ -1549,7 +1549,9 @@ public class PactCompiler {
 			throw new CompilerException(e.getMessage());
 		}
 		catch (Throwable t) {
-			throw new CompilerException("Error while connecting to the job manager: " + t.getMessage(), t);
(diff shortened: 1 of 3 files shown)


rmetzger:
Also, the PR correctly enforces maven 3.0.3.
I was unable to create a maven enforcer rule that fails on oraclejdk6. (there is a way by writing a custom rule in Java).

stratosphere-compiler/src/main/java/eu/stratosphere/compiler/PactCompiler.java
```diff
@@ -1560,7 +1560,8 @@ private InstanceTypeDescription getInstanceTypeInfo() {
 			}
 		}
 		catch (Throwable t) {
-			throw new CompilerException("Available instances could not be determined from job manager: " + 
+			throw new CompilerException("Cannot connect to the JobManager to determine the available machines. "
```
fhueske: "available machines" -> "available TaskManagers"

fhueske:
I have only one minor comment.
Otherwise good to merge.

StephanEwen:
Agree, good to merge

PR #721, opened 2014-04-25
A Connected Components test case which requires 2 joins with the solution set

vasia:
This is a Connected Components test case, which recomputes only the elements of the solution set whose at least one dependency (in-neighbor) has changed since the last iteration.
It is used to test the multiple joins with solution set feature (see #162).


StephanEwen:
Merged in c8dbe684b229cb463359b819e898de6e7d248f32

Thanks for the contribution!

PR #726, opened 2014-04-28
Divided Reduce and GroupReduce

filiphaase:
Divided the Reduce and the GroupReduce classes from each other. So far Reduce was just wrapped into a GroupReduce. Now there are OptimizerClasses, Pact-Driver,... for both. 

I only use the same drivers for the combine case (PARTIAL_GROUP_COMBINE), since the combine logic is exactly the same for both cases so far. (Therefore both Operators implement the new interface GenericCombine now)


StephanEwen:
Merged in de80196067f6a46699e3ecb9d45ebdc68c5da5e8

PR #728, opened 2014-04-28
added ExecutionEnvironment.registerCachedFile()

zentol:
Instead of

```
l.executePlan(p);
```

you can now write

```
env.registerCachedFile(path, "name");
env.execute();
```


StephanEwen:
Very good!

stratosphere-java/src/main/java/eu/stratosphere/api/java/ExecutionEnvironment.java
```diff
@@ -247,6 +250,17 @@ public JobExecutionResult execute() throws Exception {
 	
 	public abstract String getExecutionPlan() throws Exception;
 	
+	public void registerCachedFile(String filePath, String name){
+		this.cacheFile.add(filePath);
```
StephanEwen: These things are error prone. Can you not use a `Tuple2<String,String>` here?

StephanEwen:
Merged in 06baf89844af2cb07f08a1a55e6c6367afec28d1

Issue #82, opened 2013-09-06
(Hash) Aggregators (fold style)


dimalabs:
This issue depends on issue https://github.com/dimalabs/ozone/issues/81

rmetzger:
@StephanEwen: Can you estimate the amount of work required to implement this?

rmetzger:
I talked with @StephanEwen about it today. It is okay to implement a first version just using a Java-objects based-hash map approach for this.
This would be a matter of hours.

filiphaase:
I'm going to work on this issue next. 
I will start using a Java-objects based-hash map, like proposed by @StephanEwen / @rmetzger.

PR #729, opened 2014-04-28
HashReduce based on java-util-hash-map

filiphaase:
This is a first step for the hash-aggregator, corresponding issue: https://github.com/stratosphere/stratosphere/issues/82

This implementation is a ReduceOperator based on a hash-table, instead of sorting the elements.
It is just a prototype because it is still using the java-util-hash-map, and furthermore the CostEstimation is just 0 for now. 

This PR is based on the PR for dividing the reduce and group reduce operators(https://github.com/stratosphere/stratosphere/pull/726), only commit bb5640a is the HashReduce code. I just opened the PR to get feedback on the current code and how I should proceed with a stratosphere-hash-map implementation and adjusting the CostEstimation.

### DIFF
--- a/stratosphere-compiler/src/main/java/eu/stratosphere/compiler/costs/CostEstimator.java
+++ b/stratosphere-compiler/src/main/java/eu/stratosphere/compiler/costs/CostEstimator.java
@@ -158,13 +158,15 @@ public abstract class CostEstimator {
 		case FLAT_MAP:
 			
 		case ALL_GROUP:
-			// this operation does not do any actual grouping, since every element is in the same single group
+		case ALL_REDUCE:
+			// this operations does not do any actual grouping, since every element is in the same single group
 			
 		case CO_GROUP:
 		case SORTED_GROUP:
+		case SORTED_REDUCE:
 			// grouping or co-grouping over sorted streams for free
 			
-		case PARTIAL_GROUP:
+		case PARTIAL_GROUP_COMBINE:
 			// partial grouping is always local and main memory resident. we should add a relative cpu cost at some point
 		
 		case UNION:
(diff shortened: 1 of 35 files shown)


filiphaase:
Hey, I updated this PR with the Hash-Reduce Code using the CompactingHashTable.

Some things about the current state, where I don't know how to proceed:
- I'm using the HashTableProber and TypePairComparator with just one type since it seems to implement exactly what I needed for this task: A getMatch method and an efficient updateMatch method. I talked to @StephanEwen about the efficient updateMatch, isn't this exactly what we need? And is it from design perspective okay to just use it with a TypePairComparator with one type?
- I looked into the cost-estimations and don't know how to adjust them appropriate for the HashReduce, therefore I added no cost for now.
- For testing I used the existing ReduceITCase tests and assured that the hash-reduce driver is used by commenting the sorted-reduce strategy out in the ReduceNode. @rmetzger said that there was a way to assure that a certain strategy is used over configuration-parameter. Is that also working with the new API? If so, I could include it in the ReduceITCase to test both strategies all the time. 

I looked shortly into the performance of the hashreduce to get a feeling for it, therefore I measured the times of the Drivers and the total execution time of the job for a WordCount.
For ~60MB input size of random text with a few runs, the results on my machine were:
- Hash-Reduce run(): ~3.5s / total time: ~17s
- For the "Sorted-Reduce" there are two Drivers:
  -> SynchronousChainedCombineDriver sortAndCombine(): ~6s
  -> ReduceDriver run() ~0.5s
  -> total time: 16s
  So the pure driver code runs around 50% quicker, but the total time for the job actually gets a little bit worse. I assumed the reason for this is that the sorted-reduce is executed chained, could that be the reason?

uce:
I think this is on halt. Please re-open in new repo if there is any progress.

PR #730, opened 2014-04-28
Cache: exists() called for files / relative path support

zentol:
The distributed cache now supports relative paths, and checks whether the file actually exists.

### DIFF
--- a/stratosphere-core/src/main/java/eu/stratosphere/api/common/Plan.java
+++ b/stratosphere-core/src/main/java/eu/stratosphere/api/common/Plan.java
@@ -303,25 +303,23 @@ public class Plan implements Visitable<Operator> {
 	 *  register cache files in program level
 	 * @param filePath The files must be stored in a place that can be accessed from all workers (most commonly HDFS)
 	 * @param name user defined name of that file
+	 * @throws java.io.IOException
 	 */
-	public void registerCachedFile(String filePath, String name) throws RuntimeException {
+	public void registerCachedFile(String filePath, String name) throws RuntimeException, IOException {
 		if (!this.cacheFile.containsKey(name)) {
 			try {
-				FileSystem fs = FileSystem.get(new URI(filePath));
-				if (fs.exists(new Path(filePath))) {
-					this.cacheFile.put(name, filePath);
-				} else {
-					throw new RuntimeException("File "+filePath+" doesn't exist.");
+				URI u = new URI(filePath);
+				if (!u.getPath().startsWith("/")) {
+					u = new URI(new File(filePath).getAbsolutePath());
 				}
-			} catch (IOException ex) {
-				File f = new File(filePath);
-				if (f.exists()) {
-					this.cacheFile.put(name, f.getAbsolutePath());
+				FileSystem fs = FileSystem.get(u);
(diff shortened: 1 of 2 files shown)


rmetzger:
Checking if the file exists is a good idea, but this code here only works for local files. 
I think users can also add files such as "hdfs:///blabla/" or "file:///tmp/abc". 
But you can probably do the same check using the `FileSystems` class of Stratosphere.

zentol:
good call. it now uses the FileSystem class, and if that fails (which happens if you pass something like "src/main/...") it falls back to File class.

zentol:
hmm its not really a nice solution, Ill keep working on it.

stratosphere-core/src/main/java/eu/stratosphere/api/common/Plan.java
```diff
@@ -292,15 +298,29 @@ public void accept(Visitor<Operator> visitor) {
...
+			try {
+				URI u = new URI(filePath);
+				if (!u.getPath().startsWith("/")) {
+					u = new URI(new File(filePath).getAbsolutePath());
+				}
+				FileSystem fs = FileSystem.get(u);
+				if (fs.exists(new Path(u.getPath()))) {
+					this.cacheFile.put(name, u.toString());
+				} else {
+					throw new RuntimeException("File " + u.toString() + " oesn't exist.");
```
rmetzger: typo

stratosphere-core/src/main/java/eu/stratosphere/api/common/Plan.java
```diff
@@ -292,15 +298,29 @@ public void accept(Visitor<Operator> visitor) {
...
+				}
+			} catch (URISyntaxException ex) {
+				throw new RuntimeException("Invalid path: " + filePath);
```
rmetzger: Can you add `ex` as a second argument to the RuntimException? It does not harm to show the full stack trace.

stratosphere-tests/src/test/java/eu/stratosphere/test/distributedCache/DistributedCacheTest.java
```diff
@@ -133,7 +133,11 @@ protected void preSubmit() throws Exception {
 	@Override
 	protected Plan getTestJob() {
 		Plan plan =  getPlan(1 , textPath, resultPath);
-		plan.registerCachedFile(cachePath, "cache_test");
+		try {
+			plan.registerCachedFile(cachePath, "cache_test");
+		} catch (IOException ex) {
+			ex.printStackTrace();
```
rmetzger: Mhhh .. not so nice ;) Can you fail the test in that case? Or rethrow the exception?

StephanEwen: Agree, test cases need to fail on exceptions!

StephanEwen:
Looks good to me.

@zentol Is this ready, or do you want to further add to it?

zentol:
its ready.

StephanEwen:
Merged in 8bc853ac71c5abe6476693360043f518e8c4c44e

PR #737, opened 2014-04-29
Unit tests to check for valid parameters of cogroup

markus-h:
See #548
I basically copied the test for join and used it also for cogroup.
I also activated the tests for mixed key types.


fhueske:
Will merge after rebase passed Travis.

fhueske:
Merged

Issue #739, opened 2014-04-29
IllegalMonitorStateException in TaskManager.


rmetzger:
The code has been introduced by commit 2cbb93a436eaa78591677ca5ac3886f94471f485.

Snip:

``` java
// Create a new task manager object
try {
    new TaskManager();
} catch (Exception e) {
    LOG.fatal("Taskmanager startup failed: " + e.getMessage(), e);
    System.exit(FAILURE_RETURN_CODE);
}

// park the main thread to keep the JVM alive (all other threads may be daemon threads)
try {
    new Object().wait(); // line 421
} catch (InterruptedException ex) {}
```

The exception

```
java.lang.IllegalMonitorStateException
    at java.lang.Object.wait(Native Method)
    at java.lang.Object.wait(Object.java:502)
    at eu.stratosphere.nephele.taskmanager.TaskManager.main(TaskManager.java:421)
    at eu.stratosphere.yarn.YarnTaskManagerRunner$1.run(YarnTaskManagerRunner.java:55)
    at java.security.AccessController.doPrivileged(Native Method)
    at javax.security.auth.Subject.doAs(Subject.java:357)
    at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1471)
    at eu.stratosphere.yarn.YarnTaskManagerRunner.main(YarnTaskManagerRunner.java:51)
```

I think you can only call `.wait()` inside a synchronized block?

StephanEwen:
Right. Hmmm, I thought I tested it with start-cluster, but apparently I did not. This is pretty critical.

rmetzger:
Fixed in #740

Issue #736, opened 2014-04-28
Fix temporary directory behavior of YARN and .yarn-jobmanager permissions for different users.


rmetzger:
yarn-session: root
strato client: not-root.

PR #740, opened 2014-04-29
Many small YARN improvements.

rmetzger:
This pull request fixes some YARN-related bugs:
- Fix temporary directory behavior of YARN and .yarn-jobmanager permissions for different users. #736
- YARN session client should use "fs.hdfs.hadoopconf" if property set in conf #733
- warning if the user is using "file:" as the distributed file system
- Don't show JobManager URL in stratosphere-client if submitting to YARN. 
- Suggested fix for: IllegalMonitorStateException in TaskManager. #739

### DIFF
--- a/stratosphere-addons/yarn/pom.xml
+++ b/stratosphere-addons/yarn/pom.xml
@@ -26,7 +26,13 @@
 				</exclusion>
 			</exclusions>
 		</dependency>
-
+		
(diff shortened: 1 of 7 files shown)


rmetzger:
Pushed some more improvements in 68100e5c9414da6826ab04fc1d40f2ea6e8744c2

PR #743, opened 2014-04-30
Refactor test utils into own package

aljoscha:
When trying to use packages that use stratosphere-tests in
stratosphere-examples you used to get a circular dependency. Now
packages can use the testing utilities without pulling in all the
transitive dependencies of stratosphere-tests.


StephanEwen:
Sounds like a good idea to me.

While at that, can we get rid of the old TestBase?

aljoscha:
OK, i'll work on removing that.

aljoscha:
This has grown a bit. Now it's the refactoring of the testing utils into a separate package plus the removal of the old TestBase and the renaming of TestBase2 to RecordAPITestBase.

StephanEwen:
Looks good, will merge.

StephanEwen:
Merged in dcda680f89dd055800dda89c6bd3b16567d4f35c

Issue #748, opened 2014-05-02
inconsistent read/write in StringValue


zentol:
the static writeString() method in StringValue adds an offset of 1 to differentiate between null and "". while the readString() method accounts for this, the non-static read()/write() methods do not.

this currently causes issues in the StringComparator, which uses writeString() to serialize but read() to deserialize, causing it to read an additional character.

is there a reason why these methods behave differently?

StephanEwen:
The reason is that you can statically read a null. But if you have already
a StringValue instance which deserializes itself  from the stream, it is
naturally not null.

zentol:
would it be a problem to add the offsets to the non static methods?

StephanEwen:
I think you can add the offsets to the non static method as well

StephanEwen:
Fixed in d2a5178014470591c155989158d3d8629a6ee0ef

Issue #738, opened 2014-04-29
Implement comparators for Value types


StephanEwen:

markus-h:
I think  #741 is quite important for a good implementation.

markus-h:
This issue can be closed. Implemented in #754 merged in 808ae18f66614ba4bc81c2f3007f913e04ce7b80 and 01583434ea14d1a73b538d72940cc5c77e009153.

PR #749, opened 2014-05-02
StringComparator incompatible read/write

zentol:
The StringComparator uses a mixture of static and non-static read/write methods which are not compatible with each other due to different offsets.

This PR makes sure that it only uses the static methods (similar to the StringSerializer).

this fixes the issue mentioned in #748 .

### DIFF
--- a/stratosphere-core/src/main/java/eu/stratosphere/api/common/typeutils/base/StringComparator.java
+++ b/stratosphere-core/src/main/java/eu/stratosphere/api/common/typeutils/base/StringComparator.java
@@ -31,10 +31,6 @@ public final class StringComparator extends BasicTypeComparator<String> {
 	private static final int HIGH_BIT2 = 0x1 << 13;
 	
 	private static final int HIGH_BIT2_MASK = 0x3 << 6;
-
-	
-	private final StringValue holder1 = new StringValue();
-	private final StringValue holder2 = new StringValue();
 	
 	
 	public StringComparator(boolean ascending) {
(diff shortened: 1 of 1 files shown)


StephanEwen:
The purpose of the StringValue is to support mutable objects that can be reused. Otherwise, one could use String directly. This purpose is voided with this implementation.

zentol:
I'm a bit confused. The StringComparator is only used for strings, correct? and not for StringValues, which is why #738 exists. So why is this change connected to the general purpose of StringValues?

zentol:
instead of the StringComparator i now modify the StringValue class, adding an offset to the read()/write() method. They now behave more similarly to the StringSerializer methods, fixing the issue in the StringComparator just as well.

zentol:
hmm this turned out trickier than i thought. i tried different positions for adding the offsets, but even the "best" version broke several tests. :/

eu.stratosphere.RecordTest.blackBoxTests
eu.stratosphere.pact.runtime.io.ChannelViewsTest (timeout)

zentol:
```
Running eu.stratosphere.pact.runtime.hash.HashTableITCase

No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.

The build has been terminated
```

can this be related to this PR? it only happens on a single build.

rmetzger:
No, its a common error.

On Tue, May 6, 2014 at 2:04 PM, zentol notifications@github.com wrote:

> Running eu.stratosphere.pact.runtime.hash.HashTableITCase
> 
> No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.
> 
> The build has been terminated
> 
> can this be related to this PR? it only happens on a single build.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/stratosphere/stratosphere/pull/749#issuecomment-42294207
> .

## 

StephanEwen:
Merged in d2a5178014470591c155989158d3d8629a6ee0ef

PR #750, opened 2014-05-02
Add getSolutionSet() and .getWorkset() for delta iterations

aljoscha:
Before, the result of dataset.iterateDelta() would be the workset and
the solution set could be obtained via iteration.getSolutionSet().

This is counterintuitive for two reasons: 1. in BulkIteration the result of
the dataset.iterate() is the solution set. 2. you call .iterateDelta()
on what is to be the initial solution set and pass the workset as
an argument, so it makes sense that what you get as a result is also the
solution set placeholder.

### DIFF
--- a/stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/graph/ConnectedComponents.java
+++ b/stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/graph/ConnectedComponents.java
@@ -15,7 +15,7 @@ package eu.stratosphere.example.java.graph;
 
 import eu.stratosphere.api.common.ProgramDescription;
 import eu.stratosphere.api.java.DataSet;
-import eu.stratosphere.api.java.DeltaIterativeDataSet;
+import eu.stratosphere.api.java.DeltaIteration;
 import eu.stratosphere.api.java.ExecutionEnvironment;
 import eu.stratosphere.api.java.aggregation.Aggregations;
 import eu.stratosphere.api.java.functions.FlatMapFunction;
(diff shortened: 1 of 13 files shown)


StephanEwen:
I am not sure about this. The reason why we did it as we did is that you cannot really do anything on the solution set, except joining with it. So you cannot take the set and run some functions on it, as one might expect seeing the syntax of the bulk iterations. So

```
DeltaIterativeSet<...> set = ...
set.map.reduce.something();
```

would not work then. The way we have it right now, the delta iterations are like the bulk ones, except that you can access a state in addition.

aljoscha:
Hmm ok, it's just that I immediately was confused about what the result from the iterateDelta() call was supposed to be. Maybe we should have .getWorkset() and .getSolutionSet() then?

StephanEwen:
We could do that. That would at least make it very clear what is what.

The way it is now is with the patch confusing to me as well, to be honest.

aljoscha:
Ok, then I will change this and update the pull request.

aljoscha:
I updated it with the following commit message:

Before, the result of dataset.iterateDelta() would be the workset and
    the solution set could be obtained via iteration.getSolutionSet().

```
This is confusing since you are not told what the result of
.iterateDelta() is supposed to be. Now, when using the result an
exception is thrown during pre-flight that advises the user to use
the two methods to get the correct dataset (work, solution).
```

StephanEwen:
Would it make sense to let the `DeltaIterativeDataSet` not inherit from `DataSet` such that you cannot use it directly? Also, we should rename it then, many to simply DeltaIteration.

aljoscha:
Yes, you are right. I did that now.

StephanEwen:
Good to merge.

StephanEwen:
Merged in 7f32534cf1110e489f38befe1128aba6e42c97a3

Issue #744, opened 2014-04-30
Implement tests für Basic- and TupleComparators


zentol:
@filiphaase and myself are working on this.  

Regarding the TupleComparator, how exactly is this one specified?
some specific things i wonder about whether they are intended:

a)
if you look at the following code taken from the compare method:

```
int i = 0;
try {
    for (; i < this.keyPositions.length; i++) {
        int cmp = this.comparators[i].compare(firstSource, secondSource);
        if (cmp != 0) {
            return cmp;
        }
    }
    return 0;
}
```

it is evident that the _values_ of the keypositions arent really used, but only the _number_ of positions. this means that the comparison only works if the first n fields are used as keys without skipping a field. (and match the type)

b)
if one uses all fields as keys, you hit an EOFException.

StephanEwen:
Okay, not using the field positions is clearly a bug ;-)

I do not understand why we can hit an EOFException?

I asume one problem here is that we need to deserialize the tuple if we
want to access the individual fields.

zentol:
could we pass a comparator for every field, regardless of whether or not it is used as a key (or pass _any_ kind of information covering the type of all fields)? we could then adjust the offsets accordingly and access the fields without deserializing.

zentol:
clarification: only if a String is used as a key an EOFException occurs.

zentol:
lets say you have two tuple2<Integer,Integer>. if the tuplecomparator returns a result on the first field, is it supposed to skip the remaining bytes of the second integer? (because it doesn't, which breaks my testcases atm)

StephanEwen:
Yes, I see. That is specific to the way the tests happen, I guess.

The sorter code that runs this part seeks the views to the beginning of the
record before starting the comparisons. Can you do something similar?

zentol:
okay, if thats how it works then i will adjust the tests accordingly :)

StephanEwen:
If you compare on the second field of the tuple, the sorter will seek to
the beginning of the tuple as well. That means you may need to deserialize
the leading fields, unfortunately.

PR #752, opened 2014-05-05
Fix for Tuple Comparator

filiphaase:
I thought about a fix for the bug in the TupleComparator (mentioned in https://github.com/stratosphere/stratosphere/issues/744#issuecomment-42018637) which currently leads to the wrong result in sorts. 

My idea was to change the TupleSeriailizer to serialize the fields in the same order as the comparator needs them. Then the current implementation of the TupleComparator runs without any changes and without any serializiation overhead at runtime. Since it wasn't much code I did just write and tested it and it seems to work. Or do you prefer an other fix?

If this fix is okay, I got an other question: I added the KeyPositions to the TypeInformation and set them in the JavaAPiPostPass. I get the KeyPositions over channel.getLocalStrategyKeys() and pass it to the TypeInformation, but there are also the channel.getShipStrategyKeys(). In my tests the ShipStrategyKeys were a subset of the LocalStrategyKeys therefore I took the LocalStrategyKeys, but is this always the case?


StephanEwen:
This pull request would introduce quite a bit of instability and potential errors for tasks with multiple outputs.

@filiphaase and me decided to fix this in a different way.

PR #758, opened 2014-05-06
Add count method to DataSet and implement CountOperator

uce:
At the request of @twalthr. This is the count operator I've implemented some time ago to get the to know the new Java API. It introduces `DataSet.count()`, which is executed as a map (to ones) and reduce (sum up the ones). I initially didn't do the PR, because of the following problem: empty DataSets don't work as the first map won't have any input to operate on.

If more people think that we should include this operator we can think about a possible solution to the problem.

### DIFF
--- a/stratosphere-java/src/main/java/eu/stratosphere/api/java/operators/CountOperator.java
+++ b/stratosphere-java/src/main/java/eu/stratosphere/api/java/operators/CountOperator.java
@@ -1,0 +1,68 @@
+/***********************************************************************************************************************
+ *
+ * Copyright (C) 2010-2014 by the Stratosphere project (http://stratosphere.eu)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ **********************************************************************************************************************/
+package eu.stratosphere.api.java.operators;
+
+import eu.stratosphere.api.common.operators.Operator;
+import eu.stratosphere.api.java.DataSet;
+import eu.stratosphere.api.java.functions.MapFunction;
+import eu.stratosphere.api.java.functions.ReduceFunction;
+import eu.stratosphere.api.java.operators.translation.PlanMapOperator;
+import eu.stratosphere.api.java.operators.translation.PlanReduceOperator;
+import eu.stratosphere.api.java.typeutils.BasicTypeInfo;
+
+/**
+ * The CountOperator will be translated to a map and reduce function.
+ * <p/>
+ * The map function will map every input element to a 1 and the following reduce will sum up all ones.
+ *
+ * @param <IN> The type of the data set filtered by the operator.
+ */
+public class CountOperator<IN> extends SingleInputUdfOperator<IN, Long, CountOperator<IN>> {
+
+	public CountOperator(DataSet<IN> input) {
+		super(input, BasicTypeInfo.LONG_TYPE_INFO);
+	}
+
+	@Override
+	protected Operator translateToDataFlow(Operator input) {
+		PlanMapOperator<IN, Long> countMapOperator = new PlanMapOperator<IN, Long>(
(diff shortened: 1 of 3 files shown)


twalthr:
I think it would be great if count of 0 records would also be possible. Couldn't we just create a default record containing 0 and make a union?

uce:
I saw that some internals changed (`UnaryNodeTranslation` got removed; I think its better now). I just updated this PR to reflect that as the initial one didn't compile any more.

---

@twalthr: Good idea. How do we distribute the other DataSet?

StephanEwen:
I like the idea. Can we come up with a design that also supports zero size sets?

Also, do we have the an idea of how we want to get about methods that we want to add in the future for computing scalar values, or values that are pulled back to the client?

Would the following work:
- `DataSet.count()` --> a DataSet with one value
- `DataSet.collectCount()` --> a long with the count in the DataSet

We may also think about introducing a subclass of DataSet: ScalaValueSet or so, which we use for data sets that we know to contain only value. Examples are the result of `count()`, `aggregate()`, or `reduce()`. The latter two without grouping.

StephanEwen:
I think we should merge this for 0.5.1 or 0.6

rmetzger:
This feature was also requested by a user on the mailing list.

stratosphere-java/src/main/java/eu/stratosphere/api/java/operators/CountOperator.java
```diff
@@ -0,0 +1,68 @@
...
+
+	public CountOperator(DataSet<IN> input) {
+		super(input, BasicTypeInfo.LONG_TYPE_INFO);
+	}
+
+	@Override
+	protected Operator translateToDataFlow(Operator input) {
+		PlanMapOperator<IN, Long> countMapOperator = new PlanMapOperator<IN, Long>(
+				new CountingMapUdf<IN>(), "Count: map to ones", getInputType(), BasicTypeInfo.LONG_TYPE_INFO);
+		countMapOperator.setInput(input);
```
StephanEwen: I think you need to set the parallelism here. Have a look at how the operators that inject key extracting mappers do it.

stratosphere-java/src/main/java/eu/stratosphere/api/java/operators/CountOperator.java
```diff
@@ -0,0 +1,68 @@
...
+		PlanReduceOperator<Long> countReduceOperator = new PlanReduceOperator<Long>(
+				new CountingReduceUdf(), new int[0], "Count: sum up ones", BasicTypeInfo.LONG_TYPE_INFO);
+		countReduceOperator.setInput(countMapOperator);
+
+		return countReduceOperator;
+	}
+
+	// -----------------------------------------------------------------------------------------------------------------
+
+	public static class CountingMapUdf<IN> extends MapFunction<IN, Long> {
```
StephanEwen, line 80: I am getting serializability warnings here. Can you add the serial version UID?

StephanEwen:
Looks good to me, modulo some inline comments.

I think the counting will happen directly on the intermediate results in the next versions, so don't worry too much about the 0 count. You can union in a "size 1" data set with a count of 0. It will freak you out once you look at the execution plan, though ;-)

uce:
I've addressed Stephan's comments and added a dummy source with a single 0 element to ensure that the reducer receives at least a single element (fold style initial value).

I didn't do this initially, because I've assumed that the CollectionInputFormat does _not_ work in distributed mode, but @aljoscha challenged this.

uce:
I've tested it on the test cluster and it works. Travis fails due to time outs, but it should be fine.

StephanEwen:
We could add a very simple runtime operator for count. That would save us
from doing the trick of song a union with zero.

aljoscha:
At what level? Do you mean a custom implementation of a `PactDriver`? Because the implementation would have to be tailored to a specific API (Java NEW, Record, ...), right?

StephanEwen:
We could make it as a special case of the GroupReduce (or AllReduce) that
gets the iterator also in cases where the iterator is empty. Then we would
not need to make it API specific, we could still work with the
TypeInformation and serializers.

StephanEwen:
What do we do with this?

uce:
Either
1. just merge it like it is, create a new issue and refactor later, or
2. point me to the place where you would do the special case.

I think it would make sense to go the extra mile and offer a fold style reduce, where you can specify the initial value. The count operator would than be a summing fold with initial value of 0.

StephanEwen:
I very much like the fold-style reduce idea.

uce:
I will reintroduce this PR in the new repo after finishing apache/incubator-flink#20 and use the implemented reduce with initial value.

Issue #761, opened 2014-05-06
Inconsistent Boolean comparison


zentol:
The BooleanComparator.compare() works the opposite way of Boolean/BooleanValue.compareTo().

Consider the following piece of code:

```
public static void main(String[] args) throws IOException{
        Boolean x = true;
        Boolean y = false;
        System.out.println(x.compareTo(y));

        BooleanValue t = new BooleanValue(true);
        BooleanValue s = new BooleanValue(false);
        System.out.println(t.compareTo(s));

        BooleanSerializer bs = new BooleanSerializer();
        TestOutputView out1 = new TestOutputView();
        bs.serialize(true, out1);
        TestInputView in1 = out1.getInputView();
        TestOutputView out2 = new TestOutputView();
        bs.serialize(false, out2);
        TestInputView in2 = out2.getInputView();

        System.out.println(new BooleanComparator(true).compare(in1, in2));
    }
```

This prints

```
1
1
-1
```

intended?

zentol:
the basicTypeComparator.compareToReference() is also odd. 

```
@Override
    public int compareToReference(TypeComparator<T> referencedComparator) {
        int comp = ((BasicTypeComparator<T>) referencedComparator).reference.compareTo(reference);
        return ascendingComparison ? comp : -comp;
    }
```

shouldnt it compare its own reference with the passed one instead of the other way around?

markus-h:
I think everything should be consistent to the Boolean implementation of Java, so BooleanComparator and the basicTypeComparator seems to be wrong for me.

StephanEwen:
The TypeComparators kind of swap the order of arguments internally. I think
all of the TypeComparators do it like that (but please verify that
statement).

If you want to change it, you would need to adjust every place where it is
invoked.

zentol:
only the boolean comparator works like this.
exempt from the int comparator (i did check them all):

```
int i1 = firstSource.readInt();
int i2 = secondSource.readInt();
int comp = (i1 < i2 ? -1 : (i1 == i2 ? 0 : 1)); 
```

in contrast the boolean one:

```
final int fs = firstSource.readBoolean() ? 1 : 0;
final int ss = secondSource.readBoolean() ? 1 : 0;
int comp = ss - fs; 
```

StephanEwen:
Okay, then it seems like a bug there. So, do the other comparators return a
consistent result in your sample test?

zentol:
resolved in #764

PR #764, opened 2014-05-06
BasicType-/BooleanComparator consistent comparison

zentol:
#761

### DIFF
--- a/stratosphere-core/src/main/java/eu/stratosphere/api/common/typeutils/base/BooleanComparator.java
+++ b/stratosphere-core/src/main/java/eu/stratosphere/api/common/typeutils/base/BooleanComparator.java
@@ -34,7 +34,7 @@ public final class BooleanComparator extends BasicTypeComparator<Boolean> {
 	public int compare(DataInputView firstSource, DataInputView secondSource) throws IOException {
 		final int fs = firstSource.readBoolean() ? 1 : 0;
 		final int ss = secondSource.readBoolean() ? 1 : 0;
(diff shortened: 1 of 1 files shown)


StephanEwen:
The current order is important to make sure that a requested ascending comparison is really ascending. If you switch the order in this one place, then a group order of ascending over boolean will no longer be a ascending, because the code assumes the current behavior.

zentol:
This is actually a larger issue than i expected. for _any_ class extending from TypeComparator, or any of its subclasses, it is basically a coin flip how compareToReference is implemented. Even classes at the very top (for example the TupleComparator) can disagree.

StephanEwen:
It may be that this comes from the fact that different people implemented those comparators. Since they are not yet tested, the build accepted the behavior.

To  fix this, have a look at the way the `RecordComparator` does it. That one is consistent with the way the runtime expects the behavior to correctly produce ascending/descending orders.

zentol:
it seems like i made a mistake and created an issue were none actually exists. 

The TypeComparator.cTF() methods are fine. The BooleanComparator should be fixed though.

Turns out there are only 2 groups of Comparators that uses a different order than the TypeComparator, namely TupleComparators and a few in the test packages of runtime. 
Since the TupleComparators only delegate the comparisons to others, no change is necessary here. 
The ones used in the tests _could_ be changed, but i doubt its necessary.

StephanEwen:
merged in 8e72db9f1935ce6ee063f5a23cbc544cc1dc3210

Issue #769, opened 2014-05-07
BulkIteration does not work correctly if not all records are consumed


twalthr:
This is a old but still existing bug. A similar bug already occurred with DeltaIteration.

A test case that shows the bug is following in a minute.

StephanEwen:
Fixed in d56565c31e3cf4af8888e77a4de9e0b0dfa08dea

PR #770, opened 2014-05-07
Test case for showing the still existing BulkIteration bug.

twalthr:
Shows the bug mentioned in #769.

The lines 

``` java
while(records.hasNext()) {
    records.next();
}
```

should not be necessary.


StephanEwen:
Okay, there is a different issue than the last time.
This here is about non exhausted non-iterative inputs. The cache caches "to the side", so it only caches what the first iteration consumed. The second iteration has only that data available.

StephanEwen:
Merged in 982e1e0905dcdce71977cf408b6fac97b6f8f45c

PR #771, opened 2014-05-07
More detailed error messages for CompactingHashTable

rwaury:
Every time compaction or insertion fails because of too few MemorySegments exception now contains number of partitions, number of segments in largest and smallest partition and number of segments used for buckets.


StephanEwen:
Wow, that was quick, thanks!

StephanEwen:
Merged in 34ab689ad210b7448d16e647296dc6c671c9bf76

Issue #767, opened 2014-05-07
Implement extractInputFormatTypes() for TypeExtractor 


twalthr:
TypeExtractor.extractInputFormatTypes() is not implemented yet.

StephanEwen:
I will take a quick stab at this.

twalthr:
I also can do it since I need it later this afternoon for the Hadoop Input Format.

StephanEwen:
Sounds good.

StephanEwen:
Fixed in #776

Issue #747, opened 2014-05-02
Implement a special type for Hadoop Writables


twalthr:
As mentioned in #579:

Please add also Types that implement Hadoop's Writable interface. They should have a special serializer that calls the write() and readFields() method.

A discussion about the necessary dependencies can be found on the mailing list:
https://groups.google.com/forum/#!searchin/stratosphere-dev/writable/stratosphere-dev/OybICx1IxLM/SeLUp2wH0N4J

twalthr:
Actually most `Writables` are AtomicTypes, right? Would the following lines of code ok or should I not treat them as AtomicTypes?

``` java
@Override
    public TypeComparator<T> createComparator(boolean sortOrderAscending) {
        if(Comparable.class.isAssignableFrom(typeClass)) {
            return new WritableComparator<T>(sortOrderAscending);
        }
        else {
            throw new UnsupportedOperationException("Writable does not implement Comparable interface.");
        }
    }
```

StephanEwen:
Yes, we see writables as atomic types.

But please check that when creating the comparator from the type
information, such that you get the error early.

StephanEwen:
Can you keep this in the addons/hadoop-compatibility package? I think it would be nice if the stratosphere-java project had no hadoop dependency.

fhueske:
Done with PR #776.

PR #776, opened 2014-05-08
Introduction of a Writable type

twalthr:
This PR introduces a Writable stratosphere type. All related classes are updated to support Writables.

Additionally some method names of the TypeExtractor have been refactored.

This PR implements #767 and #747.

### DIFF
--- a/stratosphere-java/src/main/java/eu/stratosphere/api/java/ExecutionEnvironment.java
+++ b/stratosphere-java/src/main/java/eu/stratosphere/api/java/ExecutionEnvironment.java
@@ -277,7 +277,7 @@ public abstract class ExecutionEnvironment {
 			@SuppressWarnings("unchecked")
 			TypeInformation<X> producedType = (inputFormat instanceof ResultTypeQueryable) ?
 					((ResultTypeQueryable<X>) inputFormat).getProducedType() :
-					TypeExtractor.extractInputFormatTypes(inputFormat);
+					TypeExtractor.getInputFormatTypes(inputFormat);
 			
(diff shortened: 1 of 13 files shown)


rmetzger:
Looks good (except for some indentation in `WritableTypeInfo`, but thats only a minor issue)
(note that I'm not an expert with the serialization details, so there might be still some issues)

StephanEwen:
All builds fail because some files have no proper Apache headers...

StephanEwen:
The WritableSerializer throws an exception in the copy method.
Tests for the WritableSerializer are missing.

StephanEwen:
The Serializer and the Comparator instantiate new Object everytime they internally deserialize for comparison.

StephanEwen:
The Comparator fails the `duplicate()` method. As such, all runtime use of the comparator would fail.

StephanEwen:
The `WritableComparator` caches only a Java reference of the reference Object. That is not enough for mutable types.

Have a look at the `ValueComparator`. You can pretty much copy that one, actually, since the `Value` interface and the `Writable` interface are almost identical.

StephanEwen:
Most of this would have been caught in test cases...

twalthr:
Sorry for the bad pull request. I reworked it completely so that it now works similar to the ValueComparator and Serializer. Two test classes test the new implementation.

fhueske:
Looks good to me. Thanks!

fhueske:
Merged in a07dba6f1d8929ddf21e55d1bc1471045ea2a087

PR #780, opened 2014-05-09
 Introduce TypeInformation at lowest Operator level

aljoscha:
This touches a LOT of stuff but will allow the removal of the separate
JavaAPI plan operators and other redundant code.

There is a new RecordTypeInfo that operators create behind the scenes.
The *Source and *Sink operators are now Record specific as to not break
existing programs. There are agnostic base classes with the *Base
suffix. Same is true for the iteration operators. Those operators and
sources/sinks are now in stratosphere-java, only the agnostic ones are
in stratosphere-core.


aljoscha:
Please do not merge this, I think I came upon a way to also add input type information in a not to annoying way. Let me see how that goes.

aljoscha:
Instead of adding only a TypeInformation<T> to Operator I would like to create a new OperatorType (plus BinaryOperatorType and UnaryOperatorType) that gets attached to an Operator. This can then hold the information that we want operators to have. Otherwise, this stuff will keep blowing up all the constructors if we decide to add input type information or other stuff later.

Alternatively, I could extend the user code wrapper to hold this information, similarly to how the scala API does it.

What do you think?

(The only problem I see with the user code wrapper approach is that not all types of operators have user code attached. They could retrieve it from the actual user code operator that represents their input.)

aljoscha:
I added another commit that introduces OperatorInformation, so that they can have more than just a TypeInformation for the output type.

aljoscha:
I squashed it into one commit so that it is easier for people to see the differences.

aljoscha:
Rebased on top of current master.

aljoscha:
Rebased it on top of current master again. This is getting somewhat annoying, because people keep adding stuff to the now-removed Plan*Operators.

StephanEwen:
Don't rebase it again. Will merge it today...

stratosphere-addons/hadoop-compatibility/src/main/java/eu/stratosphere/hadoopcompatibility/example/WordCountWithHadoopOutputFormat.java
```diff
@@ -79,8 +79,8 @@ public void map(Record record, Collector<Record> collector) {
 	 * Sums up the counts for a certain given key. The counts are assumed to be at position <code>1</code>
 	 * in the record. The other fields are not modified.
 	 */
-	@Combinable
-	@ConstantFields(0)
+	@GroupReduceFunction.Combinable
```
StephanEwen: That does not seem right. It previously was an annotation from the Record-API, not it is one from the new JAVA API.

StephanEwen: Just leave me a note whether I see that correctly. I am merging it already, so I can fix it just as well...

stratosphere-java/src/main/java/eu/stratosphere/api/java/operators/OperatorTranslation.java
```diff
@@ -85,22 +83,12 @@ else if (dataSet instanceof SingleInputOperator) {
...
-			if (dataSet instanceof UdfOperator<?> ) {
-				SingleInputUdfOperator<?, ?, ?> udfOp = (SingleInputUdfOperator<?, ?, ?>) op;
-				
-				// set configuration parameters
-				Configuration opParams = udfOp.getParameters();
-				if (opParams != null) {
-					singleInDataFlowOp.getParameters().addAll(opParams);
-				}
-				
-				// set the semantic properties
```
StephanEwen, line 100: Did you move the setting of the semantic properties somewhere else, or did it get lost?

stratosphere-java/src/main/java/eu/stratosphere/api/java/operators/OperatorTranslation.java
```diff
@@ -170,8 +149,9 @@ else if (dataSet instanceof DeltaIterationResultSet<?, ?>) {
 		return iterationOperator;
 	}
 	
-	private <D, W> eu.stratosphere.api.common.operators.DeltaIteration translateDeltaIteration(DeltaIterationResultSet<D, W> iterationEnd) {
-		PlanDeltaIterationOperator<D, W> iterationOperator = new PlanDeltaIterationOperator<D, W>(iterationEnd.getKeyPositions(), "Unnamed Java Delta Iteration", iterationEnd.getType(), iterationEnd.getWorksetType()); // always assume 0 as key position?
+	private <D, W> DeltaIterationBase<D, W> translateDeltaIteration(DeltaIterationResultSet<D, W> iterationEnd) {
+		DeltaIterationBase<D, W> iterationOperator = new DeltaIterationBase<D, W>(new BinaryOperatorInformation<D, W, D>(iterationEnd.getType(), iterationEnd.getWorksetType(), iterationEnd.getType()),
+				iterationEnd.getKeyPositions(), "Unnamed Java Delta Iteration"); // always assume 0 as key position?
```
StephanEwen: Why do we always assume position 0 as key? I thought we did not...

stratosphere-addons/hadoop-compatibility/src/main/java/eu/stratosphere/hadoopcompatibility/example/WordCountWithHadoopOutputFormat.java
aljoscha: Yes, you're seeing that correctly.

stratosphere-java/src/main/java/eu/stratosphere/api/java/operators/OperatorTranslation.java
aljoscha, line 100: I think I lost it here. The extraction of the UDF Annotations were moved from the Plan*Operators to the SingleInput and DualInput Operators.

stratosphere-java/src/main/java/eu/stratosphere/api/java/operators/OperatorTranslation.java
aljoscha: That comment is not from me, I don't know why it is there and just left it as is.

aljoscha:
I incorporated @StephanEwen's fixes, rebased it on top of HEAD again and fixed scala warnings.

rmetzger:
Ouch. I was not aware that this pull request is again breaking the existing codebase for our old java API users. `FileDataSink` and `FileDataSource` were moved out of the `eu.stratosphere.api.common.operators` package, right?

Users of the old java API have to change their packages from
0.2 to 0.4
0.4 to 0.5
and since everybody has to:
0.5 to 0.6

I thought we are able to avoid this mess at least for the 0.5 release.

Do you think we should not add this change to the 0.5 release?

StephanEwen:
I think this is only a package move.

If it eases the migration, lets add a copy of the record API FileDataSource
in eu.stratosphere.api.common.operators in addition to the
GenericDataSourceBase.

We could mark it as deprecated and carry it along for one release.

I would prefer that rather than undoing that monster patch. It was a lot of
work and affects so much that is almost immediately goes out of sync.

PR #793, opened 2014-05-12
Disable LocalExecutor when using Client

markus-h:
If you submit a job using the Client, but use the LocalExecutor in your job by accident you can get weird results.
This pull requests makes the LocalExecutor throw an exception if an instance of Client was created before.
I'm not 100% happy with this implementation, since it absolutely prevents you from using LocalExecutor when you also want to create an Client instance, and perhaps there are some cases when you want to do this. What do you think?

### DIFF
--- a/stratosphere-clients/src/main/java/eu/stratosphere/client/LocalExecutor.java
+++ b/stratosphere-clients/src/main/java/eu/stratosphere/client/LocalExecutor.java
@@ -17,10 +17,12 @@ import java.util.List;
 
 import org.apache.log4j.Level;
 
+import eu.stratosphere.api.common.InvalidProgramException;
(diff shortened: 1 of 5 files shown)


StephanEwen:
Can you make the methods to deactivate the local and remote environment protected methods? That way they are not visible for the users. You can still call them from the `ContextEnvironment`. Have a look at how the `ContextEnvironment` is set.

markus-h:
I made the method in ExecutionEnvironment protected and call it from another method in ContextEnvironment, hope that is what you meant.
Currently there is only a method to deactivate the local environmnet.

StephanEwen:
Looks good, will merge this now.

StephanEwen:
merged in 4b8c3e899bb2f2b9ab801a455f7f4b052537af34

Issue #781, opened 2014-05-09
Expose WriteMode in JAPI.


rmetzger:
I just got a mail from a user, asking how to set the Write mode for csv output.
Currently, you have to do the following:

``` java
CsvOutputFormat<ShippingPriorityItem> csvOut = 
    new CsvOutputFormat<ShippingPriorityItem>(new Path(resultPath), "\n", "|");
csvOut.setWriteMode(WriteMode.OVERWRITE);
joined.output(csvOut);
```

Do you think we should add another variant of `writeAsCsv()` with the write mode?

StephanEwen:
Should be easy to add. Do you (or the user) want to make a pull request for
this?

rmetzger:
Sure, I can add it. (The question was, if we want this in general or not. But I think that somebody asked answers already the question.

fhueske:
Fixed with PR #797

PR #797, opened 2014-05-13
Expose WriteMode in JAPI. 

rmetzger:
Fix issue #781.

Pass Configuration parameters to runtime operators in JAPI (+test)

### DIFF
--- a/stratosphere-core/src/main/java/eu/stratosphere/api/common/io/FileOutputFormat.java
+++ b/stratosphere-core/src/main/java/eu/stratosphere/api/common/io/FileOutputFormat.java
@@ -120,7 +120,6 @@ public abstract class FileOutputFormat<IT> implements OutputFormat<IT> {
 		this.outputFilePath = outputPath;
 	}
 	
-	
 	public void setOutputFilePath(Path path) {
 		if (path == null) {
 			throw new IllegalArgumentException("Output file path may not be null.");
--- a/stratosphere-java/src/main/java/eu/stratosphere/api/java/DataSet.java
+++ b/stratosphere-java/src/main/java/eu/stratosphere/api/java/DataSet.java
@@ -631,6 +631,22 @@ public abstract class DataSet<T> {
 	}
 	
 	/**
+	 * Writes a DataSet as a text file to the specified location.<br/>
+	 * For each element of the DataSet the result of {@link Object#toString()} is written.  
+	 * 
+	 * @param filePath The path pointing to the location the text file is written to.
+	 * @param writeMode Control the behavior for existing files. Options are NO_OVERWRITE and OVERWRITE.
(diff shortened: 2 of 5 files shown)


stratosphere-core/src/main/java/eu/stratosphere/api/common/io/FileOutputFormat.java
```diff
@@ -52,7 +52,7 @@
 	
 	// --------------------------------------------------------------------------------------------
 
-	private static WriteMode DEFAULT_WRITE_MODE;
```
fhueske, line 55: Why do you make the default mode public? It should not be possible to overwrite the default imho.

rmetzger, line 55: Good point.. I'll thought its a constant variable (uppercase letters).
I'll change that.

rmetzger:
I addressed @fhueske's comment.

stratosphere-java/src/main/java/eu/stratosphere/api/java/DataSet.java
```diff
@@ -653,13 +653,32 @@ public ExecutionEnvironment getExecutionEnvironment() {
 	 * @see CsvOutputFormat
 	 */
 	public DataSink<T> writeAsCsv(String filePath, String rowDelimiter, String fieldDelimiter) {
-		Validate.isTrue(this.type.isTupleType(), "The writeAsCsv() method can only be used on data sets of tuples.");
-		return internalWriteAsCsv(new Path(filePath), rowDelimiter, fieldDelimiter);
+		return internalWriteAsCsv(new Path(filePath), rowDelimiter, fieldDelimiter, FileOutputFormat.getDefaultWriteMode());
```
fhueske: No sure, if we should do it this way. If no write mode is explicitly specified, the system default write mode will be used. If we decide to introduce something like a client option to set the write mode for a session, this would be overwritten by the system default. 
Why not having two internalWriteAsCsv methods, one with explicit write mode and one without?

fhueske:
Can you add the same functionality for writeAsText() as well?

rmetzger:
Yes, I'll rework the pull request.

On Tue, May 13, 2014 at 11:03 PM, Fabian Hueske notifications@github.comwrote:

> Can you add the same functionality for writeAsText() as well?
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/stratosphere/stratosphere/pull/797#issuecomment-43012501
> .

## 

rmetzger:
Okay, ready to review again.

fhueske:
:+1:
Looks good to me.

rmetzger:
the pr now also contains a fix for https://github.com/stratosphere/stratosphere/issues/811

PR #803, opened 2014-05-13
JAPI String Annotations

skunert:
This PR contains the new String Annotations. Constant sets are not annotated with a string array. 
The parsing is done in the new SemanticPropUtil class. There is also a test case for this class.

What is left is additional functionality like a \* operator. Also it would be nice to be able to add the annotation strings to the operators via functions.
Example:
@ConstantFields({"1->2", "2->3,4,5"})

### DIFF
--- a/stratosphere-java/src/main/java/eu/stratosphere/api/java/functions/FunctionAnnotation.java
+++ b/stratosphere-java/src/main/java/eu/stratosphere/api/java/functions/FunctionAnnotation.java
@@ -13,71 +13,63 @@
  **********************************************************************************************************************/
 package eu.stratosphere.api.java.functions;
 
+import eu.stratosphere.api.common.operators.util.UserCodeWrapper;
+
 import java.lang.annotation.Annotation;
 import java.lang.annotation.ElementType;
-import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
+import java.lang.annotation.Retention;
 import java.util.HashSet;
 import java.util.Set;
-
-import com.google.common.primitives.Ints;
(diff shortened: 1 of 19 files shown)


skunert:
It is now possible to use wildcards. For example

```
"1->*"
"*"
```

Also there is the possibility to add the string annotations directly to the operators.

StephanEwen:
What do the wild-cards mean?

skunert:
```
@ConstantFields{"*"} replaces the AllFieldsConstant
@ConstantFields{"1->*"} means field 1 is forwarded to all output fields
```

skunert:
What do you think about the notation? At the moment it is implemented such that the constantfield annotation takes a String Array. One string for each inputfield. I could also implement that all constantfields are given in one string.

```
1. @ConstantFields({"1->2,3", "2->3,4"})
2. @ConstantFields({"1->2,3; 2->3.4")
```

At the moment I like the second option better and think I should change it, how about you?

StephanEwen:
Would it hurt to support both?

StephanEwen:
I am merging the current state now. So please only append commits, do not squash any.

StephanEwen:
Hmmm... The checkstyle fails for me when compiling the code.

StephanEwen:
Tuples always use zero based indexes, here they use one-based indexes. I have changed that, because it seemed a bit off...

StephanEwen:
I have refactored the fluent API methods on the Unary/Binary-UDF Operators. The long methods kind of break the feeling of the API, so I simplified them.

StephanEwen:
Merged in 1e1101781ef2662a185c3b53befe0a3bb64f550a

Issue #804, opened 2014-05-13
Add PairComparatorTests for TuplePairComparators


StephanEwen:

zentol:
Ive started working on this.

StephanEwen:
Solved in #806

PR #806, opened 2014-05-14
TuplePairComparatorTests

zentol:
#804


StephanEwen:
Okay, I will merge the pull request.

I will change the type of one of the test inputs to Tuple4, though, to make sure that different types are used on the two differnt sides. That verifies that the comparators do not assume equal types.

StephanEwen:
BTW: in the tuple construction:

Instead of 

```
new Tuple3<Integer, Double, Long>(5, 20.0, new Long(15))
```

you can write

```
new Tuple3<Integer, Double, Long>(5, 20.0, 15L)
```

StephanEwen:
Merged in dc72904849873f817f0ecf3eb16871b21d4c7ec6

PR #808, opened 2014-05-14
Remove cloud code

tillrohrmann:
Removed legacy Nephele cloud code. Among others, all InstanceType related code is removed. Furthermore, a simple slot system for the TaskManager is introduced. Each TaskManager has a set of slots. Either the number is set in the configuration or the number of cores is taken as the default value. This implies that the intra-node parallelism was abandoned. Additionally a relative memory assignment was implemented. During the compiling process, for each task its memory share is calculated. Once deployed this value is translated with respect to the available memory on the TaskManager.


StephanEwen:
Nice. That one is also a contender for "biggest pull request" ;-)

I think we will merge that after the 0.5 release, as it changes quite a bit of the model

uce:
+1

I think we should go over this PR as soon as 0.5 is out. Looking forward to it. :-)

StephanEwen:
Yes, it will be one of the first after the 0.5 release...

StephanEwen:
I am trying to merge this in the 0.6-SNAPSHOT master

rmetzger:
+1

Sent from my iPhone

On 31.05.2014, at 19:25, Stephan Ewen notifications@github.com wrote:

I am trying to merge this in the 0.6-SNAPSHOT master

StephanEwen:
This changes a lot:
- The task managers are no longer a single unit that takes "subtasksPerInstance". Instead a task manager as a number of "slots"
- A slot can take one parallel instance from each operator 
- It implies that if you have a cluster with 10 node each with 4 slots (40 slots total) and schedule a job with DOP 20, it will potentially run on 5 machines with 4 tasks on each machine (one per slot). Before it ran on all machines and decreased the per-machine DOP. That was a bit unpredictable, it always depended on how much the current cluster load was.
- This will allow us to simplify the scheduling logic
- It is a prerequisite for our strategy to reduce memory waste

StephanEwen:
That one will keep me busy tonight ;-)

tillrohrmann:
The most of it should only be of cosmetic nature. Next time, I'll split the
changes...

StephanEwen:
The pull request removed the "connection IDs" from the edges. I think they are important to decide what can be multiplexed though the same TCP connection and what cannot.

StephanEwen:
I am re-adding them

uce:
@StephanEwen Yes, the netstack relies on the connection IDs.

uce:
This has been merged in the new repo in commits 86d206c41922a1b7b8c2839b65d3568f9be55e0c...b4b633eab9a70e14d2e0dd5252f4b092a3689093.

PR #809, opened 2014-05-14
Fix for branching merging logic in BulkIterationPlanNode

fhueske:
Fixes a NPE in the BulkIterationPlanNode that occurs when a the output of the bulk iteration is used as a broadcast set with the (branched) input of the bulk iteration.
See the added test case which reproduced the bug.


StephanEwen:
Merged in ff8bceec7fca1394fb3a223018c6c0380d7eab2a

PR #821, opened 2014-05-15
Some minor fixes

rmetzger:
- Check if plan is null and die with nicer exception than NPS
- don't die with indexoutofbounds (or so) exception
- fix IllegalMonitorStateException in JM (see also https://github.com/stratosphere/stratosphere/issues/739)
- show exceptions in log, with INFO log lvl (see mailing list).

### DIFF
--- a/stratosphere-runtime/src/main/java/eu/stratosphere/nephele/taskmanager/TaskManager.java
+++ b/stratosphere-runtime/src/main/java/eu/stratosphere/nephele/taskmanager/TaskManager.java
@@ -556,8 +556,9 @@ public class TaskManager implements TaskOperationProtocol {
 			socket.bind(bindP);
 			socket.connect(toSocket, timeout);
 		} catch (Exception ex) {
+			LOG.info("Failed to determine own IP address from '" + fromAddress + "': " + ex.getMessage());
 			if (LOG.isDebugEnabled()) {
-				LOG.debug("Failed on this address: " + ex.getMessage());
+				LOG.debug("Failed with exception", ex);
 			}
 			connectable = false;
 		} finally {
--- a/stratosphere-addons/yarn/src/main/java/eu/stratosphere/yarn/ApplicationMaster.java
+++ b/stratosphere-addons/yarn/src/main/java/eu/stratosphere/yarn/ApplicationMaster.java
@@ -62,6 +62,7 @@ import eu.stratosphere.nephele.jobmanager.JobManager;
 public class ApplicationMaster {
 
 	private static final Log LOG = LogFactory.getLog(ApplicationMaster.class);
+	private static final int HEAP_LIMIT_CAP = 500;
 	
 	private void run() throws Exception  {
 		//Utils.logFilesInCurrentDirectory(LOG);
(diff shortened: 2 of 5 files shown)


stratosphere-runtime/src/main/java/eu/stratosphere/nephele/taskmanager/TaskManager.java
```diff
@@ -550,8 +550,9 @@ public static boolean tryToConnect(InetAddress fromAddress, SocketAddress toSock
 			socket.bind(bindP);
 			socket.connect(toSocket, timeout);
 		} catch (Exception ex) {
+			LOG.info("Failed to determine own IP address from '" + fromAddress + "': " + ex.getMessage());
```
StephanEwen, line 559: Can you make this 

```
LOG.error("Failed to determine own IP address from '" + fromAddress + "': " + ex.getMessage(), ex);
```

That way it always logs the stack trace in the event of an error (which I think is desirable).

StephanEwen:
Looks good aside from the one inline comment. Good to merge.

stratosphere-runtime/src/main/java/eu/stratosphere/nephele/taskmanager/TaskManager.java
rmetzger, line 559: tl;dr: Users that have TM and JM in different networks will have their logs full with exceptions even in non-error cases.

I do not pass the exception intentionally!
The method is called `tryToConnect()` and nomen est omen. The method exists to determine if it is possible to connect via a network interface to the JobManager address.
So it is very likely (also in cases without errors) that there will be exceptions in the log.
For example the loopback device on linux systems is certainly not able to connect to the JobManager, so it will throw an exception here. Which is not an exception.

The `tryToConnect()` method is only being used if the JobManager and Taskmanager are in different networks. The usual case is that we use the network interface that is in the same network (shares the ip-prefix with the JobManager). I hope that makes sense for you.

rmetzger:
I added the changed heapspace setting for YARN to the pull request.
I think Spark has 384 MB that they subtract.

rmetzger:
While testing the YARN change on amazon EMR, I found that the YARN client is not compatible with Hadoop 2.4.0

```
Exception in thread "main" java.lang.NoSuchMethodError: org.apache.hadoop.yarn.util.Apps.addToEnvironment(Ljava/util/Map;Ljava/lang/String;Ljava/lang/String;)V
    at eu.stratosphere.yarn.Utils.setupEnv(Utils.java:160)
    at eu.stratosphere.yarn.ApplicationMaster.run(ApplicationMaster.java:253)
    at eu.stratosphere.yarn.ApplicationMaster.access$000(ApplicationMaster.java:62)
    at eu.stratosphere.yarn.ApplicationMaster$1.run(ApplicationMaster.java:314)
    at java.security.AccessController.doPrivileged(Native Method)
    at javax.security.auth.Subject.doAs(Subject.java:356)
    at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1528)
    at eu.stratosphere.yarn.ApplicationMaster.main(ApplicationMaster.java:310)
```

But I think the memory calculation is working properly

StephanEwen:
Should we cap it at 400 MB then?

rmetzger:
Merged in commit b31ae3ac3533441d2aa687ef11d3e637a32846ce.
I've left the 500 MB. I'll do some tests with 400 to see if its also working well.

Issue #795, opened 2014-05-12
Possibly extend the cost model of the optimizer


markus-h:
I have started the task to integrate the AbstractCachedBuildSideMatchDriver into the optimizer. The driver caches one side of the join and thereby can accellerate iterations if there are joins with static (non-changing) datasets inside the iteration.

The current way of calculating the cost of operators inside of iterations is basically to multiply them by the number of iterations. I would like to propose to extend this to have one static part of costs, that is counted only once for all iterations, and one dynamic part that is multiplied by the number of iterations.
In my opinion that would be the cleanest way to intergrate the cached match, by assigning it a higher starting cost then the regular match and a cheaper dynamic part.

One other approach would be to always use the cached match inside of iterations. For that I would probably have to add a new RequestedLocalProperty that tells the optimizer if the operator is used inside of a iteration.
A simple hacked solution could also be to simply exchange all suitable regular matches inside of an iteration by the cached alternative.

What do you think is the best approach?

markus-h:
I had a deeper look at this issue and now I think adjusting the cost model is the wrong way to go.
I will proceed with exchanging regular joins with cached ones inside of iterations when applicable.

StephanEwen:
For the cost calculation, you need to incorporate whether the build side is
on the dynamic or static path.

markus-h:
Ok, so if I do understand it correctly I exchange the driver in the NepheleJobGraphGenerator if one side of the match inside of an iteration is on a static path.

PR #822, opened 2014-05-15
Integration of the CachedBuildSideMatchDriver

markus-h:
This is a first version of integrating the CachedBuildSideMatchDriver into the Compiler.
In the current implementation it switches a regular hybrid hash join inside of iterations to the cached variant. The switching is done after the optimization in the NepheleJobGraphGenerator.

I debugged myself through a few of the example iterative algorithms we have and in many cases this implementation correctly switches the driver. But there are some cases where the optimizer decides to use a merge join and that can not so easily be replaced.

All in all I don't like this implementation very much. It is quite hacky and I think we should integrate the cached variant at an earlier point inside the optimization. But that would probably require a few extensions to the cost model. I created an issue to discuss this further:  #795


markus-h:
I am working on a cleaner integration, so please don't merge this yet.

markus-h:
I have decided for a different way to integrate this. So this pull request was replace by #836

PR #825, opened 2014-05-16
Fix windows line endings#791 - Windows delimiter is accepted as standard 

tobwiens:
This is a fix for issue: #791 

If the standard delimiter is \n, a \r\n is accepted as well. Therefore a
marker was introduced to indicate that specific case. The delimiter
length is variably set to the windows delimiter size then. To prevent
failing in reading input with windows line endings.
The DelimitedInputFormat.open() method did not check if it has parsers
registered. Added throwing an exception when no parser is registered
when calling open. CsvInputFormat does now accept windows line endings
with standards settings. The |\r\n" will be replaced by "\n_"( _ is a
space)... To prevent from errors the returned length is set to the
space. So that this byte is known as parsed and will not get interpreted
anymore.

### DIFF
--- a/stratosphere-java/src/main/java/eu/stratosphere/api/java/io/CsvInputFormat.java
+++ b/stratosphere-java/src/main/java/eu/stratosphere/api/java/io/CsvInputFormat.java
@@ -19,6 +19,7 @@ import java.io.IOException;
 import java.util.Map;
 import java.util.TreeMap;
 
+import com.google.common.base.Charsets;
 import com.google.common.base.Preconditions;
 
 import eu.stratosphere.api.common.io.GenericCsvInputFormat;
@@ -39,6 +40,9 @@ public class CsvInputFormat<OUT extends Tuple> extends GenericCsvInputFormat<OUT
 
 
 	private transient Object[] parsedValues;
+	
+	//to speed up readRecord processing. Used to find windows line endings
+	private boolean lineDelimiterIsLinebreak = false;
 	
 	
 	public CsvInputFormat(Path filePath) {
@@ -103,6 +107,12 @@ public class CsvInputFormat<OUT extends Tuple> extends GenericCsvInputFormat<OUT
 		for (int i = 0; i < fieldParsers.length; i++) {
 			this.parsedValues[i] = fieldParsers[i].createValue();
 		}
+		
+		//left to right evaluation makes access [0] okay
+		//this marker is used to fasten up readRecord, so that it doesn't have to check each call if the line ending is set to default
+		if(this.getDelimiter().length == 1 && this.getDelimiter()[0] == "\n".getBytes(Charsets.UTF_8)[0] ) {
+			this.lineDelimiterIsLinebreak = true;
+		}
 	}
 
 	@Override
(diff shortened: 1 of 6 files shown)


tobwiens:
Is not stable yet.

stratosphere-java/src/main/java/eu/stratosphere/api/java/io/CsvInputFormat.java
```diff
@@ -93,26 +98,50 @@ public void open(FileInputSplit split) throws IOException {
...
+		
 		// create the value holders
 		this.parsedValues = new Object[fieldParsers.length];
 		for (int i = 0; i < fieldParsers.length; i++) {
 			this.parsedValues[i] = fieldParsers[i].createValue();
 		}
+		
+		//left to right evaluation makes access [0] okay
+		//this marker is used to fasten up readRecord, so that it doesn't have to check each call if the line ending is set to default
+		if(this.getDelimiter().length == 1 && this.getDelimiter()[0] == "\n".getBytes(Charsets.UTF_8)[0] ) {
```
StephanEwen: I think you can simplify this to `getDelimiter()[0] == '\n'`

I think that line feed is the same in all charsets that matter. We do exclude mainframe and C64 charsets that way, but I am okay with that ;-)

stratosphere-java/src/main/java/eu/stratosphere/api/java/io/CsvInputFormat.java
```diff
@@ -93,26 +98,50 @@ public void open(FileInputSplit split) throws IOException {
...
+		}
 	}
 
 	@Override
 	public OUT readRecord(OUT reuse, byte[] bytes, int offset, int numBytes) {
+		/*
+		 * Fix to support windows line endings in CSVInputFiles with standard delimiter setup = \n
+		 */
+		//Find windows end line, so find chariage return before the newline 
+		if(this.lineDelimiterIsLinebreak == true && bytes[offset + numBytes -1] == "\r".getBytes(Charsets.UTF_8)[0]) {
```
StephanEwen: `"\r".getBytes(Charsets.UTF_8)[0]` is a really expensive call. It goes through the charset codecs.
Just do a `bytes[offset + numBytes -1] == '\r'`.

tobwiens:
Thank you Stefan, I addressed the issues!

I left the length == 1 check, to not interfere with arbitrary delimiters which might contain a \n at position 0.

stratosphere-java/src/main/java/eu/stratosphere/api/java/io/CsvInputFormat.java
```diff
@@ -93,26 +97,50 @@ public void open(FileInputSplit split) throws IOException {
...
 
 	@Override
 	public OUT readRecord(OUT reuse, byte[] bytes, int offset, int numBytes) {
+		/*
+		 * Fix to support windows line endings in CSVInputFiles with standard delimiter setup = \n
+		 */
+		//Find windows end line, so find chariage return before the newline 
+		if(this.lineDelimiterIsLinebreak == true && bytes[offset + numBytes -1] == '\r' ) {
+			//reduce the number of bytes so that the Carriage return is not taken as data
+			numBytes--;
```
StephanEwen, line 126: I think it would be good to add a check if that reduces teh line to length 0. Skip the line in that case.

StephanEwen:
Fixed up and merged in b0557d08865d31762fb2ddb913b4aa7a2f3ed3b3

PR #836, opened 2014-05-19
Integration of the CachedBuildSideMatchDriver into the optimizer

markus-h:
I have redone the integration of the cached match driver. It is now integrated into the optimizer instead of just switching the DriverStrategy in the NepheleJobGraphGenerator. I feel it is much cleaner this way.
So this pull request replaces #822. 
Instead of extending the cost model like proposed in #795 the cost estimation function get passed the number of iterations, so that you can decide for every operator on its own how to incorporate iterations.

### DIFF
--- a/stratosphere-compiler/src/main/java/eu/stratosphere/compiler/PactCompiler.java
+++ b/stratosphere-compiler/src/main/java/eu/stratosphere/compiler/PactCompiler.java
@@ -792,8 +792,6 @@ public class PactCompiler {
 		private final GraphCreatingVisitor parent;	// reference to enclosing creator, in case of a recursive translation
 		
 		private final boolean forceDOP;
-		
-		private boolean insideIteration = false; // used to mark inner nodes of iterations
 
 		
 		private GraphCreatingVisitor(int maxMachines, int defaultParallelism) {
@@ -813,12 +811,6 @@ public class PactCompiler {
 			this.defaultParallelism = defaultParallelism;
 			this.parent = parent;
 			this.forceDOP = forceDOP;
-		}
-		
-		private GraphCreatingVisitor(GraphCreatingVisitor parent, boolean forceDOP, int maxMachines,
-				int defaultParallelism, HashMap<Operator, OptimizerNode> closure, boolean insideIteration) {
-			this(parent, forceDOP, maxMachines, defaultParallelism, closure);
-			this.insideIteration = insideIteration;
 		}
 
 		@Override
@@ -952,9 +944,6 @@ public class PactCompiler {
 				// value required to obey the maximum number of machines
 				n.setSubtasksPerInstance(tasksPerInstance);
 			}
-			
-			n.setInsideIteration(insideIteration);
-			
 			return true;
 		}
 
@@ -977,7 +966,7 @@ public class PactCompiler {
 
 				// first, recursively build the data flow for the step function
 				final GraphCreatingVisitor recursiveCreator = new GraphCreatingVisitor(this, true,
-					this.maxMachines, iterNode.getDegreeOfParallelism(), closure, true);
+					this.maxMachines, iterNode.getDegreeOfParallelism(), closure);
 				
 				BulkPartialSolutionNode partialSolution = null;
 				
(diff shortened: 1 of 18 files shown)


stratosphere-compiler/src/main/java/eu/stratosphere/compiler/dag/TwoInputNode.java
```diff
@@ -521,7 +521,7 @@ protected void addLocalCandidates(Channel template1, Channel template2, List<Set
 							{
 								// all right, co compatible
 								instantiate(dps, in1, in2, broadcastPlanChannels, target, estimator, rgps1, rgps2, ilp1, ilp2);
-								break allPossibleLoop;
+								continue allPossibleLoop;
```
markus-h, line 524: I did not understand why this was a break before. I think with a break many senseful plans get skipped.

StephanEwen, line 524: Can you verify that? Maybe with a test ;-)

StephanEwen:
This pull request has not a single test. We cannot merge it like this.
Please make a list of the tests you want to add and how they cover the code.

stratosphere-compiler/src/main/java/eu/stratosphere/compiler/dag/TwoInputNode.java
markus-h, line 524: I can try, but it will cost me some time. I discovered this issue when I was debugging myself through some complex plans, but I am not sure how to construct an artificial situation.

markus-h:
I was thinking about tests before I created the pull request but in my opinion it is quite hard to write good compiler tests (perhaps that is also the reason why we have such few).
I could verify that an (by hint) enforced hash join inside of an iteration is correctly replaced, but that would only cover a very small part of the optimizer and even if there is a fault in my code it will most probably not discover it.
An other possibility I was thinking about is constructing a very specific example plan (including manually created statistics), but such an test would only work as long as the cost calculation of all operators is not changed and no new operators are added.

markus-h:
I added testcases for my changes.
I also found a way to address the break/continue issue. The testcase "testCorrectChoosing" fails when there is a break. Because both hash left and hash right always are compatible the break causes that the right variant is never instantiated.

stratosphere-compiler/src/main/java/eu/stratosphere/compiler/PactCompiler.java
```diff
@@ -792,6 +792,8 @@ else if (defaultParallelism > maxMachinesJob * this.maxIntraNodeParallelism) {
 		private final GraphCreatingVisitor parent;	// reference to enclosing creator, in case of a recursive translation
 		
 		private final boolean forceDOP;
+		
+		private boolean insideIteration = false; // used to mark inner nodes of iterations
```
StephanEwen: Do we need those flags?

Would it be sufficient to identify situations for the cached hash tables by finding joins where on part is on the static path, one part is on the dynamic path?

StephanEwen:
I think before we can merge this, we need to also change the `BuildFirstCachedMatchDriver` and `BuildSecondCachedMatchDriver` to use the `ReOpenableMutableHashTable`.

Otherwise we loose out-of-core functionality in iterations.

stratosphere-compiler/src/main/java/eu/stratosphere/compiler/PactCompiler.java
markus-h: Yes your are right, we don't need them. I didn't realize that dynamic paths are only identified inside iterations

markus-h:
I changed the MutableHashTable to the ReOpenableMutableHashTable. I am not very familiar with this level of the runtime, is there anything else I have to do?

StephanEwen:
The life cycle has to be properly mapped. Initialization of the hash table,
resetting the hash table after each superstep, tearing it down at the very
end.

I would suggest to write a small dedicated test for the driver that uses a
hash table with minimal memory (30 buffers or so) and goes through three
iterations.

markus-h:
I did a rework of the AbstractCachedBuildSideMatchDriver to properly use the ReOpenableMutableHashTable in combination with the BuildFirstReOpenableHashMatchIterator. I also added testcases similiar to the test of the regular match driver.

StephanEwen:
Merged in https://github.com/apache/incubator-flink 47a68adc7fa5da7743cb51fe69f81264e908e630

PR #850, opened 2014-05-22
Add rCF() call to ContextEnvironment

zentol:
Addendum to #728.


uce:
LGTM.

Your branch name is `dist-cache-fix`. What was the problem for the distributed cache here?

zentol:
if you use the distributed cache via `ExecutionEnvironment.registerCachedFile()` it straight up doesn't work in an actual setting. When you submit a plan to stratosphere it seems to use `ContextEnvironment`, which until now never actually registered the files.

(Note that the version that #728 was supposed to replace does work)

```
l.executePlan(p);
```

rmetzger:
So this is a bugfix that we should also add to 0.5.1 right?

zentol:
yup

StephanEwen:
The fix has been subsumed by abe88b7972c24b85338033042b07cba64ca91054

The above commit registers files in the `createProgramPlan()` method, which is the common point shared by all environments, etc.

PR #856, opened 2014-05-24
Improve exception for nonexistent local files

rmetzger:


stratosphere-core/src/main/java/eu/stratosphere/core/fs/local/LocalFileSystem.java
```diff
@@ -97,7 +97,8 @@ public FileStatus getFileStatus(Path f) throws IOException {
 		if (path.exists()) {
 			return new LocalFileStatus(pathToFile(f), this);
 		} else {
-			throw new FileNotFoundException("File " + f + " does not exist.");
+			throw new FileNotFoundException("File " + f + " does not exist or the user running "
+					+ "Stratosphere ('"+System.getProperty("user.name")+"') has insufficient permissions to access it.");
```
uce, line 101: LGTM.

Initially I thought to scrap "user running Stratosphere" from the message as we need to replace all Stratosphere appearances soon ;-), but it's more precisie than just saying "user".

rmetzger:
Merged in 7ad6c8baee710254608c6f4a929cdcc0ba34c7d1.

Issue #857, opened 2014-05-24
Exception while executing wordcount example


rmetzger:
Hi,
while testing the current `master` branch to see if we can release it this way, I found an issue with the new Java API / Wordcount example.

```
05/24/2014 21:56:32:    Job execution switched to status FAILED
Error: The program execution failed: java.lang.ArrayIndexOutOfBoundsException: -128
    at java.util.ArrayList.elementData(ArrayList.java:400)
    at java.util.ArrayList.get(ArrayList.java:413)
    at eu.stratosphere.nephele.io.RuntimeOutputGate.writeRecord(RuntimeOutputGate.java:268)
    at eu.stratosphere.nephele.io.AbstractRecordWriter.emit(AbstractRecordWriter.java:92)
    at eu.stratosphere.pact.runtime.shipping.OutputCollector.collect(OutputCollector.java:82)
    at eu.stratosphere.example.java.wordcount.WordCount$Tokenizer.flatMap(WordCount.java:101)
    at eu.stratosphere.example.java.wordcount.WordCount$Tokenizer.flatMap(WordCount.java:91)
    at eu.stratosphere.pact.runtime.task.chaining.ChainedFlatMapDriver.collect(ChainedFlatMapDriver.java:71)
    at eu.stratosphere.pact.runtime.task.DataSourceTask.invoke(DataSourceTask.java:219)
    at eu.stratosphere.nephele.execution.RuntimeEnvironment.run(RuntimeEnvironment.java:351)
    at java.lang.Thread.run(Thread.java:745)

For a more detailed error message use the vebose output option '-v'.
```

I successfully executed the Scala and Record Wordcount implementations with the same configuration (cluster size, data set).
The error is reproducible.

StephanEwen:
Bug in the hash partitioner.

rmetzger:
The bug is caused by a the following problem: 
`Integer.MIN_VALUE == -Integer.MIN_VALUE`. Also, `Math.abs(Integer.MIN_VALUE) = Integer.MIN_VALUE`.
See also here: http://stackoverflow.com/questions/18565485/why-is-absolute-of-integer-min-value-equivalent-to-integer-min-value

The error only occurs if the hash value is accidentally `Integer.MIN_VALUE`.

The RecordOutputEmitter is doing some hash-salting magic:

``` java
for (int i = 0; i < DEFAULT_SALT.length; i++) {
[...]

PR #859, opened 2014-05-25
Fix hash partitioning for new java api

rmetzger:
This is a fix for this issue https://github.com/stratosphere/stratosphere/issues/857

### DIFF
--- a/stratosphere-runtime/src/main/java/eu/stratosphere/pact/runtime/shipping/OutputEmitter.java
+++ b/stratosphere-runtime/src/main/java/eu/stratosphere/pact/runtime/shipping/OutputEmitter.java
@@ -148,7 +148,15 @@ public class OutputEmitter<T> implements ChannelSelector<SerializationDelegate<T
 		
 		int hash = this.comparator.hash(record);
 	
-		this.channels[0] = (hash < 0) ? -hash % numberOfChannels : hash % numberOfChannels;
+		if(hash < 0) {
+			if(hash == Integer.MIN_VALUE) {
+				this.channels[0] = Integer.MAX_VALUE % numberOfChannels;
+			} else {
+				this.channels[0] = -hash % numberOfChannels;
+			}
+		} else {
+			this.channels[0] = hash % numberOfChannels;
+		}
 		return this.channels;
 	}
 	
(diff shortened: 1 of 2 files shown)


stratosphere-runtime/src/main/java/eu/stratosphere/pact/runtime/shipping/OutputEmitter.java
```diff
@@ -148,7 +148,15 @@ public OutputEmitter(ShipStrategyType strategy, TypeComparator<T> comparator, Da
 		
 		int hash = this.comparator.hash(record);
 	
-		this.channels[0] = (hash < 0) ? -hash % numberOfChannels : hash % numberOfChannels;
+		if(hash < 0) {
+			if(hash == Integer.MIN_VALUE) {
```
uce, line 152: LGTM.

PS: I vote to enforce `if (` instead of `if(`... let the flame war begin :laughing:

uce:
Can you infer the value which gets hashed to Integer.MIN_VALUE and add a test to `OutputEmitterTest`?

uce:
And the problem should also apply to `RecordOutputEmitter`, no?

StephanEwen:
Looks, good. Please add the same thing to the RecordOutputEmitter and merge soon.

I order to test it: Right now, the OutputEmitter uses the hash value returned by the TypeComparator. So you can use a custom type comparator that returns Integer.MIN_VALUE for the test.

uce:
I wrote before thinking... even better if we can just use a custom Comparator.

@rmetzger Please introduce mockito as a dependency; I will be doing that with an upcoming PR anyways.

``` xml
<dependency>
    <groupId>org.mockito</groupId>
    <artifactId>mockito-all</artifactId>
    <version>1.9.5</version>
    <scope>test</scope>
</dependency>
```

Then you can mock the TypeComparator easily:

``` java
TypeComparator<Record> comparatorMock = mock(TypeComparator.class);
when(comparatorMock.hash(Matchers.<Record>anyObject()))
    .thenReturn(Integer.MIN_VALUE);

int hash = comparatorMock.hash(new Record(new IntValue(1))); // Integer.MIN_VALUE
```

fhueske:
@rmetzger should I continue to fix this or do you want to do it?

rmetzger:
Yeah, I will not have much time in the next two days to finish the fix.
Ufuk also offered to finish the PR. So you should synchronize.

Sent from my iPhone

On 26.05.2014, at 10:50, Fabian Hueske notifications@github.com wrote:

@rmetzger https://github.com/rmetzger should I continue to fix this or do
you want to do it?

—
GitHubhttps://github.com/stratosphere/stratosphere/pull/859#issuecomment-44169178
.

fhueske:
@uce I'm done with this, tests are currently running.

rmetzger:
Cool. I think I'm pushing the next RC later today.

Sent from my iPhone

On 26.05.2014, at 11:27, Fabian Hueske notifications@github.com wrote:

@uce https://github.com/uce I'm done with this, tests are currently
running.

—
GitHubhttps://github.com/stratosphere/stratosphere/pull/859#issuecomment-44172413
.

fhueske:
:+1:

uce:
@fhueske Nice. Let's close this PR and merge your fix with a new PR later.

fhueske:
Merged

Issue #760, opened 2014-05-06
Introduce a DataSet.distinct() operator


twalthr:
I think it would be more readable and user-friendly to have a DataSet.distinct() operator instead of manually use e.g. for Tuple5:

``` java
DataSet.groupBy(0,1,2,3,4).reduceGroup(new GroupReduceFunction<...>() {
    private static final long serialVersionUID = 1L;

    @Override
    public void reduce(Iterator<...> values, Collector<...> out) throws Exception {
        out.collect(values.next());                     
    }                   
});
```

markus-h:
Good idea, I think it would be nice to have a distinct operator

PR #872, opened 2014-05-27
Distinct Operator Implementation

markus-h:
See #760

### DIFF
--- a/stratosphere-java/src/main/java/eu/stratosphere/api/java/DataSet.java
+++ b/stratosphere-java/src/main/java/eu/stratosphere/api/java/DataSet.java
@@ -36,6 +36,7 @@ import eu.stratosphere.api.java.operators.CrossOperator;
 import eu.stratosphere.api.java.operators.CrossOperator.DefaultCross;
 import eu.stratosphere.api.java.operators.CustomUnaryOperation;
 import eu.stratosphere.api.java.operators.DataSink;
+import eu.stratosphere.api.java.operators.DistinctOperator;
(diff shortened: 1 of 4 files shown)


fhueske:
I know the distinct method headers have been there before, but I'm not sure if it makes sense to select keys for a distinct transformation. IMO, distinct should work on all fields of a tuple (or the full object to be more general).

StephanEwen:
I think distinct with a subset of fields makes sense. Same thing is also
supported in SQL.

I would like this pull request to have some better unit tests. The current
ones simply try to see whether an exception is thrown.

StephanEwen:
Have a look at the reduce translation tests and the reduce compiler tests.
They give you an idea how that works.

Also, duplicate elimination should be "combinable".

fhueske:
I think we should also provide a `distinct` method without arguments that uses the full tuple or object for the distinct check. In that case we must make sure that all objects are comparable.

markus-h:
Thanks for your feedback! I already included a plain distinct() method, that is currently only working for tuples and uses the full tuple.

fhueske:
Cool, thanks! Had not noticed that one :smile:

markus-h:
I added translation and compilation tests. Since the distinct operator currently is directly mapped to a groupReduce I almost had to copy the test from reduce.

rmetzger:
Is this pull request ready to merge?

uce:
Merged in e2aabd90604d5880ef8e7b82ce58de218ddd48a7.

Further comments I have are in [FLINK-984](https://issues.apache.org/jira/browse/FLINK-984).

PR #873, opened 2014-05-27
Aligned the Scala KMeans with the Java KMeans example.

aalexandrov:
Some fixes in the Scala KMeans example to unify the Scala with the Java example so both behave the same. This should avoid unnecessary confusion when people try the [Quick Start Guide](http://stratosphere.eu/quickstart/example.html) with the Scala version instead.

PS. I'm not sure whether the Scala KMeans job is currenly used for integration testing. If yes, the IT test cases should be adapted before merge as well.

### DIFF
--- a/stratosphere-examples/stratosphere-scala-examples/src/main/scala/eu/stratosphere/examples/scala/datamining/KMeans.scala
+++ b/stratosphere-examples/stratosphere-scala-examples/src/main/scala/eu/stratosphere/examples/scala/datamining/KMeans.scala
@@ -25,55 +25,65 @@ import eu.stratosphere.api.scala.operators._
 
 class KMeans extends Program with ProgramDescription with Serializable {
 
-  case class Point(x: Double, y: Double, z: Double) {
-    
-    def +(other: Point) = { Point(x + other.x, y + other.y, z + other.z) }
-    
-    def /(div: Int) = { Point(x / div, y / div, z / div) }
-    
-    def computeEuclidianDistance(other: Point) = {
-      math.sqrt(math.pow(x - other.x, 2) + math.pow(y - other.y, 2) + math.pow(z - other.z, 2))
+  case class Point(x: Double, y: Double) {
+
+    def +(other: Point) = { Point(x + other.x, y + other.y) }
+
+    def /(div: Int) = { Point(x / div, y / div) }
+
+    def euclidianDistance(other: Point) = {
+      math.sqrt(math.pow(x - other.x, 2) + math.pow(y - other.y, 2))
     }
   }
 
-  case class Distance(dataPoint: Point, clusterId: Int, distance: Double)
-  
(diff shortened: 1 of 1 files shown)


aalexandrov:
OK, the tests fail because the `IterativeKMeansITCase` still calls `getScalaPlan(dop, ...)` and I removed the DOP from the example in order to align it with the Java job.

It seems that quite a lot different KMeans with heterogeneous input/output formats co-exist at the moment. We have:
1. three variants of using the old Java API in `stratosphere-tests` which work 
2. one variant (which I adapted) with the Scala API in `stratosphere-scala-examples`
3. one variant with the new Java API in `stratosphere-java-examples`

Integration tests exist against 1. and (the old version of) 2. Both rely on the DataSets specified in `eu.stratosphere.test.testdata.KMeansData`. 

On the other hand, 3. does not seem to have an IT case, and relies on a different input format which can be seen in `eu.stratosphere.example.java.clustering.util.KMeansData`.

After the changes made in this PR, 2. would also require input / output data like the one specified in `eu.stratosphere.example.java.clustering.util.KMeansData`. 

We need some strategy to align everything. I suggest to keep code entropy low in order to reduce the maintenance effort in the future.

fhueske:
A big :+1: for keeping things in sync.
IMO we should:
- remove the old Java API examples incl. Integration tests (but check if we will go down in test coverage)
- add an integration test for the new Java API KMeans example
- port the Scala KMeans test to use the same data as the new Java KMeans test.

We have probably similar situations with some of the other examples.

aalexandrov:
I suggest to take the following steps:
- Adapt the 0.4 Java API examples and their ITcases in (1) to use the data provided by `eu.stratosphere.example.java.clustering.util.KMeansData`. Mark those unit-tests as deprecated and remove at some point in the future.
- Add an ITcase for the 0.5 Java API example job (2).
- Adapt the current ITCase for the Scala KMeans (3) to reflect the modifications in the example job.
- Remove the obsolete `eu.stratosphere.test.testdata.KMeansData` file.

Does everybody agree with this strategy?

fhueske:
Not sure if the Java 0.4 KMeans ITCases add any benefit. So porting them might be not have a gain.
I would opt to simply drop them (esp. since there are two other ITCases running this job).

Apart from that, I fully agree.

aalexandrov:
Anybody against Fabian's suggestion?

StephanEwen:
I am all for it.

Keeping test coverage but reducing dependencies on the old API by migrating tests is the way to go.

StephanEwen:
We are moving active pull requests to the apache mirror and closing inactive ones.

@aalexandrov Please let us know whether you will implement this suggestion.

uce:
Assigned @aalexandrov to corresponding [FLINK-873](https://issues.apache.org/jira/browse/FLINK-873). He agreed to work on it during the week. :-)

PR #874, opened 2014-05-27
fixes toString return value of min aggregation

vasia:
This is very minor but was enough to confuse me while trying to make sense of my logs today, so I fixed it :)


uce:
Indeed, this can be confusing. Good catch :) I'll merge it now.

Since this is a minor String change, you can add `[ci skip]` to the commit message in order to not have Travis build and test your commit (http://docs.travis-ci.com/user/how-to-skip-a-build/).

uce:
Sorry @rmetzger, @StephanEwen, I regret not merging it manually (because of the extra merge commit). Won't happen again. :sob:

vasia:
Great, thnx for the tip @uce!

PR #882, opened 2014-05-28
DistCache: executable flag, directories

zentol:
new features:
- you can now distribute whole directories (this copies all files tracked by `listStatus` recursively)
- when distributing single files they can be tagged as executable (automatically checked for local files)

other changes:
- the name/path scheme was changed to preserve the original name. all files are now stored in a job-specific folder, instead of having a job/identifier specific name.
- the arguments passed when registering a file are abstracted away in its own data structure, making it easier to expand the DC the future. (there is no reason why one should have to change `NepheleJobGraphGenerator` or `TaskManager` for this)
- reworded/expanded some javadocs, and made it clear that local files can only be used when they are accessible form all workers

sidenotes:
- `FileCache.CopyProcess.create()` is basically a copy function. i think we could add this as a static method to `FileSystem` as `copy(URI sourcePath, URI targetPath)`.
- the code will be tested with the python interface


rmetzger:
Cool! Can you add some more tests for the DistributedCache ? 
Especially for bugs you've encountered with the python interface.

zentol:
sure.

zentol:
verified on the cluster.

StephanEwen:
I think this is good to merge.

Will merge in the next batch

PR #884, opened 2014-05-29
Add missing Usercode Classloader for the "info" command in the CliFrontend

rmetzger:
Rework stratosphere-clients tests to not expect a binary blob in the test resources (build the jar file using maven)

add a test for ensuring the presense of a custom classloader in the CliFrontend
Add missing ClassLoader for the "info" command in the CliFrontend


rmetzger:
Where else in the project are we shipping jar files for testing purposes? If you want, I can also add the mechanism to generate this jar file there.

stratosphere-clients/src/main/java/eu/stratosphere/client/program/PackagedProgram.java
```diff
@@ -312,6 +313,12 @@ public void invokeInteractiveModeForExecution() throws ProgramInvocationExceptio
 	public ClassLoader getUserCodeClassLoader() {
 		return this.userCodeClassLoader;
 	}
+
+	public void setUserCodeClassLoader(ClassLoader cl) {
```
StephanEwen, line 317: Do we need this? It may encourage to override the user code class loader.

rmetzger, line 317: I've added this for testing purposes only.
Do you think its better to use Mockito or reflection to change the visibility of the method for testing? (Or is there even a better approach for that?)

rmetzger:
I'm currently preparing an update for this pull request to remove all ".jar" files in our project.

StephanEwen:
Very good!

fhueske:
:+1: !

rmetzger:
The removal of the jar is done. The only blocker for this pull request is the question regarding the public `setUserCodeClassLoader()` method.

uce:
I removed the setter and used mockito to create a partial mock in the test case, which returns the test class loader: https://github.com/uce/stratosphere/tree/rmetzger-infoFixPr.

I'm merging it if Travis gives the OK: https://travis-ci.org/uce/stratosphere/builds/26847638.

uce:
Merged in d80b929f57e3a52a4ce688acd19159236a697c7e.

PR #891, opened 2014-06-01
Potential bug due to unnecessary cast from long to int

rmetzger:
The `FSDataInputStreamWrapper`  accepts an int, but the length is a long (and all methods passing the length are passing long.


uce:
Merged in 1ce11d3a38f0e310bb8287a51208ac2e42b13d63.

Issue #612, opened 2014-03-19
Add Java 8 Lambda support to new Java API


rmetzger:
Due to the limitations with Java 8 lambdas, we probably have to introduce another argument next to the lambda to pass the return type.

carabolic:
I'm looking into it. It seems to be that type erasure is the biggest problem. The [TypeExtractor createTypeInfo(Type)](https://github.com/stratosphere/stratosphere/blob/master/stratosphere-java/src/main/java/eu/stratosphere/api/java/typeutils/TypeExtractor.java#L79) returns `null` instead of the instance type. There might be a solution using [SerializedLambda](http://download.java.net/jdk8/docs/api/java/lang/invoke/SerializedLambda.html).

StephanEwen:
The abstract base classes implemented lambdas are all serializable. I
assume that helps?

carabolic:
Alright I made tests and it turns out that we don't need `SerializedLambda` as it doesn't seem to help at all. But by using reflection I was able to get the parameter types of the lambda.

Consider the following example:

``` Java
@FunctionalInterface
interface LambdaFlatMap<IN, OUT> {
  void flatMapL(IN value, Collector<OUT> out) throws Exception;
}
```

``` Java
/**
   * Custom FlatMapOperator that supports Lambdas.
   * @param <IN>
   * @param <OUT>
   */
public class FlatMapOperator<IN, OUT> {
  public FlatMapOperator(DataSet<IN> input, LambdaFlatMap<IN, OUT> function) {
    // do something
  }
}
```

``` Java
public class TypeInference {

  public static void main(String... args) throws IOException {
    final ExecutionEnvironment env = ExecutionEnvironment.getExecutionEnvironment();

    DataSet<String> input = env.fromElements("This", "is", "a", "test", "test");
    FlatMapOperator<String, String> flatMapOperator =
            new FlatMapOperator<>(input, (value, out) -> out.collect(value.toLowerCase()));
  }
}
```

If we want to extract the actual types for `IN` and `OUT` of the `LambdaFlatMap` (and any other lambda function) interface we could do the following:

``` Java
    Arrays.stream(TypeInference.class.getDeclaredMethods())
            // get all methods that are actually lambdas
            .filter(m -> m.getName().contains("lambda$"))
            // print all parameter types
            .forEach(m -> Arrays.stream(m.getParameterTypes()).forEach(System.out::println));
```

As a result this procedure would print:

```
class java.lang.String
[...]

PR #895, opened 2014-06-02
Exemplary implementation of map function with Java 8 Lambdas.

twalthr:
This PR shows how we could implement Java 8 Lambdas into the current master without changing much code. I implemented a Lambda Mapper. An example job shows that it is working.

See also #612.

### DIFF
--- a/stratosphere-core/src/main/java/eu/stratosphere/api/common/functions/Function.java
+++ b/stratosphere-core/src/main/java/eu/stratosphere/api/common/functions/Function.java
@@ -56,7 +56,9 @@ public interface Function {
 	 * 
 	 * @see eu.stratosphere.configuration.Configuration
 	 */
-	void open(Configuration parameters) throws Exception;
+	default void open(Configuration parameters) throws Exception {
+		
+	}
 
 	/**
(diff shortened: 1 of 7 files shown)


aljoscha:
Very nice. It might not be pretty, but I think it's the only way to get the lambda function types. I don't like the Wrapper, I see that it was the easiest way to do it here but for a release we can easily get rid of that.

Also, this will fail right now when you pass in an object (anonymous inner class) that directly derives from `GenericMap`, here we cannot use the lambda resolution logic but the old logic with `getGenericInterfaces()`.

twalthr:
Yes, you are right. I have prepared the new `getMapReturnTypes()` method. We then need to rework `getParameterType()` as it does not support interfaces as "base class". After the rework is done we can get rid of `getMapReturnTypes(MapFunction<IN, OUT> mapFunction, TypeInformation<IN> inType)`.

uce:
I'm closing this PR for now as I think it was just to gather some feedback. I've linked this PR in [FLINK-701](https://issues.apache.org/jira/browse/FLINK-701).

PR #896, opened 2014-06-02
Fixes bugs where the TypeExtractor throws an NPE instead of the operators

twalthr:
If the user accidentally passes `null` instead of e.g. a MapFunction. The TypeExtractor throws the NPE instead of the higher layer.


StephanEwen:
Looks good to merge

StephanEwen:
Will merge now...

StephanEwen:
Merged in 2b0baea9b8a6dd99052c2dfa98cae719a39d6bbc

Issue #766, opened 2014-05-07
Web interface progress monitoring for DataSources and DataSinks with splits


rmetzger:
The progress monitoring for DataSources and DataSinks can be improved by including the number of processed vs total splits into the progress.

uce:
+1

StephanEwen:
+1

fhueske:
:+1:

rmetzger:
I think @tobwiens is working on this. Are there any updates?

tobwiens:
Yes, since I am working on the interface I will fix both issues. 

Current progress: The input splits of the data source are correctly displayed. After the page is displayed for the first time, an update routine is shifting the progress bars. The update routine works completely different  and makes it a struggle.

PR #899, opened 2014-06-02
Fix data sources and data sinks with splits #766

tobwiens:
Fix for issue #766. 

DataSource displays processed input splits. DataSink is not displaying output, though.

### DIFF
--- a/stratosphere-runtime/resources/web-docs-infoserver/js/jobmanagerFrontend.js
+++ b/stratosphere-runtime/resources/web-docs-infoserver/js/jobmanagerFrontend.js
@@ -149,7 +149,7 @@ function updateSourceSink(json) {
 	$.each(json, function(i, job) {
 		$.each(job.groupvertices, function(j, groupvertex) {
 		if(groupvertex.processedsplits > 0) {
-			$("#"+groupvertex.groupvertexid).parent().contents().find(".progress-bar-inner-blue").css("width", (groupvertex.processedsplits/groupvertex.numberofgroupmembers)*100+"%")
+			$("#"+groupvertex.groupvertexid).parent().contents().find(".progress-bar-inner-blue").css("width", (groupvertex.processedsplits/groupvertex.numberofgroupmembers)*100+"%");
 			}
 		})
 	})
--- a/stratosphere-runtime/resources/web-docs-infoserver/js/jobmanagerFrontend.js
+++ b/stratosphere-runtime/resources/web-docs-infoserver/js/jobmanagerFrontend.js
@@ -149,7 +149,7 @@ function updateSourceSink(json) {
 	$.each(json, function(i, job) {
 		$.each(job.groupvertices, function(j, groupvertex) {
 		if(groupvertex.processedsplits > 0) {
-			$("#"+groupvertex.groupvertexid).parent().contents().find(".progress-bar-inner-blue").css("width", (groupvertex.numberofgroupmembers/groupvertex.processedsplits)*100+"%")
+			$("#"+groupvertex.groupvertexid).parent().contents().find(".progress-bar-inner-blue").css("width", (groupvertex.processedsplits/groupvertex.numberofgroupmembers)*100+"%")
 			}
 		})
 	})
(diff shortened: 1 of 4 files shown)


stratosphere-runtime/resources/web-docs-infoserver/js/jobmanagerFrontend.js
```diff
@@ -29,6 +29,15 @@ function init() {
 // Init once on page load
 $(init());
 
+function pollForDataSourceSink() {
+$.ajax({ url : "jobsInfo", type : "GET", cache: false, success : function(json) {
```
rmetzger: do you make dedicated requests just to get the DataSource progress?

stratosphere-runtime/resources/web-docs-infoserver/js/jobmanagerFrontend.js
```diff
@@ -267,6 +288,31 @@ function progressBar2(maximum, input) {
...
+ * Generates two progress bars in one
+ */
+function progressBarTwoInOne(firstMaximum, firstInput, secondMaximum, secondInput, classVal) {
+	if (firstInput != 0) {
+		return "<td class=\""+classVal+"\" val=\""+firstInput+"\"><div class=\"progress\"><div class=\"progress-bar progress-bar-success\" role=\"progressbar\""
+		 			+ "aria-valuemin=\"0\" aria-valuemax=\""+widthProgressbar+"\" style=\"width:"
+					+ (firstInput / firstMaximum)*100+"%;\">"
+					+ firstInput
+					+ "</div>" //close class=progress-bar progress-bar-success
+					//+ "</div>" //close class=progress
```
rmetzger, line 291: I think its not necessary to have this commented out code, right?

rmetzger:
Can you squash the commits together into one commit? (every pull request should roughly be one commit (a few more for really large PRs)
After you've squashed (git rebase -i HEAD~11) the commits together, do a force push to update the pull request

stratosphere-runtime/resources/web-docs-infoserver/js/jobmanagerFrontend.js
tobwiens: Exactly, the updates which are pulled from the other poll functions don't give me GroupVertex information.

stratosphere-runtime/resources/web-docs-infoserver/js/jobmanagerFrontend.js
tobwiens, line 291: Okay I see! I will remove it....

stratosphere-runtime/resources/web-docs-infoserver/js/jobmanagerFrontend.js
rmetzger: So we now have two periodic polling requests running?
Can't we merge them together to reduce the load on the JobManager?

stratosphere-runtime/src/main/java/eu/stratosphere/nephele/jobmanager/JobManager.java
```diff
@@ -1168,8 +1170,28 @@ public InputSplitWrapper requestNextInputSplit(final JobID jobID, final Executio
...
+		// Get the management vertex
+		ManagementVertex mgmtVertex = mgmtGraph.getVertexByID(vertex.getID().toManagementVertexID());
+		// Finally get the management group vertex
+		ManagementGroupVertex mgmtGroupVertex = mgmtVertex.getGroupVertex();
+				
+		// Increase the number of processed splits when input split returned is null
+		// Because then the previous input splits have been read
+		if(inputSplitWrapper.getInputSplit() == null){
+			mgmtGroupVertex.setProcessedSplits(mgmtGroupVertex.getProcessedSplits() + 1);
+		}
```
markus-h, line 1192: The ManagementGraph is administered by the EventCollector, so in my opinion it would be nicer to do the update of the input splits of a ManagementGroupVertex inside of the EventCollector. What do you think, would it be nicer to have a method similiar to the current updateManagementGraph(XY) methods for the update of the input splits? Perhaps with a new *Event type?

tobwiens, line 1192: I will have a look into that, if it makes the double polling unnecessary it will be a preferred solution. 
Do you mean that  the requestInputSplit triggers that event? Or does classes which request input splits trigger that event?

markus-h, line 1192: requestNextInputSplit() is called by RPC, right? So for now it should be good enough to trigger a event inside of requestNextInputSplit(), since its already running on the JobManager.

tobwiens:
I looked into the solution with the new event type. Since it would have been a lot of code for showing a small progress bar I went a different way. 
Through debugging if found: When a DataSouce finishes reading all its input splits it changes into the state FINISHING. It stays in that state until the data is handed over. 
Therefore I could shift the whole logic onto the webpage. 

The DataSources are identified (parsed for string DataSource) to find the correct progress bars to change. 

Therefore the webpage can stay event based. Additionally code is reduced dramatically. 

It can be found in a new pull request: Introduce two in one progress bars #915

PR #902, opened 2014-06-03
Webclient

MoeweX:
Hi,

I integrated the webclient (#711) into the old infrastructure. You should be able to use it like you are used too (upload jars, etc.).

Start the client with bin/start-webclient.sh and navigate to localhost:8080 to test it out. If you want to run a job you should also do bin/start-local.sh .


MoeweX:
Any thoughts at all? @rmetzger @StephanEwen

rmetzger:
I did not have time yet to check it out, sorry.

rmetzger:
Can you do me a favor and squash together the commits into one and then re-open the pull request against https://github.com/apache/incubator-flink?

PR #903, opened 2014-06-04
ML example for java API

wilsoncao:
Hi, here is the linear regression example for the new JAVA API.

### DIFF
--- a/stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/ml/LinearRegression.java
+++ b/stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/ml/LinearRegression.java
@@ -24,6 +24,7 @@ import eu.stratosphere.api.java.functions.MapFunction;
 import eu.stratosphere.api.java.functions.ReduceFunction;
 import eu.stratosphere.api.java.tuple.Tuple2;
 import eu.stratosphere.configuration.Configuration;
+import eu.stratosphere.example.java.ml.util.LinearRegressionData;
 
 /**
  * This example implements a basic Linear Regression using batch gradient descent algorithm.
--- a/stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/ml/LinearRegression.java
+++ b/stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/ml/LinearRegression.java
@@ -59,8 +59,6 @@ public class LinearRegression {
     // *************************************************************************
 	//     PROGRAM
 	// *************************************************************************
-    
-    
     public static void main(String[] arg) throws Exception{
         
         // set up execution environment
--- a/stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/ml/LinearRegression.java
+++ b/stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/ml/LinearRegression.java
@@ -38,9 +38,17 @@ import eu.stratosphere.example.java.ml.util.LinearRegressionData;
  * This is the Wikipedia entry for the <a href = "http://en.wikipedia.org/wiki/Linear_regression">Linear regression</a> and <a href = "http://en.wikipedia.org/wiki/Gradient_descent">Gradient descent algorithm</a>.
  * 
  * <p>
- * This implementation works on two-dimensional data.<br>
+ * This implementation works on one-dimensional data. And find the two-dimensional theta.<br>
  * It find the best Theta parameter to fit the target.
- *
+ * 
+ * <p>
+ * Input files are plain text files and must be formatted as follows:
+ * <ul>
+ * <li>Data points are represented as two double values separated by a blank character. The first one represent the X(the training data) and the second represent the Y(target).
+ * Data points are separated by newline characters.<br>
+ * For example <code>"-0.02 -0.04\n5.3 10.6\n"</code> gives two data points (x=-0.02, y=-0.04) and (x=5.3, y=10.6).
+ * </ul>
+ * 
  * <p>
  * This example shows how to use:
  * <ul>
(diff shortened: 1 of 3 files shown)


stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/ml/LinearRegression.java
```diff
@@ -0,0 +1,248 @@
...
+/**
+* This example implements a basic Linear Regression using SGD algorithm.
+*
+* <p>
+* Linear Regression with SGD(Stochastic gradient descent) algorithm is an iterative clustering algorithm and works as follows:<br>
+* Giving a data set and target set, the SGD try to find out the best parameters for the data set to fit the target set.
+* In each iteration, the algorithm computes the gradient of the cost function and use it to update all the parameters.
+* The algorithm terminates after a fixed number of iterations (as in this implementation)
+* With enough iteration, the algorithm can minimize the cost function and find the best parameters
+* This is the Wikipedia entry for the <a herf = "http://en.wikipedia.org/wiki/Linear_regression">Linear regression</a> and <a herf = "http://en.wikipedia.org/wiki/Stochastic_gradient_descent">SGD algorithm</a>.
```
fhueske: herf should be href (twice)

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/ml/LinearRegression.java
```diff
@@ -0,0 +1,248 @@
...
+* 
+* <p>
+* This implementation works on two-dimensional data.<br>
+* It find the best Theta parameter to fit the target.
+*
+* <p>
+* This exmaple shows how to use:
```
fhueske: exmaple -> example

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/ml/LinearRegression.java
```diff
@@ -0,0 +1,248 @@
...
+                                              new Data(1.0,2.0));
+        // get the parameters from elements
+		DataSet<Params> parameters = env.fromElements(
+                                                      new Params(0.5, 1.0));
+        // set number of bulk iterations for SGD linear Regression
+		IterativeDataSet<Params> loop = parameters.iterate(700);
+        
+		DataSet<Params> new_parameters = data
+        // compute a single step using every sample
+        .map(new SubUpdate()).withBroadcastSet(loop, "parameters")
```
fhueske: The project uses tabs (and not spaces) for code indentions. 
Our build process is configured to check for that and will fail if it finds a violation. We do this to prevent improve source diffs.
Could you please replace space indentions by tabs?

wilsoncao: I will fix it now.

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/ml/LinearRegression.java
```diff
@@ -0,0 +1,248 @@
...
+	 * Compute a single SGD type update for every parameters.
+	 */
+	public static class SubUpdate extends MapFunction<Data,Tuple2<Params,Integer>>{
+
+
+		private Collection<Params> parameters; 
+
+		private Params parameter;
+
+		private int index = 1;
```
fhueske: What is the purpose of the index?
It is nowhere updated and basically adds a constant 1-valued field to each output tuple.

fhueske: Ah, sorry... Should have looked at the whole code...
You use it for counting :+1:

A different name might make that more intutiive though ;-)

wilsoncao: Is it 'count' better?

fhueske:
Hi @wilsoncao 

Thanks for the pull request. The code looks really good! :+1: 
I made some minor comments in line and have some more:
- Our other examples are written in a way that they run out of the box on default data (such as yours) but also accept program parameters to read data from files (checkout for example the [KMeans example](https://github.com/stratosphere/stratosphere/blob/master/stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/clustering/KMeans.java)). Could you extend the example such that it also accepts program parameters and works on data which is read from files? Also some more default data would be nice (can be put in another class to keep the example concise).
- There are a few double blank lines. I know its a question of taste, but could you change it to single blank lines to keep the example shorter?

Again, thanks for contributing this example :-)
Since we are going to Apache, you have to sign an ICLA: http://www.apache.org/licenses/icla.pdf before we can merge the code.

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/ml/LinearRegression.java
fhueske: Yep, perfect IMO :smile:

rmetzger:
@sscdotopen just told me that contributors do not need to sign an ICLA.

fhueske:
Alright, even better :smile:
Sorry for the confusion

wilsoncao:
@fhueske In DataGenerator class, should all the data follow the random pattern? I mean it is necessary to generate all the data from Normal Distribution or Uniform Distribution?

fhueske:
Hi @wilsoncao, a data generator is not mandatory for an example job, but of course very appreciated :-)
You can implement it as you like (the data should somehow show that the algorithm works).

markus-h:
I had a quick look at your code and just want to add a small remark: As far as I can see you are not using Stochastic Gradient Descent but Batch Gradient Descent. In every iteration you build the average gradient over all point, right? Perhaps you can adjust the javadocs accordingly, so that people don't get confused.

wilsoncao:
@fhueske  I have add the data generator class and data class , welcome for advice :-)

wilsoncao:
@markus-h I agree with you. It is actually Batch Gradient Descent. When I first code it I follow a tutorial about SGD working on Hadoop but after I finish it turns out to be Batch Gradient Descent....Just wondering, if I want to implement SGD in parallelize way, what should I do?

markus-h:
@wilsoncao Plain SGD is inherently not parallelizable. What you can do is change the algorithm a bit: Distribute your points accross multiple nodes, let SGD run on each node and then average the result (instead of averaging each step like in BGD)

fhueske:
Was merged.
Thanks!

PR #904, opened 2014-06-04
Adding graph centrality closeness implementation

JananiC:


rmetzger:
Thank you very much for contributing code to the Stratosphere project. I'll review your pull request and give you some feedback

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/graph/centrality/EffectiveCloseness.java
```diff
@@ -0,0 +1,442 @@
+package eu.stratosphere.example.java.graph.centrality;
```
rmetzger, line 14: This file is missing a license header

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/graph/centrality/EffectiveCloseness.java
```diff
@@ -0,0 +1,442 @@
...
+	
+	private static final String ZERO = "0";
+	private static final String TAB_DELIM = "\t" ;
+	private static final String NEWLINE = "\n" ;
+	private static final String COMMA_DELIM = "," ;
+	static final String COMMA_STR = "comma" ;
+
+	public static void main(String[] args) throws Exception {
+
+	if (args.length < 6) {
```
rmetzger: The indentation is a bit broken here.
The method signature and the body are at the same level.

stratosphere-examples/stratosphere-java-examples/src/main/java/eu/stratosphere/example/java/graph/centrality/EffectiveCloseness.java
```diff
@@ -0,0 +1,442 @@
...
+	public AverageComputation(String numVer) {
+		this.noOfver = Integer.parseInt(numVer);
+	}
+	@Override
+	public Tuple2<Long, Double> map(
+		Tuple3<Long, VertexWithBitArray, Double> value)
+		throws Exception {
+		Double closeness = 0.0;
+		if (value.f2.doubleValue() > 0) {
+		closeness = (noOfver - 1) / value.f2.doubleValue();
```
rmetzger: can you also indent this one right.

rmetzger:
Some questions regarding the code (that looks pretty good already):
- Why does the user has to supply the numberOfVertices as an argument? (Know you need the total number of vertices) But can't you use Stratosphere to count them (aggregate.SUM) and broadcast the count into the operator that needs the numberOfVertices?
  In addition to that

``` java
int noOfver;
public AverageComputation(String numVer) {
    this.noOfver = Integer.parseInt(numVer);
}
```

`noOfver` is an integer value. This might be a problem for very large graphs.
- What data did you use to test the implementation on a cluster?
- If you want, you can add your name to the [CONTRIBUTORS](https://github.com/stratosphere/stratosphere/blob/master/CONTRIBUTORS) file.
- You can update pending pull requests by pushing into the branch that contains the code.
- Since we are going to Apache, you have to sign an ICLA: http://www.apache.org/licenses/icla.pdf before we can merge the code.

JananiC:
Thanks for the review comments. I will fix the broken indentations. To answer the questions,

Why does the user has to supply the numberOfVertices as an argument?
Since most of the graph input data sets that I have seen in the websites have already listed the general statistics about the data such as number of vertices, number of edges , etc., so I just decided to not compute the already known values. But I can add the vertices count as broadcast variable and update the code as well. 

What data did you use to test the implementation on a cluster?
I tested for DBLP prolific authors dataset (vertices=2879 and edges=22652) in cluster. I compared it with the result of Giraph BitfieldCloseness implementation. Following are the correlation results of EffectiveCloseness in stratosphere and BitfieldCloseness in Giraph

Pearson's : 0.9936280641299453
Spearman's : 0.9885277456081363
Kendall's : 0.9093079713728399

But I am still trying to experiment for larger datasets

rmetzger:
Cool. Your argumentation makes sense. I think you don't need to introduce the broadcast variable for now.
But I would recommend to change the type of the variable to long

fhueske:
Hi @jananiC, thanks for the PR!

+1 for changing to Long.

The PageRank example program also requires the number of vertices as an argument. I think that's fine.

JananiC:
Hi, I have updated changes for data type and other indentations. It is not clear for me to which branch I should push the code now?

rmetzger:
To your master

uce:
Thank you for your PR. It's a great application of Stratosphere, but I think it's too "big"/involved as an example. Currently (most of) the examples demonstrate certain aspects of the system and your example would be a better fit for something like a library of Stratosphere programs.

rmetzger:
I've tried to start a discussion about this here: https://issues.apache.org/jira/browse/FLINK-904
But there were no answers yet.

PR #908, opened 2014-06-05
Add support for POJO objects

aljoscha:
This adds support in the `TypeExtractor` and also one can now use field
expressions for grouping, join, and co-group.

I decided against using ReflectASM for now since it does not work on private fields. We would have to mix ReflectASM and plain java reflection which right now does not seem worth the hassle.


uce:
This PR is subsumed by the proposed changes in [FLINK-947](https://issues.apache.org/jira/browse/FLINK-947).

Issue #910, opened 2014-06-06
Type extractor fails in Java 6


StephanEwen:
The return type `Tuple2<Integer, double[]>` cannot be extracted in Java 6.

twalthr:
I can work on this issue next week.

StephanEwen:
Thanks. I added a test case #911 so it should be easy to reproduce. You
need to run java6, though. 7 & 8 work fine.

PR #911, opened 2014-06-06
Add test for broken type extraction in with Java 6 JVM.

StephanEwen:


StephanEwen:
Test case for issue #910

twalthr:
I have created an extended test case in https://github.com/apache/incubator-flink/pull/12, so this PR can be closed.

Issue #849, opened 2014-05-22
Add Tab in Webfrontend that lists all configuration values


rmetzger:
Hadoop also allows to inspect the configuration values from the webinterface.
I think it is not too much work and a very helpful feature.

rmetzger:
Showing some details about the connected TaskManagers would also be great.

uce:
+1

StephanEwen:
+1

MoeweX:
Done, waiting for merge #919 
What informations from the TaskManagers do you want? I can try to include it, too.

rmetzger:
As part of our move to the Apache Software Foundation, we are now using their JIRA for issue tracking.
I also imported this issue to JIRA: https://issues.apache.org/jira/browse/FLINK-849
Lets continue the discussion there.

PR #914, opened 2014-06-06
Added globalConfig subsection in WebInterface

MoeweX:
Hi,

I implemented a subsection in the Webinterface for the Global Configurations. It looks like this
bildschirmfoto 2014-06-06 um 15 53 49
bildschirmfoto 2014-06-06 um 15 53 58

@rmetzger


rmetzger:
Thank you.
Can you add the link to the left (next to overview and history) instead of the "Log Files" drop down (which does not make any sense anyways) ?

The configuration view should be integrated into the layout (like the overview and history)

rmetzger:
This is a fix for issue: https://github.com/stratosphere/stratosphere/issues/849

stratosphere-runtime/src/main/java/eu/stratosphere/nephele/jobmanager/web/GlobalConfigServlet.java
```diff
@@ -0,0 +1,71 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import eu.stratosphere.configuration.Configuration;
+import eu.stratosphere.configuration.GlobalConfiguration;
+
+/**
+ * A Servlet that displays the GlobalConfigruation in the webinterface.
+ * @author jonathanhasenburg
```
rmetzger: We don't use `@author` tags in our code.

stratosphere-runtime/src/main/java/eu/stratosphere/nephele/jobmanager/web/GlobalConfigServlet.java
```diff
@@ -0,0 +1,71 @@
...
+ *
+ */
+public class GlobalConfigServlet extends HttpServlet {
+
+	/**
+	 * Serial UID for serialization interoperability.
+	 */
+	private static final long serialVersionUID = 3704963598772630435L;
+	
+	private Configuration c;
```
rmetzger: how about calling the variable "config" or "configuration" or so?

MoeweX:
Hm, ok... . I choosed the "easy way" (your briefing sounded more like this)  by generating the page with java. If you want me to integrate it in the normal layout I need to create an additional html file and send the information via ajax ... . 
I can't do this today, I am going to do this in the next days... .

rmetzger:
I mean this solution is fine .. but it would be nicer to have it concise with the rest of the page.
As a next step, I would also like to add information about the connected TaskManagers (memory, cpus etc.)

MoeweX:
What do you think now?
bildschirmfoto 2014-06-09 um 01 09 32

rmetzger:
Amazing!
I suggest to rename it from "Global Configuration" to "Configuration".
Can you squash together the commits so that your pull request is only one commit? (Use force push to update the branch this PR is based on)

PR #916, opened 2014-06-06
Implement AvroOutputFormat

qmlmoon:


stratosphere-addons/avro/src/main/java/eu/stratosphere/api/java/io/AvroOutputFormat.java
```diff
@@ -0,0 +1,80 @@
...
+	public void open(int taskNumber, int numTasks) throws IOException {
+		super.open(taskNumber, numTasks);
+		DatumWriter<E> datumWriter;
+		Schema schema = null;
+		if (org.apache.avro.specific.SpecificRecordBase.class.isAssignableFrom(avroValueType)) {
+			datumWriter = new SpecificDatumWriter<E>(avroValueType);
+			try {
+				schema = ((org.apache.avro.specific.SpecificRecordBase)avroValueType.newInstance()).getSchema();
+			} catch (InstantiationException e) {
+				e.printStackTrace();
```
rmetzger: Please do not swallow exceptions. Please rethrow them with an appropriate message (using a RuntimeException so)

StephanEwen:
I think we should test that with a few different types of objects (plain
Java, Avro generated).

A few non working cases would also be good to test for proper error
handling.

Also, I think it is useful to have a method that accepts a schema in which
the objects should be written. That's important for "schema" evolution.

qmlmoon:
I think I already added the plan java and avro generated test case. Do you mean to have a few more?

For the error handling, actually i just follow the `AvroInputFormat` and there's no error handling now. To be honest I can't find some error cases here to be handled:). What error handling do you expect here?

Don't add a method for "schema" is also because i want to make it somehow symmetric to the `AvroInputFormat`

StephanEwen:
I think the output format needs a variant where you pass a schema. In order
to keep the same output data in the presence of evolving objects.

StephanEwen:
Merged in f82c61fdcf9cc8a77eb2aed0a38abc615e18ff03

